Issue #213, opened 2014-10-10
Integrate doesn't work


KOLANICH:
```Integrate[Sin[x^2],{x,a,b}]``` writes 0;

**Comment by sn6uv** [comment]
These have now been fixed:

```mathematica
In[1]:= Integrate[Sin[x^2],{x,a,b}]
Out[1]= -3 Sqrt[Pi] Sqrt[2] (-1 / 4)! fresnels[Sqrt[2] a / Sqrt[Pi]] / (8 (3 / 4)!) + 3 Sqrt[Pi] Sqrt[2] (-1 / 4)! fresnels[Sqrt[2] b / Sqrt[Pi]] / (8 (3 / 4)!)

In[2]:= z=Integrate[(1-Exp[-x^4])/(x^4), {x, -Infinity, Infinity}]
Out[2]= Integrate[(1 - E ^ (-x ^ 4)) / x ^ 4, {x, -Infinity, Infinity}]
```

PR #220, opened 2014-11-05
213 attempted solution

GitAnt:
I tried to solve #213, the result is compatible with the mathematica result even if some ratios of gamma functions are not simplified properly


sn6uv:
What's the status on this PR?

PR #257, opened 2015-06-13
Created some tests for direct evaluation in mathics

wolfv:
Hi, 

I added a unit test for direct evaluation in a 'virtual' Mathics shell. I think this will make it handy to write some tests for different functions, especially with list slicing and flattening in mind. 

Wolf

### DIFF
--- a/test/test_mathics.py
+++ b/test/test_mathics.py
@@ -1,0 +1,76 @@
+import sympy
+import random
+import sys
+import re
+
+import mathics
+from mathics import settings
+from mathics.core.evaluation import Evaluation
+from mathics.core.definitions import Definitions
+
+if sys.version_info[:2] == (2, 7):
+    import unittest
+else:
+    import unittest2 as unittest
+
+definitions = Definitions(add_builtin=True)
+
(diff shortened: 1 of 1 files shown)


wolfv:
Ah, just realized that the docstrings are test cases too. Cool! 
Though, this might make sense anyways as it would hide really strange tests from the user.

wolfv:
Ah, just realized that the docstrings are test cases too. Cool! 
Though, this might make sense anyways as it would hide really strange tests from the user.

sn6uv:
That's right, lines in the docstrings beginning with `>>` are part of the docs and also used as test cases. Additionally, lines beginning with `#>` are silent tests.

For tests that are specific to a single function we try to put them in the docstrings for that function definition.

The `tests/` directory is mostly for tests that aren't specific to a single function. For example testing of the parser or the command line interface (we don't have any tests for the web interface yet).

sn6uv:
These are are all good tests. Also, it's great to see you beginning to play with the Mathics internals

wolfv:
Great to know! Cool invention with the tests in the comments :)

wolfv:
It's really fun to create some functions for Mathics, by the way.

PR #258, opened 2015-06-13
Added three functions for linalg: SVD, Least Squares and PseudoInverse

wolfv:
SVD is only numeric and implemented via mpmath. Not sure if that's acceptable or not? Might need a check, too.

### DIFF
--- a/mathics/builtin/linalg.py
+++ b/mathics/builtin/linalg.py
@@ -5,6 +5,7 @@ Linear algebra
 """
 
 import sympy
+from mpmath import mp
 
 from mathics.builtin.base import Builtin
 from mathics.core.convert import from_sympy
@@ -24,6 +25,22 @@ def to_sympy_matrix(data, **kwargs):
         data = matrix_data(data)
     try:
         return sympy.Matrix(data)
+    except (TypeError, AssertionError, ValueError):
+        return None
+
+def to_mpmath_matrix(data, **kwargs):
+    def mpmath_matrix_data(m):
+        if not m.has_form('List', None):
+            return None
+        if not all(leaf.has_form('List', None) for leaf in m.leaves):
+            return None
+        return [[str(item) for item in row.leaves] for row in m.leaves]
+
+
+    if not isinstance(data, list):
+        data = mpmath_matrix_data(data)
+    try:
+        return mp.matrix(data)
     except (TypeError, AssertionError, ValueError):
         return None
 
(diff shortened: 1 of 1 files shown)


mathics/builtin/linalg.py
```diff
@@ -83,7 +100,93 @@ def apply(self, m, evaluation):
...
+    """
+
+
+    def apply(self, m, evaluation):
+        'SingularValueDecomposition[m_]'
+
+        matrix = to_mpmath_matrix(m)
+        U, S, V = mp.svd(matrix)
+        S = mp.diag(S)
+        print(U, S, V)
```
sn6uv: Don't forget to remove this.

sn6uv:
Sympy implements the singular value decomposition for exact inputs, see (here)[http://docs.sympy.org/dev/modules/mpmath/matrices.html#the-singular-value-decomposition]. We should probably us this version if the argument is exact and the mpmath version otherwise.

sn6uv:
Also, we have a helper class function for wrapping Sympy functions. It's called `SympyFunction` and is defined in `mathics/builtin/base.py`. There are heaps of examples scattered around `mathics.builtin`in particular `mathics/builtin/arithmetic.py`

There's also `_MPMathFunction` which subclasses `SympyFunction` and will use the relevant Sympy or mpmath functions depending on the `exactness` of the arguments.

Let me know if you have any questions about this.

wolfv:
Hey,

actually the function you're linking is the one I am using. But if you look closely, it's the mpmath Matrix package of sympy and unfortunately not the sympy Matrix... Took me some time to realize, too.

So I guess at least sympy doesn't have a algebraic solution up to now.

mathics/builtin/linalg.py
wolfv: Ahh, thanks!

sn6uv:
Oh, you're right sorry. I see there's an Sympy issue https://github.com/sympy/sympy/issues/6832 to track this missing functionality. The only option we have is to provide the numerical SVD (as you've done).

mathics/builtin/linalg.py
```diff
@@ -83,7 +100,93 @@ def apply(self, m, evaluation):
...
+class SingularValueDecomposition(Builtin):
+    """
+    <dl>
+    <dt>'SingularValueDecomposition[$m$]'
+        <dd>Calculate the singular value decomposition for matrix $m$.
+        Returns $u, s, w$ such that $m=u s v$, $u'u=1$, $v'v=1$, $s$ is diagonal.
+    </dl>
+
+    >> SingularValueDecomposition[{{1, 2}, {2, 3}, {3, 4}}]
+     = {{-11 / 6, -1 / 3, 7 / 6}, {4 / 3, 1 / 3, -2 / 3}}
```
sn6uv: This test is failing (because we get inexact answers). It's nice to have the tests pass, and unfortunately in the current system there's no way to mark a test as xfail. I'd suggest adding a Line before this test explaining why the results are inexact, and then testing the inexact results.

Let me know if that's unclear.

mathics/builtin/linalg.py
```diff
@@ -83,7 +100,93 @@ def apply(self, m, evaluation):
...
+
+    >> PseudoInverse[{{1, 2, 0}, {2, 3, 0}, {3, 4, 1}}]
+     = {{-3, 2, 0}, {2, -1, 0}, {1, -2, 1}}
```
sn6uv: This one too.

sn6uv:
Ideally I'd like to get this one into 0.9 too.

sn6uv:
Opened #321.

PR #261, opened 2015-06-20
remove initalize.py

sn6uv:
Since Mathics 0.6 the database creation has been automatic. It is probably safe to remove this file now.


sn6uv:
We should wait until #262 is fixed before merging this I think.

PR #263, opened 2015-06-29
Added correct indexing for negative steps and check for 0-index

### LINKED ISSUE #196: Span does not behave correctly
Span seems to be misbehaving for me...

The most egregious example I can find is the standard reverse operation:
{1,2,3,4,5}[;; ;;-1]

This should obviously give:
{5,4,3,2,1}

As a starting point, you might want to look at something like "genericSlice" from my np_utils project. It builds slices, but with 3 additional options (includeStop, oneBased, and checkBounds):

https://github.com/davidmashburn/np_utils/blob/master/np_utils/list_utils.py

I also made a gist with just the function with a simple test case:
https://gist.github.com/davidmashburn/9764309

I know Python's indexing is: includeStop=False, oneBased=False, checkBounds=False.
Matlab and Octave would be: includeStop=True, oneBased=True, checkBounds=True.
I think Mathematica should be as well, but you may have to double-check me...
Indexing can get tricky...

If that turns out to be right, you could simply the function by hard-coding those values to eliminate some options.

Cheers,
-David

wolfv:
This fixes stuff like 
a = {1,2,3,4,5}
a[[3;;1;;-1]] == {3,2,1}

which previously resulted in {2}.

This should fix #196

### DIFF
--- a/mathics/builtin/lists.py
+++ b/mathics/builtin/lists.py
@@ -276,10 +276,18 @@ def walk_parts(list_of_list, indices, evaluation, assign_list=None):
                         return False
             if len(index.leaves) > 2:
                 step = index.leaves[2].get_int_value()
+
+            if start == 0 or stop == 0:
+                # index 0 is undefined
+                evaluation.message('Part', 'span', 0)
+                return False
+
             if start is None or step is None:
                 evaluation.message('Part', 'span', index)
                 return False
-            start, stop = python_seq(start, stop)
+
+            start, stop = python_seq(start, stop, step)
+
             for inner in inner_list:
                 if inner.is_atom():
                     evaluation.message('Part', 'partd')
@@ -533,14 +541,31 @@ class LevelQ(Test):
             return False
 
 
-def python_seq(start, stop):
-    if start > 0:
-        start -= 1
-    if stop is not None and stop < 0:
-        stop += 1
-        if stop == 0:
+def python_seq(start, stop, step=+1):
+    if not step: step = +1
+    if step == +1:
+        if start > 0:
+            start -= 1
+        if stop is not None and stop < 0:
+            stop += 1
+            if stop == 0:
+                stop = None
+        return start, stop
+    else:
+        if stop and stop > 0:
+            stop -= 2
+        elif stop and stop < 0:
+            stop -= 1
+        if start and start > 0:
+            start -= 1
+        if not start:
+            start = None
+        if stop == -1:
             stop = None
-    return start, stop
+        elif stop == None:
+            # No stop equals len(list)!
+            stop = -1
+        return start, stop
 
 
 def convert_seq(seq):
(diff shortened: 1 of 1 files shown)


sn6uv:
Thanks for fixing this. I'm finding the `python_seq` function a little bit confusing. Do you think you could clean it up and add a short docstring explaining its purpose?

wolfv:
Do you think there is a way to fix all of these at once? 

Ie. `Table[i,{i,1,-10,-1}] -> {1,0,-1...-10}` but evaluates to `{}` currently
Also `Range[i+1, i+10] -> {i+1, i+2, i+3...}` but doesn't evaluate at all.
`Range[0,-10,-1]` doesnt work either.

wolfv:
```
def python_seq(start, stop, step=+1):
    # step can be None, which means positive stepsize
    if not step: step = +1

    # What this function does is convert a 1 based index to
    # zero based index. Additionally, the start of a list
    # is indicated by None in python.
    print("stepping")
    if step > 0:
        if start > 0:
            start -= 1
        if stop is not None and stop < 0:
            stop += 1
            if stop == 0:
                stop = None
        return start, stop
    else:
        # in the case of a negative step, 
        # admissible values are start < stop
        # so that start get's decremented until stop is 
        # reached. 
        # However, note that in python a = [1, 2, 3, 4][3:0:-1] == [4, 3, 2]
        # and not, as one might think [4, 3, 2, 1] (which is the result of [3:None:-1])
        # Therefore, stop has -2 subtracted, and None is returned when stop == -1
        if stop and stop > 0:
            stop -= 2

        elif stop and stop < 0:
            stop -= 1

        print('stop: ', stop)

        if start and start > 0:
            # zero base index
            start -= 1

        if not start:
            # no start index = include all (0)
            start = None

        if stop == -1:
            # stop equals end, which equals -1
            stop = None
        elif stop == None:
            # No stop equals len(list)
            # This is actually weird in Mathematica. 
            # E.g. a = {1,2,3,4}; a[[4;; ;; -1]] == {4}
            # but all other combinations evaluate to an error
            # a[[3 ;;  ;; -1]] = cannot take positions -4 through -1
            # in Mathics it all evaluates to an empty list currently.
            stop = -1
        return start, stop
```

sn6uv:
What's the status of this PR? That code in the comments looks much better but there are still some unhandled cases I think e.g. `python_seq(-2, -3, 1)` returns `(-2, -2)`.

wolfv:
Your example is the actually correct:

selection_081

But you're right that this piece of code should get some proper testing and specification.

sn6uv:
I'm planning to release Mathics 0.9 in the next couple of weeks. Would love to get these changes merged in if possible. @wolfv are you planning to finish this?

sn6uv:
closed in favour of #325

PR #252, opened 2015-06-11
List functions for Total, Accumulate, FoldList, Fold

wolfv:
I added some functions to lists that were missing from Mathics. 

Unfortunately, the `Flatten` doesn't work the same way as in Mathematica. Therefore, Total is incomplete. 

Maybe someone with more knowledge of Mathics can fix that easily.

Cheers,

Wolf

PR #264, opened 2015-06-29
Reopening #252

wolfv:
Something went wrong with the last pull request, but this is a hopefully cleaner version. 

Cheers,

Wolf


wolfv:
By the way, I am not sure what NHoldRest does, I kept it from another function I used as template. Care to explain? Might/Should be removed.

PR #276, opened 2015-07-21
Adding support for numerical solvers in linalg.py module

mmatera:
I just add on some of the routines in linalg.py ( Det, Eigenvalues, Eigenvectors, Eigensystem, Inverse, MatrixPower, MatrixExp) some lines checking if the argument is a numeric matrix, and if this is the case, call the routines in the numpy library instead the sympy ones.


sn6uv:
I haven't had a chance to review this properly yet. I'm hoping to get some time on the weekend.

These are really important features and it's great to see them added. Both the `Information` command and the linear algebra.

Do you think it's possible to split this into two pull requests? That is, move the information stuff into a separate PR.

sn6uv:
See #278

PR #277, opened 2015-07-29
Information

mmatera:
Adding the Information module as a new branch


sn6uv:
What's the status of this PR. @mmatera are you planning to fix the failed tests?

mmatera:
Yes, I planned to fix it, but I couldn't find the way to do it yet. The problem is that the doc string have some things that maybe you don't want to show with information (for example, the tests lines). For these reason, I think that  I should rewrite the whole function, and hence I left it by now.

PR #278, opened 2015-07-29
adding scipy to the  requirements

mmatera:
As a first step, I am trying to add the scipy library to the list of requirements


mmatera:
In this pull request I included the support for the scipy  library, and a way to support numeric evaluation for Det and Inverse. Regarding the scipy support, I found a problem making it compatible with pypy. Maybe there is a way to solve it, but I didn't found it.
In the next pull request I hope to add support for numerical evaluation of Eigenvalues problems.

sn6uv:
Hi @mmatera thanks for the contribution. I'm hesitant to add numpy+scipy to the requirements of Mathics. I think mpmath can implement the same functionality (and it's pure python so works with PyPy).

I'm not absolutely opposed to adding numpy, it's a great linear algebra library and I think they're working on PyPy support. The main problem with numpy is that it doesn't offer arbitrary precision calculations (unlike mpmath/sympy). Since Mathics aims to support arbitrary precision calculations we should use mpmath instead.

Apologies for the delay in giving you feedback on this.

I'm closing this for now, but if you want to rewrite this to use mpmath (see #321 for example) I'd love to see this functionality added to Mathics.

mmatera:
I think that to use arbitrary precision arithmetic for diagonalisation could affect strongly the performance. Actually, using generic python lists for numerical calculations represents a bottleneck in the current implementation. 

Regarding consistency with the rest of the functions, notice for example that in wolfram mma, all the numerical eigenvalues function relays on lapack. 


Regarding avoid enlarge too much the system requirements,  what I could try to implement is something that, depending on the availability of numpy in the system, use this library for numerical linear algebra, and use the standard routines if is is not installed.

sn6uv:
> I think that to use arbitrary precision arithmetic for diagonalisation could affect strongly the performance. Actually, using generic python lists for numerical calculations represents a bottleneck in the current implementation. 

There is a performance cost of course, but it's still something we should offer if possible. Mathematica does:
```
>> A = {{Pi, 1/3}, {0, E}};
>> Eigenvalues[N[A, 25]]
{3.141592653589793238462643, 2.718281828459045235360287}
```


> Regarding avoid enlarge too much the system requirements, what I could try to implement is something that, depending on the availability of numpy in the system, use this library for numerical linear algebra, and use the standard routines if is is not installed.

This is certainly an option. I.e. if the max precision is `$MachinePrecision` use numpy otherwise use mpmath. We already do something similar with cython since it also has problems with PyPy.

mmatera:
The idea is to keep the analytical solution for symbolic matrices, (such  as your example) but use a numerical code to solve numerical matrices (with only numerical entries and at least on real entry).
By the way, in mma, it seems that the way in which internally numerical and symbolic matrices are stored is different, and changes when you modify elements. It seems to be the way to improve a lot the performance when you deal with large vectors and matrices. I think we should implement something like that in the future.

M

PR #279, opened 2015-08-04
Add a basic tox setup

sn6uv:
You're no doubt familiar with the way Travis runs the test suite across
various versions of Python, with versions versions of sympy

Tox gives similar functionality on your development machine.

If you just run tox, it will run the tests defined in each environment -
in this case, there are 6 versions of python tested against 4 versions
of sympy each.

If you just want to test (for instance) py27 with sympy-master, you can
use `tox -e py26-sympymaster`

You can get a full list of all available environments with `tox -l`

Each combination will run in its own virtualenv; these are under .tox in
the project root.

I think I've managed to capture most of the logic expressed in the
travis.yml in this file

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,45 @@
+[tox]
+envlist =
+    {py26,py27,pypy,py33,py34,py35}-sympy{073,074,075,master}, pep8, py35pep8
+
+[testenv]
+basepython =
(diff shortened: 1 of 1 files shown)


sn6uv:
This is migrated from https://github.com/sn6uv/Mathics/pull/1.

PR #300, opened 2015-10-23
Fix gh-pages index repo link

Arthur2e5:
Maybe you guys should find a way to make use of mathics.github.io rather than mathics.github.io/Mathics too?

### DIFF
--- a/index.html
+++ b/index.html
@@ -16,7 +16,7 @@
     <!-- HEADER -->
     <div id="header_wrap" class="outer">
         <header class="inner">
-          <a id="forkme_banner" href="https://github.com/sn6uv/Mathics">View on GitHub</a>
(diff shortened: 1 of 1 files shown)


sn6uv:
Thanks for that fix.

I've moved this branch to a new repository mathics.github.io (Seems like this is the only way to remove /Mathics from the URL).

In future we should delete the gh-pages branch. I will leave it for now unless I can redirect mathics.github.io/Mathics to mathics.github.io

Arthur2e5:
https://help.github.com/articles/redirects-on-github-pages/ seems to provide some hints.

sn6uv:
@Arthur2e5 Thanks. I implemented the redirect and it seems to be working now.

PR #309, opened 2015-12-01
The Quit/Quit[] issue was fixed.

mmatera:
I just replace the rule by  a new definition. It is not so elegant, but works. By the way, both raise an error if we call from the kernel.


sn6uv:
I tried something like this too but it caused some problems with the overlapping patterns. Could you rebase this on the current master so that it will merge (and the tests will run)?

mmatera:
How can I do that?

Issue #312, opened 2016-02-09
stringriffle


denfromufa:
https://reference.wolfram.com/language/ref/StringRiffle.html

sn6uv:
We could implement this using `Riffle` and `StringJoin`.
e.g. (there are many cases this is just the simplest)
```mathematica
In[1]:= StringRiffle[x_List] := StringJoin[Riffle[x, " "]]
In[2]:= StringRiffle[{"a", "b", "c", "d", "e"}]
Out[2]= a b c d e
```

rocky:
As off 1.1.0rc1,  `StringRiffle` is in `strings.py`. Handling the different kinds of levels though is not. 

Note however that the Rubi package doesn't need this form, but only one of the forms handled as of 1.1.0rc1

PR #313, opened 2016-02-10
Riffle fixes

sn6uv:
`map(None, iter1, iter2)` appears to be dumped in python3.

This commit cleans up some `Riffle` edge cases.


sn6uv:
progress towards #312

PR #317, opened 2016-02-14
Python3 support

sn6uv:
Summary
------------
- ported to python3 in part using the `six` module (with `python-modernize`, `python2 -3`, and manually)
- various str/byte fixes
- fixed some unicode bugs
- a few division bugs (`/` vs `//`)
- minor change to asymptote number formatting (reducing output size)
- dropped `interruptingcow` as it's no longer maintained (implemented a `signal` based interrupt)
- TravisCI tests for python versions 3.2, 3.3, 3.4, 3.5, pypy3

TODO
----
- [x] Website under Py3
- [x] Documentation building
- [x] Unicode input under Py2
- [x] `ListLinePlot[{{-2, -1}, {-1,-1}}]` regression
- [x] django views encoding bug

Performance
----------------
~~There's a major performance regression that appeared in the benchmarks (needs fix):~~
```Mathematica
Expand[(a1+a2)^200]
```
~~e.g. 13.1s vs 1.6s on my machine. Affects both py2 and py3.~~
Fixed. It's even slightly faster now, 1.3s.

Other benchmarks seem unchanged, python3 seems about 10% faster on average.


mathics/__init__.py
```diff
@@ -1,10 +1,17 @@
...
 import sys
-import codecs
-writer = codecs.getwriter("utf-8")
-sys.stdout = writer(sys.stdout)
+import platform
+
+# TODO fix this ugly hack
+# import codecs
+# writer = codecs.getwriter("utf-8")
+# sys.stdout = writer(sys.stdout)
```
sn6uv: This was an ugly hack to get unicode input working. I'm not sure if there's a better approach.

mathics/builtin/physchemdata.py
```diff
@@ -10,10 +14,11 @@
 from mathics.core.expression import (Expression, from_python, Symbol, String,
                                      strip_context)
 from mathics.settings import ROOT_DIR
+import six
 
 
 def load_element_data():
-    element_file = open(ROOT_DIR + 'data/element.csv', 'rb')
+    element_file = open(ROOT_DIR + 'data/element.csv', 'r')
     reader = csvreader(element_file, delimiter='\t')
```
sn6uv: The delimiter can't be unicode for Py2.

If there's a way to get a `str` (not unicode) literal with `six` we can import `unicode_literals` in this module too.

Issue #158, opened 2013-09-19
"manage.py runserver" gives "Interruptingcow can only be used from the MainThread"


bnjones:
Steps to reproduce:

* Install with `setup.py develop`
* Run `python mathics/manage.py runserver` inside the source tree
* Go to http://localhost:8000/

I get a Django exception page when I do this.

I guess this means that `manage.py runserver` runs the Django app in a separate thread. It looks like this check exists because interruptingcow uses `alarm()` and a signal handler for `SIGALRM` to implement its timeouts, and the [Python docs](http://docs.python.org/2/library/signal.html) say that signal handlers can only be set from the main thread.

I'm not sure if this is important enough to be worth fixing - I don't have any ideas about what the fix might be, but I thought it was worth raising it in case it turns out to be important.

sn6uv:
I hit this issue setting up a Mathics mirror on my VPS. The workaround I found was to disable the evaluation timeout, i.e. `TIMEOUT = None` in settings.py. This is obviously not ideal.

sn6uv:
Interrupting cow is no longer maintained. I've coded up a replacement that also uses signals (so it's unlikely to fix this). Django does have it's own signals module, but I'm yet to investigate it.

sn6uv:
I found quite a few people with similar issues to this. Of note is [this stackoverflow entry](https://stackoverflow.com/questions/8572830/pythondjango-signal-handler-and-main-thread) which suggests:

`python manage.py runserver --noreload --nothreading`

sn6uv:
Fixed by #323

PR #323, opened 2016-02-19
threading solution to #158

sn6uv:
@bnjones @poeschko I'm wondering if either of you have thoughts on this. I've tested it with the django style `manage.py runserver` and it appears to work well.

There doesn't seem to be a noticeable performance difference compared to the signals approach.

### DIFF
--- a/mathics/core/evaluation.py
+++ b/mathics/core/evaluation.py
@@ -7,9 +7,10 @@ from __future__ import absolute_import
 
 import six
 import six.moves.cPickle as pickle
+from six.moves.queue import Queue
 
 import sys
-import signal
+from threading import Thread
 
 from mathics import settings
 from mathics.core.expression import ensure_context, KeyComparable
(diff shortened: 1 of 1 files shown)


sn6uv:
I backported this change to mathics 0.8 running on http://mathics.angusgriffith.com/, I'll see how it goes stability wise.

sn6uv:
The problem with this approach is that the user definitions are not stored in the main `defintions` object.

sn6uv:
I've switched to a threading approach. Using threads isn't ideal because of the GIL (we may never interrupt if things get stuck in a long python evaluation). In practise it appears to work okay (it's currently running on http://mathics.angusgriffith.com).

sn6uv:
Looking at 9c26868405cc0a26c62b8928d2b451a983221d09 it appears I've come full circle on this. IIRC `interruptingcow` was introduced to solve getting stuck in the GIL on http://mathics.net. I really don't think there's a simple solution to this.

PR #325, opened 2016-02-22
Pr263

### LINKED ISSUE #196: Span does not behave correctly

sn6uv:
cleanup of #263, fixes #196 and some other issues


wolfv:
Thanks for getting this in!

sn6uv:
Thanks for your original pr. I used some of your examples from #196 to construct some tests.

It's still not 100% compatible with MMA. e.g.

- Mathics:
```mathematica
In[1]:= Range[5][[;; ;; -1]]
Out[1]= {5, 4, 3, 2, 1}
```

- Mathematica:
```mathematica
In[1]:= Range[5][[;; ;; -1]]

Part::take: Cannot take positions 1 through -1 in {1, 2, 3, 4, 5}.

Out[1]= {1, 2, 3, 4, 5}[[1 ;; All ;; -1]]
```

But I don't really see a huge problem with it.

See also https://mathematica.stackexchange.com/questions/22312/list-reversion-inspired-by-python

PR #332, opened 2016-03-09
load unevaluated cells

mportman:
in the web interface i had trouble loading a worksheet that contained unevaluated cells.
the sheet up to and including the first unevaluated cell would load, the remainder would be
missing.

the save command saves unevaluated cells ok, the load command should also load them. 

the small change within this pull request fixes this.

but i'm not a javascript expert!  checking against undefined may not be the best way
of doing the check.


sn6uv:
Thanks for reporting and fixing this!

sn6uv:
I'm not a javascript expert either. I think the check it fine though. Thanks for your contribution!

Issue #331, opened 2016-03-02
% should return the correct value for expressions with surpressed output


triggerNZ:
Behaviour:
```
In[1]:= x=67-5;
In[2]:= %
In[3]:= 
```

This is inconsistent with mathematica, which would output 62 for %, even though In[1] ended with a semicolon

sn6uv:
The relevant section from the MMA docs: `expr1; expr2;` returns value `Null`. If it is given as input, the resulting output will not be printed. `Out[n]` will nevertheless be assigned to be the value of `expr2`.

PR #335, opened 2016-04-02
fix for #331

sn6uv:


bnjones:
If I'm reading the docs correctly, it sounds like `CompoundExpression[___, Null]` still returns `Null`, and the special case just affects the value of `Out`.

How does Mathematica treat `CompoundExpression[x, y, Null]` vs `CompoundExpression[CompoundExpression[x, y, Null], Null]`? If `Out` gets `y` in the first case and `Null` in the second, maybe it's just a special case in the evaluator which treats top-level `CompoundExpression` calls specially?

sn6uv:
@bnjones yes, you're right. I'll fix this up.

sn6uv:
It's actually slightly more nuanced than that: 
```mathematica
In[5]:= CompoundExpression[x, Null, Null]                       
In[6]:= Out[5]
Out[6]= x
```

sn6uv:
Unsurprisingly there are bugs in MMA here. Evaluating the following as a single (notebook) cell:
```mathematica
1
$Line = 1
CompoundExpression[]
```

sets `Out[1]` to `1` and not `Null`

bnjones:
Does that happen even if you take the $Line bit out?

sn6uv:
Yes, it still does.

bnjones:
I guess I didn't understand how evaluation in a notebook cell worked, I assumed it treated everything as one big line. Interesting find, it'd be nice if this stuff was actually documented!

sn6uv:
Ah, my original `$Line` example was wrong (since assigning `$Line$ increments itself immediately). There's no bug here after all
```
>>> 1
    $Line = 0
    CompoundExpression[]

>>> Out[1]
     1
>>> Out[0]
    Null
```

sn6uv:
It's quite an obscure one (but one less bug in Mathics). I guess if you're messing with `$Line$` during evaluations you're probably trying to break something in the first place.

PR #336, opened 2016-04-03
Global system bug

sn6uv:
Hit this trying to get the ```KnotTheory` ``` package working in Mathics.

```mathematica
Begin["`System`"]
```
was returning ``` `System` ```, not the fully qualified ```Global`System```. It turns out `$Context` was being set first correctly to ```Global`System` ``` and reset then to ``` `System` ```.

@bnjones does this fix look right to you?

### DIFF
--- a/mathics/builtin/assignment.py
+++ b/mathics/builtin/assignment.py
@@ -213,11 +213,11 @@ class _SetOperator(object):
             ignore_protection = True
             return True
         elif lhs_name == 'System`$ContextPath':
-            if rhs.has_form('List', None) and all(
-                    valid_context_name(s.get_string_value()) for s in rhs.leaves):
-                evaluation.definitions.context_path = [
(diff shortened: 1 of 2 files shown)


bnjones:
Looks good to me, but what's the effect of not reaching the `add_rule` on line 258? I can't find anything actually wrong:
```
In[1]:= $Context = "`test`"
Out[1]= `test`

In[2]:= $Context
Out[2]= Global`test`

In[3]:= Context[`test2]
Out[3]= Global`test`

In[4]:= OwnValues[$Context]
Out[4]= {HoldPattern[$Context] :> Global`test`}
```
but I don't see where the ownvalues are coming from if we don't add them there?

bnjones:
Oops, I forgot that `set_current_context` does this as well, so the `add_rule` is redundant. In that case, for symmetry I think we should also return after calling `set_context_path` as well.

PR #341, opened 2016-04-08
Document (nearly) everything

bnjones:
There were a few functions I skipped over which I'll come back to, but this PR adds the bulk of the missing documentation. Along the way I also fixed typos and performed a few other minor editing tasks.

One thing I changed was to replace this:
```
<dt>'Plus[a, b]' or 'a + b'
```
with this:
```
<dt>'Plus[a, b]'
<dt>'a + b'
```
to make the alternative operator forms stand out more. IMO this an improvement but I'm happy to drop this if I'm the only one who likes it.


mathics/builtin/comparison.py
```diff
@@ -349,17 +386,42 @@ def apply_other(self, x, y, evaluation):
...
 
 
 class LessEqual(_InequalityOperator, SympyFunction):
+    """
+    <dl>
+    <dt>'LessEqual[$x$, $y$]'
+    <dt>'$x$ <= $y$'
+        <dd>yields 'True' if $x$ is known to be less than or equal to $y$.
+    <dt>'$lhs$ <= $rhs$'
+        <dd>represents the inequality $lhs$ ≤ $rhs$.
```
bnjones: Unfortunately although ≤ shows up in the online documentation, it doesn't make it to the PDF. It looks like the TeX recognises the character but the font doesn't have a glyph for it.

sn6uv:
This looks really good! Looking over it now

mathics/builtin/comparison.py
sn6uv: We could replace them manually i.e. replace '≤' with '\le'. I wonder if there's a better approach though.

mathics/builtin/linalg.py
```diff
@@ -488,17 +491,11 @@ class Eigensystem(Builtin):
...
     </dl>
 
-    >> Eigenvalues[{{1, 1, 0}, {1, 0, 1}, {0, 1, 1}}] // Sort
-     = {-1, 1, 2}
-
-    >> Eigenvalues[{{Cos[theta],Sin[theta],0},{-Sin[theta],Cos[theta],0},{0,0,1}}] // Sort
-     = {1, Cos[theta] + Sqrt[-1 + Cos[theta] ^ 2], Cos[theta] - Sqrt[-1 + Cos[theta] ^ 2]}
-
-    >> Eigenvalues[{{7, 1}, {-4, 3}}]
-     = {5, 5}
```
sn6uv: note to self: these are duplicates of tests in the `Eigenvalues` docs.

mathics/builtin/lists.py
```diff
@@ -2174,27 +2314,15 @@ class Total(Builtin):
...
-            Total all values up to level $n$
-        </dd>
-    </dt>
+        <dd>adds all values up to level $n$.
     <dt>'Total[$list$, ${n}$]'
-        <dd>
-            Total at level ${n}$
-        </dd>
-    </dt>
+        <dd>totals only the values at level ${n}$.
```
sn6uv: I think this should be `'{$n$}'`

mathics/builtin/lists.py
```diff
@@ -2174,27 +2314,15 @@ class Total(Builtin):
...
     <dt>'Total[$list$, ${n_1, n_2}$]'
-        <dd>
-            Total at levels ${n_1, n_2}$
-        </dd>
-    </dt>
+        <dd>totals at levels ${n_1, n_2}$.
```
sn6uv: This one too.

mathics/builtin/inout.py
```diff
@@ -365,8 +377,16 @@ def get_op(op):
...
-
     """
+    <dl>
+    <dt>'ToBoxes[$expr$]'
+        <dd>evaluates $expr$ and converts the result to box form.
+    </dl>
+
+    Unlike 'MakeBoxes', 'ToBoxes' evaluates its argument:
+    >> ToBoxes[a + a]
+     = RowBox[{2, , a}]
```
sn6uv: I'm confused by this test failure:
```
angus@laptop> python mathics/test.py -s ToBoxes                       ~/Mathics
Testing section ToBoxes
   1. TEST ToBoxes[a + a]
Test failed: ToBoxes in Reference of built-in symbols / Input and Output
ToBoxes[a + a]
Result: RowBox[{2, ⁢, a}]
Wanted: RowBox[{2, , a}]

   2. TEST ToBoxes[a + b]
   3. TEST ToBoxes[a ^ b] // FullForm

1 test failed.
```

bnjones: I think this is a U+2062 INVISIBLE TIMES

mathics/builtin/lists.py
bnjones: Thanks! (for both of these)

mathics/builtin/comparison.py
bnjones: IMO the best way to fix this would be to fully support TeX-style math in `$...$`. Some functions could definitely benefit from this (e.g. `SingularValueDecomposition`), plus it helps if we want to say (for example) how the gamma function is defined.

mathics/builtin/inout.py
sn6uv: Ah, I had a sneaking suspicion that it was a unicode issue.

mathics/builtin/comparison.py
sn6uv: I can see the benefit of full TeX support, but I'm not sure the best way to handle the conversions. In principle we want to be able to convert to latex (for docs), html (for online docs) and plaintext (for introspection, e.g. `Information`). We could probably leverage pandoc to do the conversions. Is there a better way?

bnjones: I think we have several options here - maybe we can experiment and see what works best.

For the PDF docs we can just pass the `$...$` construct directly to LaTeX (maybe after inserting `\text` or `\mathrm` commands to make multicharacter variables like `$expr$` look better).

MathJax can interpret TeX math commands, and we already pull it in to handle the examples, so for the HTML docs we could configure MathJax to recognise `$` as a delimiter and then just pass the `$...$` constructs through verbatim (maybe special-casing simple cases, so we don't use MathJax if we just need a superscript or subscript).

If we want to accept TeX syntax in the docs, the plaintext part might be difficult. A totally different approach might be to treat the `$...$` constructs as Mathics expressions, then just apply OutputForm/TeXForm/MathMLForm/etc to produce each type of output. This sounds like it might be tricky (we'd need to do evaluation in doc.py) but it might be the best long-term solution, because we'd be able to leverage more pre-existing code.

sn6uv:
Opened #346 to track that issue with the docs. I'm happy to merge this now if you are?

bnjones:
I'm happy to merge this. Thanks for looking over it!

PR #357, opened 2016-04-27
changes for Manipulate

poke1024:
one more try


sn6uv:
Really great stuff. I'm hoping to find some time to look at this on the weekend.

poke1024:
Travis still fails, not sure how to fix this; seems to be a problem with out of date setuptools now

http://stackoverflow.com/questions/35943606/error-on-installing-ipython-for-python-3-sys-platform-darwin-and-platform

PR #362, opened 2016-04-29
Manipulate[]

poke1024:


sn6uv:
Running Maniupulate in the jupyter console gives

```
In [3]: Manipulate[N[Sin[y]], {y, 1, 20, 2}]
[IPKernelApp] WARNING | Widget Javascript not detected.  It may not be installed properly.
0.841470984807896507
Widget Javascript not detected.  It may not be installed properly.
```

which is acceptable, but not ideal.

Running in the legacy mathics console however (with ipywidgets installed so `_enabled=True)`,
```
In[1]:= Manipulate[N[Sin[y]], {y, 1, 20, 2}]
Out[1]= (<class 'traitlets.traitlets.TraitError'>, TraitError("The 'kernel' trait of a Comm instance must be a Kernel, but a value of class 'NoneType' (i.e. None) was specified.",), <traceback object at 0x7fb1defcfac8>)
```

One way to fix this is to check if `display_data_callback is None` and bail out accordingly.

A better way might be to not import `manipulate` at all if called from the console.

mathics/builtin/manipulate.py
```diff
@@ -0,0 +1,276 @@
...
+    <dl>
+    <dt>'Manipulate[$expr1$, $arg1$, ...]'
+        <dd>allows you to interactively compute and display an expression with different argument values.
+    </dl>
+
+    Manipulate[N[Sin[y]], {y, 1, 20, 2}]
+
+    Manipulate[i^3, {i, {2, x^4, a}}]
+
+    Manipulate[x^y, {x, 1, 20}, {y, 1, 3}]
```
sn6uv: With these examples, if you prefix the lines with `>>` it will add them to the tests. The next line usually has the output (but for these ones it can be blank). See http://www.mathics.net/doc/manual/implementation/documentation-markup/ for details.

mathics/builtin/manipulate.py
```diff
@@ -0,0 +1,276 @@
...
+                clear_output_callback(wait=True)
+
+                line_no = evaluation.definitions.get_line_no()
+
+                new_evaluation = Evaluation(evaluation.definitions, result_callback=display_data_callback,
+                                            out_callback=evaluation.out_callback)
+
+                vars = [Expression('Set', Symbol(name), value) for name, value in kwargs.items()]
+                evaluatable = Expression('ReleaseHold', Expression('Module', Expression('List', *vars), expr))
+                new_evaluation.evaluate([evaluatable], timeout=settings.TIMEOUT)
```
sn6uv: This gives a timeout for each time you slide the slider, nice. I was worried that long-running computations would queue up (each with the default timeout) but it seems to interrupt them?

poke1024:
Added additional checks now if Manipulate has a running Jupyter kernel. If not, it prints out a sensible error message. This works in the old Mathics text console and the old Mathics web console, e.g.:

```
In[2]:= Manipulate[x,{x,1,10}]
 : Manipulate[] only works inside a Jupyter notebook.
Out[2]= $Aborted
```

Only if a Jupyter kernel is linked, will it go on to check for ipywidgets.

sn6uv:
Rather than returning `$Aborted` it's probably better to return the unevaluated Manipulate expression. That's what mathematica does in it's console interface.

I'd suggest:
```
evaluation.message(...)
return None
```

or more succinctly
```
return evaluation.message(...)
```

over using error. Upon returning `None` from an apply method you get the unevaluated expression back.

`evaluation.error` is intended for actual errors in the evaluation (timeouts, exceptions etc.)

mathics/builtin/manipulate.py
```diff
@@ -0,0 +1,292 @@
...
+
+    def apply_discrete_options_var3(self, i, options, idefault, ilabel):
+        '{{i_Symbol, idefault_, ilabel_}, options_List}'
+        self._add_discrete_options_var(i, options, idefault, ilabel)
+
+
+class ListAnimate(Builtin):
+    rules = {
+        'ListAnimate[l_List]': 'Manipulate[Part[l, i], {i, 1, Length[l], 1}]'
+    }
```
sn6uv: Might be worth adding a comment here that this is just a workaround, i.e. `ListAnimate` should really generate an animation. I'm wondering if we'd be better to remove this entirely.

sn6uv:
Continuing this at #366. There are just a couple of things to fix up (e.g. the docs).

PR #363, opened 2016-04-29
a basic implementation for Image[] and related functions

poke1024:
I'm submitting this to the iMathics branch as well now, since it doesn't seem to work on the regular Mathics notebook. It seems to be impossible to push HTML < img > tags into the regular Mathics notebook, which I'm doing in order to display the loaded images (Mathics gives an explicit error, saying that img nodes are an unknown node type). Jupyter doesn't have that limitation though. Needs scikit-image.

### DIFF
--- a/mathics/builtin/image.py
+++ b/mathics/builtin/image.py
@@ -1,37 +1,99 @@
 from mathics.builtin.base import (
-    Builtin, InstancableBuiltin, BoxConstruct, BoxConstructError, String)
+    Builtin, Test, BoxConstruct, String)
 from mathics.core.expression import (
-    Atom, Expression, Integer, Real, NumberError, Symbol, strip_context,
-    system_symbols, system_symbols_dict)
+    Atom, Expression, Integer, Real, Symbol, from_python)
 
-import sys
+'''
+A place for Image[] and related functions.
+'''
+
+import six
 import numpy
 import base64
 
-import skimage
-import skimage.io
-import skimage.transform
-import skimage.filters
-import skimage.exposure
-import skimage.feature
-import skimage.filters.rank
-from skimage.morphology import disk
-
-import PIL
-from PIL import ImageEnhance
-
 try:
-    import io  # python3
-except ImportError:
-    from io import StringIO
+    import skimage
+    import skimage.io
+    import skimage.transform
+    import skimage.filters
+    import skimage.exposure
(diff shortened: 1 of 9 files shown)


sn6uv:
I think you only want 515aa66 and on right?

Rather than making a new pull request to fix this I'd recommend something like:
```
git checkout upstream/imathics
git cherry-pick e796872..8a50e24
git branch -d image          # delete old image branch
git checkout -b image        # create a new image branch
git push -f -u origin image  # force push the image branch over the old one
```

and the PR should update automatically.

poke1024:
cherry pick worked like magic

sn6uv:
The current notebook only supports svg images. I think implementing this for the imathics branch only makes sense. (I'm currently trying to port the graphics/plot functionality over to using jupyter).

mathics/builtin/image.py
```diff
@@ -0,0 +1,736 @@
...
+
+    def apply(self, a, evaluation):
+        'Colorize[a_?MatrixQ]'
+
+        a = numpy.array(a.to_python())
+        n = int(numpy.max(a)) + 1
+        if n > 8192:
+            return evaluation.error('Colorize', 'toomany')
+
+        cmap = matplotlib.cm.get_cmap('hot', n)
```
sn6uv: It might be worth splitting the matplotlib dependency out (separate to scikit-image) if this is the only place it's used.

poke1024: I thought about this as well, but then scikit-image depends on matplotlib (and numpy and scipy), so it's probably not worth the effort:

https://github.com/scikit-image/scikit-image/blob/master/requirements.txt

sn6uv: Good point, I agree.

sn6uv:
Lots of great functionality here!

I'm through this and woking on adding some error handling and docs/tests, see [my `Image` branch](https://github.com/sn6uv/Mathics/tree/Image).

sn6uv:
@poke1024 how do you want to handle collaborating on this? Should I open a PR against your branch with my changes?

poke1024:
I'm actually not sure what's best. I'm actually quite comfortable with doing PRs on your branch (and calling yours the definite version).

sn6uv:
Okay, if you think that's easier. I'll close this and open a separate PR.

PR #366, opened 2016-04-30
Manipulate

sn6uv:
Extension of #362 (since @poke1024 is on vacation).

### DIFF
--- a/mathics/builtin/manipulate.py
+++ b/mathics/builtin/manipulate.py
@@ -209,12 +209,6 @@ class Manipulations(PatternDispatcher):
         self._add_discrete_options_var(i, options, idefault, ilabel)
 
 
-class ListAnimate(Builtin):
-    rules = {
-        'ListAnimate[l_List]': 'Manipulate[Part[l, i], {i, 1, Length[l], 1}]'
-    }
-
-
 class Manipulate(Builtin):
     """
     <dl>
--- a/mathics/builtin/manipulate.py
+++ b/mathics/builtin/manipulate.py
@@ -222,15 +222,17 @@ class Manipulate(Builtin):
         <dd>allows you to interactively compute and display an expression with different argument values.
     </dl>
 
-    >> Manipulate[N[Sin[y]], {y, 1, 20, 2}] //FullForm
-    Null
+    >> Manipulate[N[Sin[y]], {y, 1, 20, 2}]
+     : Manipulate[] only works inside a Jupyter notebook.
+     = Manipulate[N[Sin[y]], {y, 1, 20, 2}]
 
-    Manipulate[i^3, {i, {2, x^4, a}}] //FullForm
-    Null
+    >> Manipulate[i ^ 3, {i, {2, x ^ 4, a}}]
+     : Manipulate[] only works inside a Jupyter notebook.
+     = Manipulate[i ^ 3, {i, {2, x ^ 4, a}}]
 
-    Manipulate[x^y, {x, 1, 20}, {y, 1, 3}] //FullForm
-    Null
-
+    >> Manipulate[x ^ y, {x, 1, 20}, {y, 1, 3}]
+     : Manipulate[] only works inside a Jupyter notebook.
+     = Manipulate[x ^ y, {x, 1, 20}, {y, 1, 3}]
     """
 
     attributes = ('HoldAll',)  # we'll call ReleaseHold at the time of evaluation below
@@ -245,12 +247,10 @@ class Manipulate(Builtin):
         'Manipulate[expr_, args__]'
 
         if (not _jupyter) or (not Kernel.initialized()) or (Kernel.instance() is None):
-            evaluation.error('Manipulate', 'jupyter')
-            return Symbol('Null')
+            return evaluation.message('Manipulate', 'jupyter')
 
         if not _ipywidgets:
-            evaluation.error('Manipulate', 'noipywidget')
-            return Symbol('Null')
+            return evaluation.message('Manipulate', 'noipywidget')
 
         manip = Manipulations(evaluation)  # knows about the arguments and their widgets
 
--- a/mathics/builtin/manipulate.py
+++ b/mathics/builtin/manipulate.py
@@ -232,7 +232,6 @@ class Manipulate(Builtin):
             evaluation.message('Manipulate', 'noipywidget')
             return Symbol('Null')
 
-        try:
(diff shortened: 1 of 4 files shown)


sn6uv:
@poke1024 (when you get back): I'm a bit confused by the need for a new `PatternDispatcher`.

mathics/Makefile
```diff
@@ -1,13 +0,0 @@
-put:
```
bnjones: Not really objecting to the removal of this makefile, but is this change logically related to Manipulate? Would it be better as a separate PR?

mathics/builtin/manipulate.py
```diff
@@ -0,0 +1,314 @@
...
+def _manipulate_label(x):
+    if isinstance(x, String):
+        return x.get_string_value()
+    elif isinstance(x, Symbol):
+        return _strip_namespace(x.get_name())
+    else:
+        return str(x)
+
+
+class PatternDispatcher:
```
bnjones: I think it'd be easier to subclass `object` (see below)

mathics/builtin/manipulate.py
```diff
@@ -0,0 +1,314 @@
...
+class IllegalWidgetArguments(Exception):
+    def __init__(self, var):
+        super(IllegalWidgetArguments, self).__init__()
+        self.var = var
+
+
+class Manipulations(PatternDispatcher):
+    def __init__(self, evaluation):
+        if six.PY2:
+            PatternDispatcher.__init__(self)
```
bnjones: I don't think we need this (even under Python 2) if `PatternDispatcher` inherits from `object`

mathics/builtin/manipulate.py
```diff
@@ -0,0 +1,314 @@
...
+
+    for widget in kwargs_widgets:
+        widget.on_trait_change(call_f, 'value')
+
+    container.on_displayed(lambda _: call_f(None, None, None))
+
+    return container
+
+
+def _strip_namespace(name):
```
bnjones: It looks like this function does the same thing as `mathics.core.expression.strip_context` - maybe we can use that here instead?

I think it makes sense to use this in `PatternDispatcher.dispatch` (because we need valid Python argument names), but for generating the labels `Definitions.shorten_name` might be a better choice - this only shortens the name if it's accessible via `$ContextPath`. I don't have Mathematica around to check but I'd guess this is what it does.

mathics/builtin/manipulate.py
```diff
@@ -0,0 +1,314 @@
...
+            line_no = evaluation.definitions.get_line_no()
+
+            new_evaluation = Evaluation(evaluation.definitions, result_callback=display_data_callback,
+                                        out_callback=evaluation.out_callback)
+
+            vars = [Expression('Set', Symbol(name), value) for name, value in kwargs.items()]
+            evaluatable = Expression('ReleaseHold', Expression('Module', Expression('List', *vars), expr))
+            new_evaluation.evaluate([evaluatable], timeout=settings.TIMEOUT)
+
+            evaluation.definitions.set_line_no(line_no)  # do not increment line_no for manipulate computations
```
bnjones: If we want to implement `Dynamic[]` or scheduled tasks, I can see us doing a lot of evaluation that isn't connected to a specific input line. Maybe we could make updating the line number a feature that can be disabled by something like `Evaluation(..., update_lineno=False)`?

sn6uv:
I'll move the profile and makefile changes to a seperate PR, good point.

poke1024:
@bnjones `mathics.core.expression.strip_context` is exactly the thing I wanted to do with strip_namespace.

@sn6uv `PatternDispatcher` is an ugly hack. I thought about it this morning, and tried to come up with a cleaner solution. It uses one BuiltIn, that's used to parse the Manipulate params, no more custom pattern matching. I also looked into the elegant way Table[] handles these parameter cases, but I'm quite nervous/unsettled about getting the widget instances into the evaluation pipeline (via atoms or expressions) since I saw pickle serialization errors when I tried that once. So I stay away from that. The parsing I came up this morning centers around the following class (it basically generates a rule list/dictionary that can later be used to kwargs-call a python function that creates the widget):

```
class ManipulateParameter(Builtin): # parses one Manipulate[] parameter spec, e.g. {x, 1, 2}, see _WidgetInstantiator
    context = 'System`Private`'

    rules = {
        # detect x and {x, default} and {x, default, label}.
        'System`Private`ManipulateParameter[{s_Symbol, r__}]':
            'System`Private`ManipulateParameter[{Symbol -> s, Label -> s}, {r}]',
        'System`Private`ManipulateParameter[{{s_Symbol, d_}, r__}]':
            'System`Private`ManipulateParameter[{Symbol -> s, Label -> s, Default -> d}, {r}]',
        'System`Private`ManipulateParameter[{{s_Symbol, d_, l_}, r__}]':
            'System`Private`ManipulateParameter[{Symbol -> i, Label -> l, Default -> d}, {r}]',

        # detect different kinds of widgets. on the use of the duplicate key "Default ->", see _WidgetInstantiator.add()
        'System`Private`ManipulateParameter[var_, {min_?RealNumberQ, max_?RealNumberQ}]':
            'Join[{Type -> "Continuous", Minimum -> min, Maximum -> max, Default -> min}, var]',
        'System`Private`ManipulateParameter[var_, {min_?RealNumberQ, max_?RealNumberQ, ds_?RealNumberQ}]':
            'Join[{Type -> "Discrete", Minimum -> min, Maximum -> max, Step -> ds, Default -> min}, var]',
        'System`Private`ManipulateParameter[var_, {opt_List}] /; Length[opt] > 0':
            'Join[{Type -> "Options", Options -> opt, Default -> Part[opt, 1]}, var]'
    }
```

BTW I ran into an ugly error when using Manipulate[] this morning, namely `[IPKernelApp] ERROR | No such comm target registered: jupyter.widget.version`. Obviously this can happen if multiple Jupyter versions are installed on the same user, and their settings/notebook extensions get in conflict (they share `$HOME/.jupyter` and others).  Here's what I did to finally fix it:

(1) call `jupyter -paths` to detect the paths where Jupyter stores data
(2) delete all those paths
(3) reinstall the `imathics` kernel (using its `setup.py`)

I remember you running into this nbwidgets extensions thing, and now I'm starting to wonder whether there's more of these Jupyter issues and a need for a list of workarounds/documentation.

sn6uv:
@poke1024 I've cherry-picked those commits. I think this way is better.

I've merged #369 and rebased this so it's now functional again.

PR #373, opened 2016-05-10
new list and set functions

poke1024:
This implements Gather[], GatherBy[], Tally[], Union[], Intersect[], IntersectingQ[], DisjointQ[].

It also updates the existing implementations of Complement[] and DeleteDuplicates[] so that they work in n log n if SameTest is SameQ (as opposed to the old implementations who where n^2).

It also implements SortBy[] and BinarySearch[].


mathics/builtin/lists.py
```diff
@@ -2096,14 +2099,235 @@ def apply(self, list, sep, evaluation):
...
+    def apply(self, list, test, evaluation):
+        '%(name)s[list_, test_]'
+
+        if list.is_atom():
+            expr = Expression(self.get_name(), list, test)
+            evaluation.message(self.get_name(), 'normal', 1, expr)
+            return
+
+        if list.get_head_name() != 'System`List':
+            evaluation.error(self.get_name(), 'needlist')
```
sn6uv: should be `return evaluation.message(self.get_name(), 'list', 1, expr)`

mathics/builtin/lists.py
```diff
@@ -2096,14 +2099,235 @@ def apply(self, list, sep, evaluation):
...
+
+
+class _GatherOperation(Builtin):
+    rules = {
+        '%(name)s[list_]': '%(name)s[list, SameQ]'
+    }
+
+    messages = {
+        'normal': 'Nonatomic expression expected at position `1` in `2`.',
+        'needlist': 'expecting a List as first parameter.',
```
sn6uv: We try to have compatible messages with MMA, e.g. this line should be ```'list': 'List expected at position `2` in `1`.'```

mathics/builtin/lists.py
```diff
@@ -2096,14 +2099,235 @@ def apply(self, list, sep, evaluation):
...
+        for elem in a_list.leaves:
+            selection = select(elem)
+            for prototype, add_to_bin in selection:  # find suitable bin
+                if same(prototype, elem):
+                    add_to_bin(elem)  # add to existing bin
+                    break
+            else:
+                a_bin = Bin(elem)  # create new bin
+                selection.append((elem, a_bin.add_to))
+                bins.append(a_bin)
```
sn6uv: This is `O(#bins * #elems)` e.g. `O(n^2)` for `n` distinct elements. Could we use the hash from fast equivalence to get this down to `O(#elems)`?

mathics/builtin/structure.py
```diff
@@ -85,6 +88,138 @@ def __gt__(self, other):
...
+        if l.is_atom():
+            evaluation.message('Sort', 'normal')
+        else:
+            keys = Expression('Map', f, l).evaluate(evaluation).leaves  # precompute:
+            # even though our sort function has only (n log n) comparisons, we should
+            # compute f no more than n times.
+
+            raw_keys = l.leaves
+            if len(raw_keys) != len(keys):  # this should never happen, I assume
+                return Symbol('$Aborted')
```
sn6uv: I'd guess this can happen if `f` is not a valid function. We should probably check that `l` is actually a `List` and print an error message + return None if things go wrong.

mathics/builtin/lists.py
poke1024: It should be `O(n^2)` only for the worst case, i.e. only if the hash function fails us, as it's really two nested levels of bins here: in the `_FastEquivalence` case, the data structure built in the code above is actually this: `selects = {hash1: [(prototype1, add_to_bin1)], hash2: [(prototype2, add_to_bin2)], ...}`, with `add_to_bin`s being nothing but appends to lists.

`select(elem)` calls `_FastEquivalence`'s `selector()` function which is `lambda elem: hashes[hash(elem)]`, with `hashes` being a `defaultdict(list)`.

So, `select(elem)` initially returns an empty list for an object. Later, when an object with the same hash comes along, a 1-element list with prototype as shown above is returned. So we might have `selects = {hash1: [(prototype1, add_to_bin1)]}` and an object with a hash of `hash1`, so `select(elem)` returns `(prototype1, add_to_bin1)`.

In this latter case, the for loop is entered, `same` is called (with the existing `prototype` and the new element). If `same` yields `true`, the element gets added to the nth bin, but no new prototype entry is added, i.e. we stay with `selects = {hash1: [(prototype1, add_to_bin1)]}`, only the bin behind `add_to_bin1` got larger one element.

Only if `same` yields false, will a second (or nth) prototype entry be added. Having multiple prototype entries in one hash bucket means that those entries share the same hash value but had `same` return `false` between them. This should be a border case with a proper hash function.

I think I should rename the `selector()` method to `select()` in order to match with the naming above.

mathics/builtin/lists.py
poke1024: I haven't understood the difference between `evaluation.message` and `evaluation.error` yet. Is `message` more like a warning or is it also used for errors? I'm just wondering if the thing above should be an `evaluation.error` after all?

sn6uv: `evaluation.message` acts as both a warning and an error message. It's used whenever evaluation can continue. Normally continuing just means return the expression unevaluated. In this case for example, there is no evaluation 'error', it's just that the function doesn't apply to all inputs.

`evaluation.error` on the other hand is used when there is an error with the evaluation process itself. For example, if the recursion depth gets above `$RecursionLimit` then that's an evaluation error since evaluation can't continue. We only use `evauation.error` when the calculation has to be aborted.

We try to keep things compatible with Mathematica.

Does that make sense?

poke1024: Ok, I think I understand, so it usually be evaluation.message.

poke1024:
I redid the evaluation.message stuff, and I added additional checks in SortBy[] with an error message guarding against bad things happening in the Map[].

mathics/builtin/lists.py
sn6uv: Ah, I understand now. The algorithm I was thinking of is exactly what you've implemented :)

The only thing we need to watch out for is that if the atoms change that hash is still equal when SameQ is.

It wouldn't be a bad idea to have some unit tests for this just to make sure this assumption continues to hold.

sn6uv:
Okay, I'll have a final look over tomorrow but I think this looks good to merge.

mathics/builtin/lists.py
poke1024: I added a basic unit test for the Atoms I could find

sn6uv:
I think a better place for the hash tests would be e.g. `test/test_hash.py`.

poke1024:
Moved there now.  And it runs without explicitly hooking it up. This is nice.

sn6uv:
Some performance tests for `Gather`:
verified_gather_performance
Note the expected O(n) performance when there are both few or many collisions.

mathics/builtin/lists.py
```diff
@@ -2119,38 +2342,115 @@ class DeleteDuplicates(Builtin):
...
+
+    def _elementwise(self, a, b, same):
+        for eb in b:
+            yield eb
+        for ea in a:
+            if not any(same(eb, ea) for eb in b):
+                yield ea
+
+
+class Intersect(_SetOperation):
```
sn6uv: Minor typo: the name of this function is `Intersection`.

mathics/builtin/lists.py
```diff
@@ -2096,14 +2099,234 @@ def apply(self, list, sep, evaluation):
...
+            return self._gather(list, _FastEquivalence())
+        else:
+            return self._gather(list, _SlowEquivalence(test, evaluation, self.get_name()))
+
+    def _gather(self, a_list, equivalence):
+        bins = []
+        Bin = self._bin
+
+        select = equivalence.select()
+        same = equivalence.same()
```
bnjones: Unless I'm missing something, the usage of `.select()` and `.same()` here seems more complicated than it needs to be. I don't think these need to return closures here.

How about refactoring these methods so this function can just call `equivalance.select(elem)` and `equivalence.same(prototype, elem)`? This would also allow `_make_test_pair()` to be folded into `_SlowEquivalence.same()`.

mathics/builtin/lists.py
```diff
@@ -2119,38 +2342,115 @@ class DeleteDuplicates(Builtin):
...
-            if not matched:
-                result.append(leaf)
+        for pos, e in enumerate(seq):
+            if e.is_atom():
+                return evaluation.message(
+                    self.get_name(), 'normal', pos + 1, Expression(self.get_name(), *seq))
 
-        return Expression(mlist.head, *result)
+        for pos, e in enumerate(zip(seq, seq[1:])):
+            e1, e2 = e
```
bnjones: I think this could be written as `for pos, (e1, e2) in ...`

mathics/builtin/lists.py
poke1024: I agree that the code is less readable than it should be here. I'm always concerned about Python name lookups in inner loops, that's why I came up with these lambdas in the first place, but I think the code's much cleaner without them. Changed accordingly.

sn6uv:
I'll do some final tests with this but I think it's good to merge otherwise.

sn6uv:
One minor issue I found with `Intersection`. When a different `SameTest` is used only one representative from each equivalence class should be taken. For example:
```Intersection[{2, -2, 1, 3, 1}, {2, 1, -2, -1}, SameTest -> (Abs[#1] == Abs[#2] &)]```
gives `{-2, 1, 1, 2}` expected `{1, 2}`.
The same issue holds for for `Union`.

PR #374, opened 2016-05-10
small change that yields a 25% performance gain

poke1024:
Before this change:

Timing[Length[Table[i, {i, 1, 10000}]]]
{ 11.398169 , 10000 }

After this change:

Timing[Length[Table[i, {i, 1, 10000}]]]
{ 8.621232000000001 , 10000 }

The gain is related to the __eq__ implementation of Expression using get_sort_key(), which seems to be relatively slow for Symbols.


sn6uv:
Nice one!

bnjones:
I know this has been closed but I just wanted to say: nice job! Thanks for this

PR #378, opened 2016-05-14
Image

sn6uv:
Extends #363.

Still a work in progress.


sn6uv:
The way it's implemented the entire module doesn't get imported if the libs aren't available. This could cause some confusion and maybe there's a better way to handle it. The main reason for doing it this way is that adding the same error message to every function is tedious. An alternative would be to subclass `Builtin` and wrap all the apply methods but it's quite hacky that way.

sn6uv:
Image reflect tests are failing due to #381.

poke1024:
reimplementation of parts of this to only depend on Pillow (PyPy compatible) instead of PIL (not PyPy compatible) depends on #400

PR #385, opened 2016-05-19
An implementation of $OutputSizeLimit

poke1024:


poke1024:
rebased onto current master

sn6uv:
My main objection to this is that it complicates the codebase quite significantly. I think a better place to handle this (admittedly less efficient than your solution) would be in the frontend. In fact, I think this is what MMA does.

poke1024:
What I tried to work around with this is the current inefficiency of the `MakeBoxes` phase, in which `MakeBoxes[x_, form_]` is evaluated for every single element of the output, which really stresses the pattern matcher.

For example, compare the runtimes of `Nest[Partition[#,2]&, Range[10000], 2]` and of the same command wrapped in `Length[]`.  The latter takes `2.1s` on my machine, the former, involving the `MakeBoxes` step, takes `47.3s`. It's a factor 10 to 100 in my experience, which means that Mathics basically currently freezes with large data, even though it can handle it perfectly on the computational side. You don't get these numbers in `Timing`, as `Timing` happens before `MakeBoxes`, but you can try this script:

[boxes.py.zip](https://github.com/mathics/Mathics/files/465705/boxes.py.zip)

So, working with 1000s or 10000s of items is currently not feasible. For example, working with natural language parsing and outputting a text of 100000 words feels like a crash; likewise, importing an XML file with a few thousand elements (as I'm currently doing in https://github.com/poke1024/Mathics/tree/xml), does not work, since the output phase does not finish even after several minutes (though the construction of the `Expression` is done after a second). So currently, XML import, even though it works, is simply not usable, as it freezes the system, as soon as you happen to forget to clip your output.

`$OutputLimitSize` alleviates this by simply not calling `MakeBoxes` for the vast majority of items involved. I agree that the proper solution would be to speed up the pattern matching such that `MakeBoxes` is blazingly fast; there's a lot of lookup happening there for `MakeBoxes` that probably could be optimized (or maybe cythonized). But I fear that this is a very large and difficult change, and as it's not clear, when this might happen, I'm hoping we can find a solution like `OutputSizeLimit` that gives us a solution until that happens.

mathics/core/evaluation.py
```diff
@@ -356,6 +385,66 @@ def get_quiet_messages(self):
...
+        ellipsis = [Omitted('<<%d>>' % ellipsis_size)] if ellipsis_size > 0 else []
+
+        if segment is not None:
+            if ellipsis_size > 0:
+                segment.extend((True, len(left_leaves), len(items) - len(right_leaves)))
+            else:
+                segment.extend((False, 0, 0))
+
+        return list(itertools.chain(left_leaves, ellipsis, reversed(right_leaves)))
+
```
sn6uv: This implementation is quite cool.

sn6uv:
Sorry it took me so long to get to this.

PR #386, opened 2016-05-20
Graphics boxes for colors

poke1024:
Adds little color boxes to colors (i.e. classes derived from `_Color`) like `RGBColor[0, 1, 1]` in `StandardForm`.

This should have been a fairly simple format change, but I ran into obstacles with my new unit test, that should be testing that `RGBColor[0, 1, 0] // StandardForm` is now `-Graphics-`.

The test case didn't work, it just ignored my `//StandardForm` direction. And this was due to some special case handling in `mathics/core/expression.py` involving `OutputForm` and `StandardForm` (that I don't understand). Disabling it (as below) fixes my new test case, but breaks two old test cases, namely: `ToBoxes[a + a]` and `StandardForm[a + b * c]`. Both of these test `StandardForm` expressions (`ToBoxes` defaults to `StandardForm` if no second parameter is given), and they failed, because their test cases expected the `Times` unicode dot symbol. I checked with MMA and I don't think `StandardForm` should have a `Times` dot symbol anywhere. So I changed these test cases to map `Times` to space in `StandardForm` which seems to be the right thing to me. On the other hand, I don't understand what's going on here with this special case I disabled.

I extended `ImageSize` for `Graphics` to allow a width or a {width, height} specification, which was not possible before. The whole `xmax, xmin` handling in `mathics/builtin/graphics.py` seems quite sloppy in regard to `None` values (looks to me like they subtract None values and then return it as int), so I added an `if` there.

The change in `mathics/builtin/base.py` is just about allowing `%(name)s` in format patterns.

### DIFF
--- a/mathics/builtin/graphics.py
+++ b/mathics/builtin/graphics.py
@@ -242,7 +242,7 @@ class _Color(_GraphicsElement):
         # way in the notebook, so we restrict the rule to StandardForm.
 
         (('StandardForm', ), '%(name)s[x__]'):
-            'Style[Graphics[{EdgeForm[Black], %(name)s[{x}], Rectangle[]}, ImageSize -> 16], ' +
+            'Style[Graphics[{EdgeForm[Black], %(name)s[x], Rectangle[]}, ImageSize -> 16], ' +
             'ImageSizeMultipliers -> {1, 1}]'
     }
 
--- a/mathics/builtin/graphics.py
+++ b/mathics/builtin/graphics.py
@@ -241,7 +241,7 @@ class _Color(_GraphicsElement):
         # way in the notebook, so we restrict the rule to StandardForm.
 
         (('StandardForm', ), '%(name)s[x__]'):
-            'Style[Graphics[{EdgeForm[Black], x, Rectangle[]}, ImageSize -> 16], ' +
+            'Style[Graphics[{EdgeForm[Black], %(name)s[{x}], Rectangle[]}, ImageSize -> 16], ' +
             'ImageSizeMultipliers -> {1, 1}]'
     }
 
--- a/mathics/builtin/graphics.py
+++ b/mathics/builtin/graphics.py
@@ -235,6 +235,16 @@ class _GraphicsElement(InstancableBuiltin):
 
 
 class _Color(_GraphicsElement):
+    formats = {
+        # we are adding ImageSizeMultipliers in the rule below, because we do _not_ want color boxes to
+        # diminish in size when they appear in lists or rows. we only want the display of colors this
+        # way in the notebook, so we restrict the rule to StandardForm.
+
+        (('StandardForm', ), '%(name)s[x__]'):
+            'Style[Graphics[{EdgeForm[Black], x, Rectangle[]}, ImageSize -> 16], ' +
+            'ImageSizeMultipliers -> {1, 1}]'
+    }
+
     components_sizes = []
     default_components = []
 
@@ -297,6 +307,12 @@ class RGBColor(_Color):
     </dl>
 
     >> Graphics[MapIndexed[{RGBColor @@ #1, Disk[2*#2 ~Join~ {0}]} &, IdentityMatrix[3]], ImageSize->Small]
+     = -Graphics-
+
+    >> RGBColor[0, 1, 0]
+     = RGBColor[0, 1, 0]
+
+    >> RGBColor[0, 1, 0] // StandardForm
      = -Graphics-
     """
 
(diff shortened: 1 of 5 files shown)


mathics/builtin/graphics.py
```diff
@@ -1268,14 +1284,24 @@ def _get_image_size(self, options, graphics_options, max_width):
...
-        base_width, base_height = {
-            'System`Automatic': (400, 350),
-            'System`Tiny': (100, 100),
-            'System`Small': (200, 200),
-            'System`Medium': (400, 350),
-            'System`Large': (600, 500),
-        }.get(image_size, (None, None))
+        try:
+            if image_size.is_numeric():
+                base_width = int(image_size.to_number())
```
sn6uv: Possible minor bug here: If image_size is `Symbol(Pi)` for example. I'd say it's not worrying about non-integer inputs so checking `isinstnace(image_size, Integer)` on the line above and then using `get_int_value()` would be fine.

mathics/builtin/graphics.py
```diff
@@ -1268,14 +1284,24 @@ def _get_image_size(self, options, graphics_options, max_width):
...
+                base_height = None  # will be computed later in calc_dimensions
+            elif image_size.get_head_name() == 'System`List':
+                base_width, base_height = ([x.to_number() for x in image_size.leaves] + [0, 0])[:2]
```
sn6uv: To avoid length checking here you can use `elif image_size.has_form('List', 2)`.

mathics/builtin/inout.py
```diff
@@ -385,7 +385,7 @@ class ToBoxes(Builtin):
 
     Unlike 'MakeBoxes', 'ToBoxes' evaluates its argument:
     >> ToBoxes[a + a]
-     = RowBox[{2, \u2062, a}]
+     = RowBox[{2,  , a}]
```
sn6uv: I thought this might be related to the web frontend but actually it looks better with the space I think.

sn6uv:
:+1: from me, looks great:
2016-07-29-120408_2560x1440_scrot

My only criticism is that RGB doesn't handle lists, i.e. `RGBColor[{1, 0, 0}]` but this isn't a blocker for me.

poke1024:
This was actually a bug in the pattern. `RGBColor[{1, 1, 0}]` works now (as well).

mathics/builtin/graphics.py
```diff
@@ -243,6 +244,16 @@ def init(self, graphics, item=None, style=None):
 
 
 class _Color(_GraphicsElement):
+    formats = {
+        # we are adding ImageSizeMultipliers in the rule below, because we do _not_ want color boxes to
+        # diminish in size when they appear in lists or rows. we only want the display of colors this
+        # way in the notebook, so we restrict the rule to StandardForm.
+
+        (('StandardForm', ), '%(name)s[x__]'):
```
bnjones: Maybe this pattern should be more specific - something like `x__?NumericQ`? Otherwise entering a colour with symbolic components raises a box error and we don't get any output.

To make this pattern easier to write I'd suggest a separate rule mapping `RGBColor[{r,g,b}]` to `RGBColor[r,g,b]`, so you don't have to handle the list case in this pattern. If this works, the code in `_Color.init` to handle the list case could be removed.

mathics/core/expression.py
```diff
@@ -243,7 +243,8 @@ def do_format(self, evaluation, form):
             include_form = False
             if head in formats and len(self.get_leaves()) == 1:
                 expr = self.leaves[0]
-                if not (form == 'System`OutputForm' and head == 'System`StandardForm'):
+                # removing 'True' below breaks the 'RGBColor[1,0,0]//StandardForm' test cases.
+                if True or not (form == 'System`OutputForm' and head == 'System`StandardForm'):
```
bnjones: Not sure if this was meant to make it into the PR, I think if you're removing the condition completely it'd be better to just remove it.

bnjones:
This makes sense:
```
In[1]:= StandardForm[RGBColor[1,2,3]]
Out[1]= -Graphics-
```
but I'm not sure about this - was this intentional?
```
In[2]:= StandardForm[x^2]
Out[2]= Superscript[x, 2]
```

mathics/core/expression.py
bnjones: Thinking about this a bit more, does Mathematica show "`-Graphics-`" when you do `StandardForm[RGBColor[1, 0, 0]]` in its console interface?

I think if we're just doing this for the test, another way to test this might be to use `ToBoxes[RGBColor[1, 0, 0]]` instead of `StandardForm`. If you do this the test would also fail if the box form changes.

poke1024: Evaluating `StandardForm[RGBColor[1, 0, 0]]` in Mathematica's console interface actually prints out a complex bunch of things, which contains, among other things, a `GraphicsBox`.

poke1024:
No, the new output for `StandardForm[x^2]` wasn't intentional, and it turns out, this is what the code in `mathics/core/expression.py` is for, that I disabled using `True` for color boxes. So I undid that change there (turned it into the original state). The color box output in the web console still works, so I'll guess I'll just change the test cases to use `ToBoxes` as you proposed and everything should be fine.

poke1024:
`StandardForm[x^2]` is fixed again, and colors are now only converted to `Graphics` if their values are numeric and within proper limits. If there's nothing else, this is ready to merge from my side.

PR #387, opened 2016-05-21
Hash[]

poke1024:
Mathics currently has FileHash[], but no Hash[]. This PR moves the functionality in FileHash[] to Hash[] and adds a user_hash to all BaseExpression derived classes so they know how to hash themselves with various hash algorithms. Hashing a string produces the original untampered hash, as it does in MMA.


sn6uv:
Great!

PR #388, opened 2016-05-21
RandomChoice[] and RandomSample[]

poke1024:
Adds RandomChoice[] and RandomSample[]. These two need numpy installed.

I rewired RandomInteger[] et al. to numpy (if it's installed), as there should be one consistent random number provider for the system that attaches to one random generator state, which can be set, reset, etc.. Starting to have different random generator states for e.g. RandomInteger[] and RandomChoice[] (and ImageRandom[] and others in the future for that matter) would surely spell disaster. So if numpy is available, everything is based on the numpy generator now.

If numpy is not installed (e.g. for PyPy) the implementation falls back and is equivalent in its results to the old implementation. Then, however, RandomChoice[] and RandomSample[] are not available. They could be added through adding a python only implementation in NoNumPyRandomEnv.randchoice().

I adjusted setup.py to pull the PyPy-numpy package for PyPy (https://bitbucket.org/pypy/numpy). The install works and downloads the package, but this then says that it needs PyPy4 to run, which is currently not supported by TravisCI. So I commented it out for now. It might be an option for the future though.

### DIFF
--- a/mathics/builtin/randomnumbers.py
+++ b/mathics/builtin/randomnumbers.py
@@ -673,3 +673,8 @@ class RandomSample(_RandomSelection):
      """
 
     _replace = False
+
+if not _numpy:  # hide symbols from non-numpy envs
+    _RandomSelection = None
+    RandomChoice = None
+    RandomSample = None
--- a/mathics/builtin/randomnumbers.py
+++ b/mathics/builtin/randomnumbers.py
@@ -16,6 +16,8 @@ import six.moves.cPickle as pickle
 import binascii
 import hashlib
 import numpy
+import time
+import os
 
 from mathics.builtin.base import Builtin
 from mathics.core.expression import (Integer, String, Symbol, Real, Expression,
@@ -75,7 +77,8 @@ class RandomEnv:
         return numpy.random.choice(n, size=size, replace=replace, p=p)
 
     def seed(self, x=None):
-        # numpy implementation might be different from the old python impl.
+        if x is None:  # numpy does not know to seed itself randomly
+            x = int(time.time() * 1000) ^ hash(os.urandom(16))
         # for numpy, seed must be convertible to 32 bit unsigned integer.
         numpy.random.seed(abs(x) & 0xffffffff)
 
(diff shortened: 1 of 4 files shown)


mathics/builtin/randomnumbers.py
```diff
@@ -33,17 +65,60 @@ def get_random_state():
...
+            leaves = [new_element(x) for x in a]
+        else:
+            leaves = [_from_array(e, new_element, d) for e in a]
+        return Expression('List', *leaves)
+
+    def _stack_array(a):
+        if not isinstance(a[0], list):
+            return list(chain(a))
+        else:
+            return [_stack_array([x[i] for x in a]) for i in range(len(a[0]))]
```
sn6uv: I wonder if these functions might be useful for other things that could use numpy, e.g. linear algebra routines. There's a short discussion on the mailing list see https://groups.google.com/forum/#!topic/mathics-devel/6btOa-5q87E but I think the best way to proceed would be a separate module e.g. `mathics/builtin/numpy_utils.py`. What do you think?

poke1024: Sounds like a good idea. I'll move them.

sn6uv:
In regards to the TravisCI config. Afaik pypy3 refers to the version of pypy supporting python 3. I think it would be better to try to install numpy under the standard version of pypy (and mark that as an expected failure).

poke1024:
The Travis config now uses a new PYPY_NUMPY flag, and has an expected fail for pypy + PYPY_NUMPY.

mathics/builtin/randomnumbers.py
```diff
@@ -13,17 +13,60 @@
...
+    _numpy = False
+    import random
+
+if _numpy:
+    import time
+    import os
+
+    # mathics.builtin.__init__.py module scanning logic gets confused
+    # if we assign numpy.random.get_state to a variable here. so we
+    # use defs to safely wrap the offending objects.
```
bnjones: Can we fix this in the builtin loader, rather than working around it like this?

PR #389, opened 2016-05-23
shortcut rule lookup

poke1024:
Mathics should quit early when looking rules: if a matching rule has been found and applied, no further rule lookups need to happen.

This seems to improve performance by 10% generally, might be much more dramatic in certain cases.


sn6uv:
Looks like a good improvement.

PR #390, opened 2016-05-23
avoid unwanted reevaluations

poke1024:
Mathics currently reevaluates variable values (i.e. values of symbols set via Set[]) every time they're accessed.

This can lead to worrisome performance:

`x=Range[10000]; Timing[First[x]]`
`{0.2507470000000005,1}`

This PR fixes this by not rebuilding expressions if no var is replaced. That way, the original expression's `is_evaluated` attribute does not get reset, and the expression is thus not reevaluated when `Expression.evaluate()` gets called for it later on. With change in PR:

`x=Range[10000]; Timing[First[x]]`
`{0.08305200000000035,1}`

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -1103,6 +1103,9 @@ class Expression(BaseExpression):
                     leaves = [Expression('List', *func_params), body] + \
                         self.leaves[2:]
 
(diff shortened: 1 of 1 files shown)


poke1024:
Closing this, as it's not good yet.

PR #391, opened 2016-05-23
No reevaluation of evaluated values

poke1024:
Mathics currently reevaluates variable values (i.e. values of symbols set via Set[]) every time they're accessed.

This can lead to some worrisome performance (and it gets worse for even longer lists):

`x=Range[10000]; Timing[First[x]]`
`{0.2507470000000005,1}`

This PR fixes this by not rebuilding expressions if no var is replaced. That way, the original expression's is_evaluated attribute does not get reset, and the expression is thus not reevaluated when Expression.evaluate() gets called for it later on. With change in PR:

`x=Range[10000]; Timing[First[x]]`
`{0.08288399999999996,1}`

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -1113,7 +1113,7 @@ class Expression(BaseExpression):
                     leaves = [Expression('List', *func_params), body] + \
                         self.leaves[2:]
 
-        if len(vars) == 0:  # might just be a symbol set via Set[] we looked up here
+        if not vars:  # might just be a symbol set via Set[] we looked up here
             return self.shallow_copy()
 
         return Expression(
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -492,6 +492,16 @@ class Expression(BaseExpression):
         result.original = self
         return result
 
+    def shallow_copy(self):
+        # this is a minimal, shallow copy: head, leaves are shared with
+        # the original, only the Expression instance is new. we transfer
+        # the is_evaluated state, so we don't reevaluate evaluated stuff.
+        expr = Expression(self.head)
+        expr.leaves = self.leaves
+        expr.options = self.options
+        expr.is_evaluated = self.is_evaluated
+        return expr
+
     def set_positions(self, position=None):
         self.position = position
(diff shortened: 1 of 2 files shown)


bnjones:
Can you run `benchmark.py` before/after to see if this improves performance across the board?

mathics/core/expression.py
```diff
@@ -1103,6 +1113,9 @@ def replace_vars(self, vars, options=None,
                     leaves = [Expression('List', *func_params), body] + \
                         self.leaves[2:]
 
+        if len(vars) == 0:  # might just be a symbol set via Set[] we looked up here
```
bnjones: Why `len(vars) == 0` instead of `not vars`?

mathics/core/rules.py
```diff
@@ -94,6 +94,8 @@ def __init__(self, pattern, replace, system=False):
     def do_replace(self, vars, options, evaluation):
         new = self.replace.replace_vars(vars)
         new.options = options
+        if len(options) > 0:  # need to trigger (re)evaluation?
```
bnjones: If I'm reading this correctly, this will happen whenever we do a replacement for a rule that contains an `OptionsPattern`. How come we're not applying the optimisation in this case?

mathics/core/expression.py
poke1024: Good point, I wasn't even aware that `not` could be used in this way on a dict here. Changed that.

poke1024: `benchmark.py` seems to give a very slight across the board improvement of 2-4%, but it could be as well random noise. I think it really only affects the special case of looking up a symbol that's expensive to compute.

mathics/core/rules.py
poke1024: Yes, that's correct, I added a comment in that code, trying to explain what happens there. It's related to the way `OptionValue[]`s are resolved: it happens through applying `OptionValue.apply` and that needs `evaluate()` to get called on the expression in the first place. Another way to fix this might be to change `if self.is_evaluated` in `Expression.evaluate()` to  `if self.is_evaluated and not self.options`, in order to always evaluate if options need to be applied to the expression, but I'm much less confident in the side effects this might have.

PR #395, opened 2016-05-27
Proposal to fix Plus and Times functionality

poke1024:
This is a proposal to assume that `Plus[]` and `Times[]` always work the same and cannot be overriden by users. The advantage would be that Python code doing basic calculations wouldn't need to go through the whole pattern lookup engine each time an addition or multiplication (or subtraction for that matter) is done.

I'm currently doing some experiments on implementing `FindClusters[]` and calling `Expression('Plus', ...)` each time an addition is done feels like overkill. Directly calling a plus function as proposed in this PR (see `mathics.builtin.arithmetic.plus()`) would seem so much more efficient.

Investigating this, I found that MMA seems to protect Plus and Times no matter what, i.e. calling `Unprotect[Plus]` or `Unprotect[Times]` has no effect, and each attempt to define new rules will be rejected. In these cases, `Unprotect` is a noop. I mimicked this with this PR (see change in `definitions.py`).

Regarding the changes in `arithmetic.py`: I just pulled out the code from `apply` into separate functions.

Python code could now import `plus` and do something like `plus(evaluation, a, b, Integer(1))`.

### DIFF
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -246,8 +246,8 @@ def power(evaluation, *items_sequence):
             evaluation.message('Power', 'infy')
             return Symbol('ComplexInfinity')
     else:
-        numerified_items = [item.numerify(evaluation) for item in items_sequence]
-        return Expression('Power', *numerified_items)
+        numerified_items = Expression('System`Sequence', *items_sequence).numerify(evaluation)
+        return Expression('Power', *numerified_items.get_sequence())
 
 
 def divide(x, y, evaluation):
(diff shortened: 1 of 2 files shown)


sn6uv:
Another possibility would be to implement `__add__ `and `__mul__` for `Expression` so that we can use Python syntax.

poke1024:
Which would be in accordance to the comparison operators already implemented through `KeyComparable`. One technical problem though would be to get access to the current `Evaluation` in the operator code somehow, as it's not readily available (it's not needed for the comparisons implemented in `KeyComparable`). I'll think about this some more.

poke1024:
I added the most common magic methods to BaseExpression and added a test case for this which also illustrates how this would be used.

poke1024:
Closing this, as the assumption that Plus can never be overriden seems to be false

PR #397, opened 2016-05-28
IntegerString, FromDigits, and more

poke1024:


mathics/builtin/integer.py
```diff
@@ -136,3 +136,338 @@ def apply(self, n, b, evaluation):
...
+     = 7
+    >> BitLength[-5]
+     = 3
+    >> BitLength[0]
+     = 0
+    """
+
+    def apply(self, n, evaluation):
+        'BitLength[n_Integer]'
+        return Integer(n.get_int_value().bit_length())
```
sn6uv: This implements what I would say is the expected behaviour. However, mathematica behaves strangely for negative ints, see https://reference.wolfram.com/language/ref/BitLength.html.

In short, I think a correct implementation is
```
n = n.get_int_value()
if n < 0:
    n = -1 - n
return Integer(n.bit_length())
```

so that e.g. `BitLength[-1]` is `0`.

mathics/builtin/integer.py
```diff
@@ -136,3 +136,338 @@ def apply(self, n, b, evaluation):
...
+    rules = {
+        'IntegerString[n_Integer]': 'IntegerString[n, 10]',
+    }
+
+    messages = {
+        'basf': 'Base `` must be an integer in the range from 2 to 36.',
+    }
+
+    list_of_symbols = [chr(i + ord('0')) for i in range(10)] +\
+                      [chr(i + ord('a')) for i in range(26)]
```
sn6uv: Perhaps clearer as `string.digits + string.ascii_letters`.

mathics/builtin/integer.py
```diff
@@ -136,3 +136,338 @@ def apply(self, n, b, evaluation):
...
+    <dt>'IntegerReverse[$n$]'
+        <dd>returns the integer that has the reverse decimal representation of $x$ without sign.
+    <dt>'IntegerReverse[$n$, $b$]'
+        <dd>returns the integer that has the reverse base $b$ represenation of $x$ without sign.
+    </dl>
+
+    >> IntegerReverse[1234]
+     = 4321
+    >> IntegerReverse[1022, 2]
+     = 511
```
sn6uv: can you add a test with sign, e.g. `IntegerReverse[-123]`.

mathics/builtin/integer.py
poke1024: This is neat (and much clearer) indeed.

mathics/builtin/integer.py
poke1024: added.

PR #401, opened 2016-06-06
prevent unnecessary name lookup during context search

sn6uv:
Minor speed improvement for parser. @bnjones does this look okay to you?


bnjones:
Looks good to me! Depending on your view on asserts I'd suggest adding an `assert fully_qualified_symbol_name(name)` to `have_qualified_definition`.

sn6uv:
@bnjones good suggestion, I just added that. Will merge once the tests pass.

PR #410, opened 2016-06-11
Median[], RankedMin[], RankedMax[]

poke1024:
This introduces linear time selection capabilities based on introselect (https://en.wikipedia.org/wiki/Introselect).

I implemented introselect from scratch, since the only two implementations I'm aware of (one in C++'s STL and one inside numpy's argpartition) seem quite cryptic to me in regards to certain steps, and - apart from licensing issues - a Python port of these would have been even more cryptic. For the most part, I just implemented Hoare's old algorithms as described in the 1960 ACM publication without any changes, and, indeed, they work.

Having a reliable median implementation is an important stepping stone for implementing something like kdtrees later on.

The new package "algorithm" is of course just a proposal: for functions implementing basic algorithms and data structures that are too bulky (and too reliant on private functions) to be included in the regular files.

### DIFF
--- a/mathics/algorithm/introselect.py
+++ b/mathics/algorithm/introselect.py
@@ -137,28 +137,32 @@ def introselect(a, k):  # changes a
 if __name__ == "__main__":
     import random
 
-    def test_algorithm(l, name, f):
-        a = [random.randint(-1000, 1000) for _ in range(l)]
+    def test_algorithm(l, r_max, name, f):
+        a = [random.randint(-r_max, r_max) for _ in range(l)]
         b = sorted(a)
         c = [f(a[:], i) for i in range(len(a))]
         if b == c:
-            print('OK %s %d' % (name, l))
+            print('OK %s r: %d l: %d' % (name, r_max, l))
             return True
         else:
-            print('FAIL %s %d' % (name, l))
+            print('FAIL %s r: %d l: %d' % (name, r_max, l))
             print(a, b, c)
             return False
 
-    def test_range(l_max):
+    def test_configuration(l_max, r_max):
         for l in range(l_max):
-            if not test_algorithm(l, 'bfprt', bfprt):
+            if not test_algorithm(l, r_max, 'bfprt', bfprt):
                 return False
 
         for l in range(l_max):
-            if not test_algorithm(l, 'introselect', introselect):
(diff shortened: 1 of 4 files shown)


mathics/algorithm/introselect.py
```diff
@@ -0,0 +1,170 @@
...
+        return True
+
+    def test_range(l_max):
+        # we test two cases: many same elements, and few same elements.
+        return test_configuration(l_max, 10) and test_configuration(200, 1000)
+
+    if test_range(200):
+        print('ALL OK.')
+    else:
+        print('ABORTED WITH FAILURE.')
```
sn6uv: I think a better place for these tests would be e.g. `test/algorithm/test_introselect.py`.

sn6uv:
Good stuff. Did some local testing and it is indeed faster for large lists:
```
In [11]: a = [random.randint(1, 10**5) for _ in range(10**6)]

In [12]: %timeit introselect(a, 500)
1 loop, best of 3: 366 ms per loop

In [13]: %timeit sorted(a)[500]
1 loop, best of 3: 630 ms per loop
```

mathics/builtin/lists.py
```diff
@@ -2696,3 +2697,113 @@ def apply(self, expr, levels, evaluation):
...
+    >> Median[{-11, 38, 501, 1183}]
+     = 539 / 2
+
+    Passing a matrix returns the medians of the respective columns:
+    >> Median[{{100, 1, 10, 50}, {-1, 1, -2, 2}}]
+     = {99 / 2, 1, 4, 26}
+    """
+
+    messages = {
+        'rectn': 'Expected a list of numbers or a list of lists at position 1 in ``.'
```
sn6uv: The mathematica `Medain::rectn` message is ```Rectangular array of real numbers is expected at position `2` in `1`.```, I think we should say something about requiring the input to be rectangular.

poke1024:
The `%timeit` results are great to see. The differences in a Mathics context should be even bigger since comparing `BaseExpression`s is expensive (in contrast to comparing integers) and so the smaller number of comparisons (in contrast to `sort`) should have an even bigger effect.

poke1024:
I've changed `_median5` to only do 6 comparisons; this is only used in a border case, but who knows...

poke1024:
Yes, ready to merge as far as I'm concerned.

Issue #381, opened 2016-05-15
Equal only works as a binary operator


sn6uv:
Previously Equal was a binary operator so this wasn't discovered. Now that `Equal` is correctly parsed this bug is easier to trigger.

Example
-----------
```
>> "a" == "a" == "a"
"a" == "a" == "a"
```
Expected `True`

Also affects `Unequal` and possibly others.

sn6uv:
This is related to #260.

sn6uv:
See also http://mathematica.stackexchange.com/questions/119775/chains-of-approximate-numeric-equalities

Issue #365, opened 2016-04-30
making changes to the parser is tedious


sn6uv:
Every time the parser is changed the rules must be regenerated. This takes a considerable time.

We should at least generate a warning when a  table regeneration is required (may require monkey patching ply).

sn6uv:
fixed by #416

Issue #260, opened 2015-06-19
Logical expressions behave inconsistently/incorrectly


latkin:
Encountered this via web usage, have not tried with a local client.  Chrome 43.0.2357.124

Here is a minimal example of equivalent logical expressions behaving inconsistently:

![image](https://cloud.githubusercontent.com/assets/5943573/8258338/fc9f9f0e-1668-11e5-9159-cd5cd05f4ab5.png)

Here is a slightly more complex example, showing logical expressions behaving very inconsistently.  The final example looks to be completely wrong.

![image](https://cloud.githubusercontent.com/assets/5943573/8258469/ed1a839a-1669-11e5-851a-cc76dc13c485.png)

sn6uv:
Thanks for reporting this.

This looks like a combination of two things to me. Equal is flattening it's arguments when it shouldn't:
```Mathematica
In[1]:= Hold[Equal[0, 0] == True] // FullForm
Out[1]= Hold[Equal[0, 0, True]]
```
expected
```Mathematica
Hold[Equal[Equal[0, 0], True]]
```
The second one is a bug in the evaluation of the flattened arguments
```Mathematica
In[2]:= Equal[0, 0, True]
Out[2]= 0 == 0 == True
```
expected
```Mathematica
0 == True
```

The second example is similar,
```Mathematica
In[3]:= Hold[(0 == 0) != (2 == 0)] // FullForm
Out[3]= Hold[Inequality[0, Equal, 0, Unequal, Equal[2, 0]]]
```
expected
```Mathematica
Hold[Unequal[Equal[0, 0], Equal[2, 0]]]
```
which also evaluates incorrectly
```Mathematica
In[4]:= Unequal[Equal[0, 0], Equal[2, 0]]
Out[4]= True != False
```
expected
```Mathematica
False
```

sn6uv:
The parsing issue actually affects all flat operators, e.g. `Plus[a, b] + c` parses as `Plus[a, b, c]` which is wrong.

PR #416, opened 2016-06-16
Recursive descent parser

sn6uv:
## Fixes:
 - #365
 - #260 (but not #381)
 - `Optional`/`Pattern` parsing
 - many other minor parser bugs
 - no more 15mins of silence when rebuilding the tables (when upgrading ply)

## Benefits
* `O(lines)` runtime for multi-line input, previously `O(lines^2)` with 'add a line and retry' approach.
* Parser does not rely on Mathics core so it can be reused for other projects / other purposes by defining additional AST converters.
* Better performance (see below)
* Better syntax errors (see below)

## Downsides
* More complex code (800 lines net increase but about 500 is tests)
* Possible bugs in the immature parser (I think there are actually fewer net bugs)
* More complex `LineFeed` interface for the parser
* We now have both syntax trees and expression trees

## Dependencies
Drops `ply` as a requirement.

## Benchmarks
| test | `master` | `parser` | MMA
|------|------------|------------|------ 
`1+2+...+300` | 11.8ms | 8.01ms |
`1;2;...;300` | 10.5ms | 7.47ms |
`1/2/.../300` | 24.2ms | 14.9ms |
`1^2^...^300` | 16.9ms | 11.0ms |
`! ! ... ! expr` | 10.4ms | 6.63ms |
` expr & & ... &` | 11.1ms | 6.03ms |
` Sin[Sin[...]]` | 29.6ms | 15.1ms |
``Needs["Combinatorica`"]`` | 34.4s | 3.81s | 0.243s
Launch kernel | 2.20s | 1.97s | 0.450s

Performance is marginally better but the main gains are for multi-line input. 

## Better syntax errors
Syntax errors now report more useful information including the line number.
```
$> cat hello.m
Print["Hello World!"]

x * * y
$> mathics -script hello.m
Hello World!
 : "x * " cannot be followed by "* y" (line 3 of "hello.m").
 ```

```
$> mathics -q
In[1]:= Sin[
        
 : Incomplete expression; more input is needed (line 1 of "<stdin>").
In[1]:= x * * y
 : "x * " cannot be followed by "* y" (line 1 of "<stdin>").
In[1]:= 
```

### CODE REVIEW

File: test/test_parser/test_parser.py [hunk 1]
```diff
@@ -0,0 +1,665 @@
...
+        self.check('a ^ b ^ c', 'a ^ (b ^ c)')
+
+    def test_flat(self):
+        self.check('a + b + c', 'Plus[a, b, c]')
+
+    def test_left(self):
+        self.check('a /; b /; c', '(a /; b) /; c')
+
+    def test_Subtract(self):
+        self.check('a - b - c', 'Plus[a, Times[-1, b], Times[-1, c]]')
```
wolfv: Just curious as to why `-` is parsed as plus-times-minus :)

sn6uv: That's a good question. The behaviour is correct, see the [wolfram language spec](https://reference.wolfram.com/language/tutorial/OperatorInputForms.html), and it's also what MMA does.

As to why this is part of the language I'm not completely sure but I have some ideas:
* Converting `Subtract[a, b]` to `Plus[a, Times[-1, b]]` saves implementing `Subtract.
* `Plus` and `Times` are flat operators and can, in principle, be multi-threaded. 
* Doing the conversion at parse time prevents the AST from getting too tall (reducing call stack depths when walking the tree).

For interest, something similar happens with `Divide`; `a / b` is converted to `Times[a, Power[b, -1]]`.

wolfv: Interesting stuff :) Thanks!

sn6uv:
I'm writing up some docs on this (so we know what's going on in future). I think functionality-wise this is finished though. Since this is a pretty major change I want to do a bit more testing before merging though.

poke1024:
very cool :thumbsup:

PR #419, opened 2016-06-26
Improve support for gamma functions

bnjones:
SymPy exposes separate `gamma()` and `uppergamma()` functions, whereas in Mathematica `Gamma[z]` is the gamma function and `Gamma[z, x]` is the upper incomplete gamma function.

Add support for `Gamma[z, x]` plus plumbing for mapping multiple SymPy functions onto one builtin. Right now I think it's just this one function that needs this kind of treatment, but it'd easy to factor out if there are more.


sn6uv:
Looks good.

PR #424, opened 2016-06-30
Variance[] and StandardDeviation[]

poke1024:


sn6uv:
closed and open to trigger Travis build

sn6uv:
I wonder if we can even avoid the rectangle class by adding the `Listable` attribute. The error checking might not be exactly what we want.

Looks good either way.

poke1024:
Adding `Listable`, we'd have to work around breaking the common case `Variance[{a, b, c}]` and for the matrix case it would need to do some transpose, as we need `Variance[{a1, a2}, {b1, b2}]` to become `Variance[{a1, b1}], Variance[{a2, b2}]`. We could add a new attribute `Rectangular`, but then that would be visible to users, which would be strange. Then there's the different error messages `rectn` for `Median` and `rectt` for `Variance`, we have to deal with. 

I thought about making a `rule` that does the transform for the matrix case, but then all the rules, including `apply`, need to guard their cases, so they don't intersect (e.g. by adding `Length[list] >= 1 && Not[MatchQ[list, List[__list]]]` or the like), which gets pretty ugly.

poke1024:
added a `_NotRectangularException`, I think this is cleaner now

sn6uv:
I agree, `Listable` doesn't quite fit here. Are you finished editing this?

poke1024:
Yes, done.

PR #426, opened 2016-06-30
Basic stub

Piruzzolo:
There is a Distribution class with the standard interface (CDF, PDF, probability, expectation...) and distributions are implemented as subclasses. Each distribution has various attributes, among which the assumptions on the parameters and the prototype. The huge lists in the file will serve as a source to copy from. Each Distribution must have its proper number of parameters, otherwise will throw the argr message. When applying some stats function to them (e.g. CDF[]), the correctness of the parameters must be checked. To do this, I think we could take the string "arg_assumptions" (which is a logical combination of the parameters in Mathematica format), replace the occurrences of the slot symbols with actual symbols passed in the instance, and then evaluate the resulting string and check if it equals "System'True". This idea of replancing stuff in that string might also be useful when implementing DistributionParameterAssumptions.


sn6uv:
Closing on @Piruzzolo's request.

PR #427, opened 2016-06-30
Statistics

Piruzzolo:
There is a Distribution class with the standard interface (CDF, PDF, probability, expectation...) and distributions are implemented as subclasses. Each distribution has various attributes, among which the assumptions on the parameters and the prototype. The huge lists in the file will serve as a source to copy from. Each Distribution must have its proper number of parameters, otherwise will throw the argr message. When applying some stats function to them (e.g. CDF[]), the correctness of the parameters must be checked. To do this, I think we could take the string "arg_assumptions" (which is a logical combination of the parameters in Mathematica format), replace the occurrences of the slot symbols with actual symbols passed in the instance, and then evaluate the resulting string and check if it equals "System'True". This idea of replancing stuff in that string might also be useful when implementing DistributionParameterAssumptions.

### DIFF
--- a/mathics/builtin/combinatorial.py
+++ b/mathics/builtin/combinatorial.py
@@ -266,3 +266,923 @@ class RogersTanimotoDissimilarity(_BooleanDissimilarity):
     def _compute(self, n, c_ff, c_ft, c_tf, c_tt):
         r = 2 * (c_tf + c_ft)
         return Expression('Divide', r, c_tt + c_ff + r)
+
+
+distribsAssumptions = {'BernoulliDistribution[p]':'0 <= p <= 1',
+                       'BetaBinomialDistribution[alpha, beta, n]':'n \[Element] Integers && n > 0 && alpha > 0 && beta > 0',
+                       'BetaDistribution[alpha, beta]':'alpha > 0 && beta > 0',
+                       'BetaNegativeBinomialDistribution[alpha, beta, n]':'alpha > 0 && beta > 0 && n > 0',
+                       'CauchyDistribution[a, b]':'a \[Element] Reals && b > 0',
+                       'ChiDistribution[nu]':'nu > 0',
+                       'ChiSquareDistribution[nu]':'nu > 0',
+                       'DiscreteUniformDistribution[{imin, imax}]':'imin \[Element] Integers && imax \[Element] Integers && imax >= imin',
+                       'ExponentialDistribution[lambda]':'lambda > 0',
+                       'ExtremeValueDistribution[alpha, beta]':'alpha \[Element] Reals && beta > 0',
+                       'FRatioDistribution[n, m]':'n > 0 && m > 0',
+                       'GammaDistribution[alpha, beta]':'alpha > 0 && beta > 0',
+                       'GammaDistribution[alpha, beta, gamma, nu]':'alpha > 0 && beta > 0 && gamma > 0 && nu \[Element] Reals}',
+                       'GumbelDistribution[alpha, beta]':'alpha \[Element] Reals && beta > 0',
+                       'HalfNormalDistribution[theta]':'theta > 0',
+                       'InverseGaussianDistribution[mu, lambda]':'mu > 0 && lambda > 0',
+                       'InverseGaussianDistribution[mu, lambda, theta]':'mu > 0 && lambda > 0 && theta \[Element] Reals',
+                       'LaplaceDistribution[mu, beta]':'mu \[Element] Reals && beta > 0',
+                       'LogisticDistribution[mu, beta]':'mu \[Element] Reals && beta > 0',
+                       'LogNormalDistribution[mu, sigma]':'mu \[Element] Reals && sigma > 0',
+                       'LogSeriesDistribution[theta]':'theta > 0 && theta < 1',
+                       'MaxwellDistribution[sigma]':'sigma > 0',
+                       'NegativeBinomialDistribution[n, p]':'n >= 0 && 0 < p <= 1',
+                       'NoncentralChiSquareDistribution[nu, lambda]':'nu > 0 && lambda >= 0',
+                       'NoncentralFRatioDistribution[n, m, lambda]':'n > 0 && m > 0 && lambda >= 0',
+                       'NoncentralFRatioDistribution[n, m, lambda, eta]':'n > 0 && m > 0 && lambda >= 0 && eta >= 0',
+                       'NormalDistribution[mu, sigma]':'mu \[Element] Reals && sigma > 0',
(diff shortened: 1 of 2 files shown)


mathics/builtin/combinatorial.py
```diff
@@ -266,3 +266,923 @@ class RogersTanimotoDissimilarity(_BooleanDissimilarity):
...
+    def probability(self, conditions):
+
+
+
+class DistributionParameterQ(Builtin):
+
+    def apply(self, expr, evaluation):
+        'DistributionParameterQ[expr]'
+    ##todo
+
```
Piruzzolo: this is to check if the distribution is valid. First, it would check the distribution itself, eventually yielding argr message. Second, if the test passes, it will check numerical parameters for correctness. Symbolic params are assumed correct

mathics/builtin/combinatorial.py
```diff
@@ -266,3 +266,923 @@ class RogersTanimotoDissimilarity(_BooleanDissimilarity):
...
+
+class DistributionParameterAssumptions(Builtin):
+
+    def apply(self, distribution, evaluation):
+        'DistributionParameterAssumptions[distribution]'
+
+        if isinstance(distribution, Distribution):
+            return String(distribution.arg_assumptions)
+
```
Piruzzolo: this gives the arg_assumptions string with slot parameters replaced by actual ones

mathics/builtin/combinatorial.py
```diff
@@ -266,3 +266,923 @@ class RogersTanimotoDissimilarity(_BooleanDissimilarity):
...
+
+
+class SurvivalFunction(Builtin):
+
+    def apply(self, distribution, evaluation):
+        'SurvivalFunction[distribution]'
+        
+        return Expression(evaluate('1 - CDF[distribution, x]'))
+
```
Piruzzolo: is it correct?

mathics/builtin/combinatorial.py
```diff
@@ -266,3 +266,923 @@ class RogersTanimotoDissimilarity(_BooleanDissimilarity):
...
+class BetaDistribution(Distribution):
+
+    'BetaDistribution[alpha, beta]'
+    
+    parameters = [alpha, beta]
+    nargs = 2
+    sympy_name = 'Beta'
+    arg_assumptions = 'alpha > 0 && beta > 0'
+
```
Piruzzolo: I plan to call the sympy corresponding function (parameters are nearly the same)

sn6uv:
Take a look at how I've set things up in my [Stats](https://github.com/sn6uv/Mathics/tree/Stats) branch, that is, commit 311aa91849cc637f1b37a768db5ad6df600f45dc.

mathics/builtin/combinatorial.py
sn6uv: Yes, but it's better to use a rule for this, e.g.
```python
rules = {
    'SurvivalFunction[dist_]': '1 - CDF[dist]',
}
```

mathics/builtin/combinatorial.py
sn6uv: Something I came across playing with sympy, sympy requires that you qualify explicitly that the params are correct. E.g. 
```python
>>> mu = Symbol("mu")
>>> sigma = Symbol("sigma", positive=True)    # fails if this flag isn't set!
>>> z = Symbol("z")

>>> X = Normal("x", mu, sigma)
```

mathics/builtin/stats.py
```diff
@@ -0,0 +1,1743 @@
...
+    msg = ''
+
+    @staticmethod
+    def convert(value):
+        return
+
+    @staticmethod
+    def get(value): #gets value for check_too -- todo
+        return
+
```
Piruzzolo: this is to get the param value, to be used by check_too. Is it a simpler way to check parameters?

mathics/builtin/stats.py
```diff
@@ -0,0 +1,1743 @@
...
+
+class IntegerParam(DistributionParam):
+    msg = 'intprm'
+
+    @staticmethod
+    def convert(value):
+        if isinstance(value, Symbol):
+            # TODO
```
Piruzzolo: What to do in case it is a Symbol?

mathics/builtin/stats.py
```diff
@@ -0,0 +1,1743 @@
...
+     = Sqrt[2] InverseErfc[2 - 2 #1]&
+    '''
+
+    rules = {
+        'InverseCDF[dist_, x_]': 'InverseCDF[dist][x]',
+        'InverseCDF[dist_, xs_List]': 'InverseCDF[dist] /@ xs',
+    }
+
+    def apply(self, dist, evaluation):
+        'InverseCDF[dist_, x]' #does not support operator form!
```
Piruzzolo: does not support operator form

mathics/builtin/stats.py
```diff
@@ -0,0 +1,1743 @@
...
+        'TriangularDistribution[]': 'TriangularDistribution[{0, 1}, 1/2]',
+        'TriangularDistribution[{min, max}]': 'TriangularDistribution[{min, max}, (max+min)/2]'
+    }
+
+
+class UniformDistribution(_SympyDistribution):
+
+    params = (RealParam, RealParam) #-infinity<first, first<b<infinity
+    sympy_name = 'Uniform'
+
```
Piruzzolo: How do I get a list of parameters? Using something like 
`params = (ListParam(RealParam, RealParam), RealParam)`? How?

mathics/builtin/stats.py
sn6uv: I think it's okay to just pass it on (as a sympy symbol) but I didn't check that everything worked. That is, do we need `sympy.Symbol('foo', integer=True)`?

mathics/builtin/stats.py
```diff
@@ -0,0 +1,1743 @@
...
+    sympy_name = 'Geometric'
+
+    def check_too(self):
+        par = self.params[0].get()
+        if par > 0 and par < 1:
+            return True
+        else:
+            return False
+        
+'''
```
Piruzzolo: not yet implemented distributions are commented out

sn6uv:
Are all the lists of parameters just intervals? Maybe `IntervalParam(lower=None, upper=None)`.

sn6uv:
Yes, using a big dictionary would be simpler but harder to write tests and documentation for.

Piruzzolo:
the big problem is that there are too much differences with sympy output, even for common distributions. Now, I think the priority is to get main methods work (expectation, probability, etc). I can begin tracking the differences between sympy and MMA output, so to make the fixings easier

PR #435, opened 2016-07-03
implement ExpIntegralEi and ExpIntegralE.

sn6uv:


mathics/builtin/specialfunctions.py
```diff
@@ -858,3 +858,34 @@ def prepare_sympy(self, leaves):
...
+    <dl>
+    <dt>'ExpIntegralEi[$z$]'
+      <dd>returns the exponential integral function $Ei(z)$.
+    </dl>
+
+    >> ExpIntegralEi[2.0]
+     = 4.95423435600189016
+    '''
+
+    sympy_name = ''
```
bnjones: Can we use sympy's `Ei` function here?

sn6uv: Ah, yes. I was looking for something like that.

bnjones: http://reference.wolfram.com/language/ref/ExpIntegralEi.html has some special values for `ExpIntegralEi[0]` and `ExpIntegralEi[Infinity]`, but it looks like SymPy doesn't automatically return them:
```
>>> Ei(0)
Ei(0)
>>> Ei(oo)
Ei(oo)
```
Maybe we could add a couple of rules for these?

sn6uv: I've opened https://github.com/sympy/sympy/pull/11341 to add this functionality to sympy.

PR #436, opened 2016-07-03
Fuzz fixes

sn6uv:
I wrote a small fuzzing script and found a bunch of minor bugs. Mostly caused by not checking for `None` when calling `to_sympy()`.

### DIFF
--- a/mathics/builtin/algebra.py
+++ b/mathics/builtin/algebra.py
@@ -480,29 +480,33 @@ class Variables(Builtin):
 
         variables = set()
 
-        def find_vars(e):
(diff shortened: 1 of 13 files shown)


bnjones:
Nice! The fuzzing script sounds like it'd be very useful - is this something we could add?

sn6uv:
The fuzzing script takes quite a while to run (and still causes some crashes). It's in my `function_fuzzer` branch if you want to take a look. I've also got some parser fuzzing scripts I could add so maybe it's worth making a separate repository.

PR #445, opened 2016-07-08
implement Position

sn6uv:


mathics/builtin/lists.py
```diff
@@ -1389,6 +1389,47 @@ class DeleteCases(Builtin):
...
+    >> Position[{1, 2, 2, 1, 2, 3, 2}, 2]
+     = {{2}, {3}, {5}, {7}}
+    '''
+
+    options = {
+        'Heads': 'True'
+    }
+
+    def apply_invalidlevel(self, patt, expr, ls, evaluation, options={}):
+        'Position[expr_, patt_, ls_, OptionsPattern[Map]]'
```
bnjones: Should this (and the one below) be `OptionsPattern[Position]`?

bnjones:
Looks great! The only thing I can think of that might be missing is a test (plus a bit in the `<dl>` block) for `Pattern[expr, patt, levelspec]`.

mathics/builtin/lists.py
sn6uv: Yes, thanks for catching that.

PR #446, opened 2016-07-09
Clustering for Mathics

poke1024:
`FindClusters[]` and `ClusteringComponents[]`.

The automatic cluster count detection tries to resemble what MMA does, which sometimes works and sometimes doesn't.

Haven't tested this yet with large point samples and need to do some more stress tests.

### DIFF
--- a/mathics/algorithm/clusters.py
+++ b/mathics/algorithm/clusters.py
@@ -119,6 +119,9 @@ def _ratio_bigger_than(a, b):
     return a1 * b2 > b1 * a2
 
 
+def _pairwise_sum(a, b):
+    return [x + y for x, y in zip(a, b)]
+
 class InfiniteSilhouette(Exception):
     # thrown when two clusters have distance 0
     pass
@@ -1008,7 +1011,7 @@ class _KMeans:
         for i, xi in enumerate(x):
             ai, u[i], _, l[i] = _smallest2(d(xi, cj) for cj in c)
             q[ai] += 1
-            cc[ai] = [a + b for a, b in zip(cc[ai], xi)]
+            cc[ai] = _pairwise_sum(cc[ai], xi)
             a[i] = ai
 
         s = [0] * len(c)
--- a/mathics/algorithm/clusters.py
+++ b/mathics/algorithm/clusters.py
@@ -36,11 +36,6 @@ import math
 # improvement was made in the meantime.
 
 
-# the following value defines by how much the McClain-Rao limit (for considering a clustering finished) changes
-# with each binary split of clustering. this is an experimental value.
-_mcclain_rao_limit_factor = 10.
-
-
 def _index(i, j):  # i > j, returns j + sum(1, 2, ..., i - 1)
     # x x x
     # a x x
@@ -193,13 +188,12 @@ def _agoras(n, k, distance):  # see [Rangel2016]
 
 
 class _Cluster:
-    def __init__(self, medoid, members, within_cluster_cost):
+    def __init__(self, medoid, members):
         self.medoid = medoid
         self.members = members
-        self.within_cluster_cost = within_cluster_cost
 
     def translate(self, i_to_i0):
-        return _Cluster(i_to_i0[self.medoid], [i_to_i0[i] for i in self.members], self. within_cluster_cost)
+        return _Cluster(i_to_i0[self.medoid], [i_to_i0[i] for i in self.members])
 
     @staticmethod
     def criterion(clusters, distance):
@@ -220,12 +214,12 @@ class _Cluster:
                     s_w += distance(i, m)
                 n_w += 1
 
-        if n_w == 0:
+        if n_w <= len(clusters):  # only single-medoid clusters remain.
             return None
 
         # s_b is the sum of between-cluster (medoid to medoid) distances.
         s_b = 0
-        n_b = 0
+        n_b = len(clusters)
         for z, c in enumerate(clusters):
             for d in clusters[:z]:
                 s_b += distance(c.medoid, d.medoid)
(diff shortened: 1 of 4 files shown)


mathics/algorithm/clusters.py
```diff
@@ -0,0 +1,714 @@
...
+# we use the approach described in [Hamerly2003]. clustering quality is measures using McClain-Rao's measure (see
+# [Desgraupes2013]).
+
+# we modify ClaraNS by keeping track of the randomly picked swaps and not trying the same swap twice if no
+# improvement was made in the meantime.
+
+
+# the following value defines by how much the McClain-Rao limit (for considering a clustering finished) changes
+# with each binary split of clustering. this is an experimental value.
+_mcclain_rao_limit_factor = 10.
```
bnjones: Does the best value of this limit depend on the input in some way? If so, should it be user-tunable?

bnjones:
Looks good, but how about a graphical example? Something like:

```
a = Partition[RandomReal[{1, 3}, 40], 2];
b = Partition[RandomReal[{2, 5}, 40], 2];
Graphics[{Blue, Point[a], Red, Point[b]}]
Graphics[{Red} ~Join~
       Riffle[Point /@ FindClusters[Join[a, b], Method->"Optimize"],
       {Blue, Green, Yellow}]]
Graphics[{Red} ~Join~
       Riffle[Point /@ FindClusters[Join[a, b], Method->"Agglomerate"],
       {Blue, Green, Yellow}]]
```

sn6uv:
Did some testing on the new `_shuffled_range` function (since I didn't believe the code) and it appears to be solid.

The only other thing I could thing of (sacrificing randomness for performance) would be to take a number `m` coprime to `n` and yield `k n` for `1 <= k <= n`. That is:

```
import random
from fractions import gcd


def shuffle(n):
    # find m coprime to n (exclude 1 and -1)
    m = n
    while gcd(n, m) != 1:
        m = random.randint(3, n-2)

    # start somewhere random
    p = random.randint(0, n-1)

    # generate 0, 1, 2, ..., n - 1 shuffled.
    for k in range(n):
        yield (p + k * m) % n
```

Of course it's not random (2 random numbers determine the whole sequence) but it's fast.

mathics/algorithm/clusters.py
poke1024: The assumption of the current algorithm on cluster count detection is this: for each possible clustering, we compute the ratio of the average in-cluster distance (i.e. the distances from all points of a cluster to the cluster's center) to the average between-cluster distance (the distance between cluster centers). With each split, that doubles the count of clusters, this ratio has to lower by a constant factor, or we won't do the split. For example, if we go from 2 to 4 clusters, then the ratio of each of the 4 clusters has to be smaller than the 2-cluster-ratio divided by some `x`. This `x` is `_mcclain_rao_limit_factor`. It basically says: if we split into more clusters, and the configuration isn't more compact (i.e. centered around cluster points) by this and that degree, we don't split, since then these new clusters are not delineated properly. So, to summarize, a `_mcclain_rao_limit_factor` of 10 states that the distance ratio needs to diminish by factor of 10 with each split, but we do not care what the absolute value of the distance ratio is.

Even though `_mcclain_rao_limit_factor` does not depend on the input data, it does obviously influence the granularity of the result. Maybe an option `Granularity` would be nice?

poke1024:
@sn6uv: that `shuffle` is a neat idea indeed. I also thought along the lines having a generator that does not repeat, but then I had qualms about not being random enough.

These are two other approaches for non-repeating generators:

https://en.wikipedia.org/wiki/Linear_feedback_shift_register
https://en.wikipedia.org/wiki/Format-preserving_encryption

This guy has an algorithm using quadratic residues:

http://preshing.com/20121224/how-to-generate-a-sequence-of-unique-random-integers/

I'm quite happy though with the Fisher-Yates shuffle currently implemented as it's as random as the Mersenne Twister in Python will get. I haven't seen it anywhere before in this form, but it's really just the plain thing (https://en.wikipedia.org/wiki/Fisher–Yates_shuffle) with materializing only the items that get swapped in the "upper" (> i) range of the array.

sn6uv:
I like your implementation of Fisher-Yates; I've never seen it implemented like that before. Thanks for the links!

mathics/builtin/lists.py
```diff
@@ -3332,3 +3336,309 @@ class TakeSmallestBy(_RankedTakeSmallest):
...
+
+
+# to_precision_float should be somewhere more central than this.
+def _to_precision_float(x):
+    if isinstance(x, Real):
+        return x.to_sympy().num
+    elif isinstance(x, Integer):
+        return mpf(x.get_int_value())
+    else:
+        raise _IllegalDataPoint
```
sn6uv: Should we also handle `Rational` here?

mathics/builtin/lists.py
```diff
@@ -3332,3 +3336,309 @@ class TakeSmallestBy(_RankedTakeSmallest):
...
+        d = Expression('N', self._df(p[i], p[j])).evaluate(self._evaluation)
+        if not d.is_numeric():
+            raise _IllegalDistance(d)
+        sympy_d = d.to_sympy()
+        if not sympy_d.is_real or sympy_d < 0:
+            raise _IllegalDistance(d)
+        return mpf(sympy_d)
+
+
+# to_precision_float should be somewhere more central than this.
```
sn6uv: How about a `to_mpmath` method on `BaseExpression`?

poke1024: Good idea. Should this be part of #463 or shall I change it here?

sn6uv:
Is this ready to be merged? I've rebased this on the current `master` and pushed to #491 (minor conflict with the #484).

sn6uv:
Great stuff!

PR #450, opened 2016-07-12
implement On and Off

sn6uv:
Had to be a bit careful to preserve the set of quiet messages between evaluations.

### DIFF
--- a/mathics/core/evaluation.py
+++ b/mathics/core/evaluation.py
@@ -20,6 +20,8 @@ FORMATS = ['StandardForm', 'FullForm', 'TraditionalForm',
            'TeXForm', 'MathMLForm',
            'MatrixForm', 'TableForm']
 
+QUIET_MESSAGES = set()
+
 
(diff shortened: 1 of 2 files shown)


mathics/core/evaluation.py
```diff
@@ -161,7 +163,7 @@ def __init__(self, definitions=None,
         self.predetermined_out = None
 
         self.quiet_all = False
-        self.quiet_messages = set()
+        self.quiet_messages = QUIET_MESSAGES
```
bnjones: What happens with `QUIET_MESSAGES` in a multi-user session? Does one user's `Off` affect everyone?

sn6uv:
Yes, it does. I think a similar thing might be happening for RecursionLimit. I thought it's not an issue because we're not doing multithreaded evaluation but I forgot about multiple users. Maybe it's worth bundling this stuff into a e.g. KernelConfig class.

PR #451, opened 2016-07-13
implement Scan and Return

sn6uv:
essentially a copy of `Map` but returns `Null`.

### DIFF
--- a/mathics/builtin/control.py
+++ b/mathics/builtin/control.py
@@ -308,6 +308,12 @@ class For(Builtin):
      = 3628800
     >> n == 10!
      = True
+
+    #> n := 1
+    #> For[i=1, i<=10, i=i+1, If[i > 5, Return[i]]; n = n * i]
+     = 6
+    #> n
+     = 120
     """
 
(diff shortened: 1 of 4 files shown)


mathics/builtin/control.py
```diff
@@ -599,8 +599,36 @@ def apply(self, evaluation):
...
+    <dl>
+    <dt>'Return[$expr$]'
+      <dd>aborts a function call and returns $expr$.
+    </dl>
+
+    >> f[x_] := (If[x < 0, Return[0]]; x)
+    >> f[-1]
+     = 0
+
+    >> g[x_] := (Do[If[x < 0, Return[0]], {2, 1, 0, -1}]; x)
```
bnjones: Not sure if I get what's happening in the `Do` here - is this testing what happens with `Return` and scoping constructs that don't evaluate?

sn6uv: It's an example I copied off the MMA `Return` docs. I don't think it's a good example. I'll fix it :)

PR #453, opened 2016-07-13
Disable spellchecking on query fields

bnjones:
This PR just disables spellchecking by setting the [HTML5 spellcheck attribute](https://www.w3.org/TR/html5/editing.html#attr-spellcheck) - doesn't seem like this buys us much, it doesn't help me spell `LaguerreL`


sn6uv:
Nice! This had been bothering me for a while but I never got round to looking into it.

PR #461, opened 2016-07-20
AngleVector[] and AnglePath[]

poke1024:
Seemed so simple, but is now quite elaborate since a special case for numerical evaluation seems important to me, as speed is an issue here.

One thing that vexes me is that numerical evaluation does not automatically get triggered when using this in `Graphics`, which means that a very slow symbolic evaluation is triggered inside `Graphics` if one does not specify numerical quantities. More specifically:

```
Timing[Graphics[Line[AnglePath[Table[90 Degree, {20}]]]]]
{3.6392160000000002, -Graphics-}

Timing[Graphics[Line[AnglePath[Table[1.57, {20}]]]]]
{0.013397999999999577, -Graphics-}
```

As you can see, the numerical evaluation (due to the special case in `AnglePath`) is about two magnitudes faster. I really wonder how we could get `Graphics` to do a recursive `N` on all its sub expressions before these get evaluated...

### DIFF
--- a/mathics/builtin/numeric.py
+++ b/mathics/builtin/numeric.py
@@ -21,7 +21,6 @@ import zlib
 import math
 from six.moves import range
 from collections import namedtuple
-import sys
 
 
 from mathics.builtin.base import Builtin, Predefined
@@ -903,36 +902,6 @@ class Fold(object):
     # allows inherited classes to specify a single algorithm implementation that
     # can be called with machine precision, arbitrary precision or symbolically.
 
-    class PrecisionExhausted(Exception):
-        pass
-
-    class SymbolicEvaluation(Exception):
-        pass
-
-    def to_machine_float(self, expr):
-        if isinstance(expr, Real):
-            precision = expr.get_precision()
-            if precision is not None and precision > _float_mantissa_digits:
-                raise Fold.PrecisionExhausted
-            return expr.to_python()
-        elif isinstance(expr, Integer):
-            x = expr.get_int_value()
-            if x.bit_length() > _float_mantissa_digits:
-                raise Fold.PrecisionExhausted
-            return x
-        else:
-            raise Fold.SymbolicEvaluation
-
-    def to_precision_float(self, expr):
-        if isinstance(expr, Real):
-            y = expr.to_sympy()
-            assert y.is_real
-            return y.num
-        elif isinstance(expr, Integer):
-            return mpf(expr.get_int_value())
-        else:
-            raise Fold.SymbolicEvaluation
-
     ComputationFunctions = namedtuple(
         'ComputationFunctions', ('sin', 'cos'))
 
@@ -955,14 +924,15 @@ class Fold(object):
         if mode == 'symbolic':
             return lambda *args: args
 
-        f = getattr(self, 'to_%s_float' % mode)
+        func_name = 'to_%s_float' % mode
 
         def convert(*args):
             for arg in args:
                 if arg is None:
                     yield None
                 else:
-                    yield f(arg)
+                    f = getattr(arg, func_name)
+                    yield f()
 
         return convert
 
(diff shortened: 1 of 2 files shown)


sn6uv:
@poke1024 I don't think it should change the numerical evaluation issues here but I'm planning an overhaul of the numerics right now, see #463.

poke1024:
@sn6uv so here's what I imagine might be useful: add `to_machine_float` and `to_precision_float` to all `Expression`s so code that might want to deal with these special cases can easily handle them and fallback; see `Fold` in `numeric.py` in this PR for how I currently use this to implement multiple precision/symbolic evaluation.

This should also fit quite nicely with #463, as `to_machine_float` and `to_precision_float` are then implemented in `MachineReal` and `PrecisionReal`.

bnjones:
Definitely happy with `AngleVector`, but I'm not sure that `AnglePath` is the right approach. As you've said in the PR, mpmath isn't significantly slower than machine-precision floats. If you call this function on a large input list, where do we spend most of the time?

poke1024:
I have absolutely no qualms kicking out the machine-precision floats, mpmath makes sense though. We should stay away from `sympy.Float` for calculations like that.

Here are some numbers:

test case: generating 10000 elements, i.e. `Length[AnglePath[Table[0.1, {10000}]]]`

CPython 3.5.1/Python floats took 5.985 s; `Fold._fold` not in profile
CPython 3.5.1/mpmath took 7.948 s; 0.971 s spent in `Fold._fold`
CPython 3.5.1/sympy.Float took 20.503 s; 15.103 s spent in `Fold._fold`; 14.429 s spent in `sympy/core`

PyPy 5.3.1/Python floats took 2.373 s; < 0.1s spent in `Fold._fold`
PyPy 5.3.1/mpmath took 3.793 s; 1.377 s spent in `Fold._fold`
PyPy 5.3.1/sympy.Float took 11.366 s; 8.724 s spent in `sympy/core`

A purely symbolic evaluation with 50 elements took 40.120 seconds, all time was spent in `core/expression.py` and `core/pattern.py` (CPython3).

For 50000 elements, the mpmath-float difference shrinks to 22% (29.804 (float) vs. 36.557 (mpmath) in CPython3).

poke1024:
@sn6uv @bnjones 

Found one major factor: changing `Real.__hash__` from the current implementation (`hash(("Real", self.to_sympy().n(dps(_prec))))`) to `hash(("Real", self.to_python()))` chops off 20% or more of the performance. Obviously this still needs to be rounded to 7 digits, as the comment there says.

Before:
`Timing[Length[AnglePath[Table[0.1, {50000}]]]]
Out[3]= {2.043277, 1}`

After:
`Timing[Length[AnglePath[Table[0.1, {50000}]]]]
Out[5]= {1.4336660000000006, 1}`

Maybe this can be addressed in #463?

mathics/builtin/numeric.py
```diff
@@ -890,3 +894,87 @@ def compute(user_hash, py_hashtype):
...
+            sin=mpmath.sin,
+        ),
+        SYMBOLIC: ComputationFunctions(
+            cos=lambda x: Expression('Cos', x),
+            sin=lambda x: Expression('Sin', x),
+        )
+    }
+
+    operands = {
+        FLOAT: lambda x: x.to_python(),
```
sn6uv: In light of #463 this could be `x.round_to_float()`.

mathics/builtin/numeric.py
```diff
@@ -890,3 +894,87 @@ def compute(user_hash, py_hashtype):
...
+        MPMATH: lambda x: x.to_mpmath(),
```
sn6uv: mpmath is a bit finicky; doing this conversion outside a `with mpmath.workprec` block might cause a loss of precision. We should at least add a test against this in `AnglePath`.

PR #462, opened 2016-07-20
optimise IntegerLength

sn6uv:
I noticed some strange performance regressions in `IntegerLength`.

Traced this down to an intermittent bug in sympy: https://github.com/sympy/sympy/issues/11414

Here's a workaround.


sn6uv:
No reply to the `sympy` issue in a week and I keep hitting this while testing `Times` so I'm going to merge.

PR #458, opened 2016-07-17
Optimise BaseExpression creation

sn6uv:
Some small changes to the core to speed up `BaseExpression` creation.

Also includes some enhancements to the benchmarking script.

### Before:
```
'Expand[(a1+a2+a3+a4+a5+a6+a7)^3]'
    10 loops, avg:  147 ms, best:  136 ms, median:  145 ms per loop
%timeit Integer(5)
    1000000 loops, best of 3: 1.52 µs per loop
%timeit Expression('Sin')
     100000 loops, best of 3: 4.03 µs per loop
```

### After:
```
'Expand[(a1+a2+a3+a4+a5+a6+a7)^3]'
      100 loops, avg: 98.3 ms, best: 93.1 ms, median: 95.5 ms per loop
%timeit Integer(5)
    1000000 loops, best of 3: 1.15 µs per loop
%timeit Expression('Sin')
     100000 loops, best of 3: 3.2  µs per loop
```

### After `lru_cache`:  **(deferred)**
```
'Expand[(a1+a2+a3+a4+a5+a6+a7)^3]'
    100 loops, avg: 87.8 ms, best: 82.4 ms, median: 86.2 ms per loop
%timeit Integer(5)
    10000000 loops, best of 3: 158 ns per loop
%timeit Expression('Sin')
    1000000 loops, best of 3: 1.99 µs per loop
```

The introduction of LRU cache caused some issues when copying and modifying expressions so I've removed them from this PR.

PR #463, opened 2016-07-20
Machine precision

### LINKED ISSUE #448: Mathics uses 64bit of machine precision
The Mathics precision system needs some reworking. We're currently using 64 bits of precision (plus some bits for the exponent etc.) but we should be using 53 (on 64-bit systems).

We could even consider using python floats for machine precision numbers (should give some performance gains for numerical stuff e.g. plotting).

sn6uv:
This is built on some parts of #458 (with `lru_cache` and forced `from_python` experiments removed).

The main change is that `Real` now has two subclasses, `MachineReal` and `PrecisionReal` to distinguish between machine-precision and arbitrary-precision reals. `PrecisionReal` uses `sympy.Float` (as in the current `Real` implementation), while `MachineReal` uses python `float` objects.

The goal is to fix #448 and allow #449 and #460 to be merged.


sn6uv:
@poke1024 . In response to your comments on #461.

The API I have so far is:

- `get_float_value : Number -> float` (could be extended to `Expression -> float or None` by giving an optional `n_evaluation` argument). I'm wondering whether `Complex` should return `None` or a python `complex` result.
- `round :  Expression -> Real or None`. Takes an optional argument `digitis` to determine whether `PrecisionReal` or `MachineReal` is returned. Also takes optional `n_evaluation` argument.

We need a way to handle `Complex` and maybe there's a better way to name things. This also replicates some of `to_python`. I think there's room for significant cleanups here.

Any thoughts? @bnjones

poke1024:
I'm happy with `round`, but I think `get_float_value` should have a name that indicates that it might chop off precision to reach machine precision; I also think that we need functions that guarantee that no loss of precision occurs. Two ideas:

(A)

`get_float_value`: returns `float`, if the number is within machine float precision, `sympy.Float` if it's a number with larger precision, and raises e.g. `NotAFloatNumber` if it's not finite or irrational or some symbolic expression.
`round_to_machine_float_value`: `get_float_value`'s current behaviour

(B)

`get_machine_float_value`: returns `float` or raises an exception if not representable by machine precision number
`get_precision_float_value`: returns `sympy.Float` or raises an exception if not representable by  a `sympy.Real`
`round_to_machine_float_value`: `get_float_value`'s current behaviour

For #461, (B) is more efficient.

bnjones:
I think we're usually given some target precision when we use these functions, so maybe it'd help to have a convenience method that handles an integer precision or `MachinePrecision` and returns a `PrecisionReal`/`MachineReal`, and if it's not a valid precision it raises some exception that we can catch and issue a message for.

If we need a Python float we can then call `MachineReal.get_float_value()`. I think I'd agree that `.get_float_value()` shouldn't round.

IMO this would make it easier to write NValues for builtins, or to implement `Plot[..., WorkingPrecision -> something]`.

sn6uv:
> I think we're usually given some target precision when we use these functions, so maybe it'd help to have a convenience method that handles an integer precision or MachinePrecision and returns a PrecisionReal/MachineReal, and if it's not a valid precision it raises some exception that we can catch and issue a message for.

`round` is currently only implemented for `Number` but we could extend it to `BaseExpression` (with an optional `n_evaluation` argument).


@poke1024 @bnjones, I agree. Calling `round` explicitly when needed is better than implicit rounding. I will change `get_float_value`. Is it worth renaming `get_float_value(permit_complex=False)` to `get_machine_value` or similar?

@poke1024 under this approach (I also prefer **B**) `round_machine_value` would be equivalent to`round().get_float_value()`.
`get_precision_float_value` is not quite `round(d).to_sympy()` because it might not return a `sympy.Float`. We could add a helper method to handle this.

sn6uv:
@poke1024 @bnjones I'm done tweaking this. Any final inputs?

mathics/builtin/arithmetic.py
```diff
@@ -785,23 +821,19 @@ def apply_check(self, x, y, evaluation):
...
         # Power uses _MPMathFunction but does some error checking first
-        if isinstance(x, Number) and x.to_sympy() == 0:
+        if isinstance(x, Number) and x.is_zero:
             if isinstance(y, Number):
                 y_err = y
             else:
                 y_err = Expression('N', y).evaluate(evaluation)
             if isinstance(y_err, Number):
-                py_y = complex(y_err.to_sympy()).real   # take real part
+                py_y = y_err.round().get_float_value(permit_complex=True).real
```
sn6uv: get_float_value(permit_complex=True)

mathics/builtin/exptrig.py
```diff
@@ -43,9 +44,13 @@ class Pi(SympyConstant):
 
     def apply_N(self, precision, evaluation):
         'N[Pi, precision_]'
-        precision = get_precision(precision, evaluation)
-        if precision is not None:
-            return Real(sympy.pi.n(dps(precision)), p=precision)
+        if precision.get_name() == 'System`MachinePrecision':
+            return MachineReal(math.pi)
+        else:
+            d = precision.round_to_float(evaluation)
```
sn6uv: what if `d` is negative.

mathics/builtin/exptrig.py
```diff
@@ -56,24 +61,28 @@ class E(SympyConstant):
...
 
     def apply_N(self, precision, evaluation):
         'N[E, precision_]'
-        precision = get_precision(precision, evaluation)
-        if precision is not None:
-            return Real(sympy.E.n(dps(precision)), p=precision)
+        if precision.get_name() == 'System`MachinePrecision':
+            return MachineReal(math.e)
+        else:
+            d = precision.round_to_float(evaluation)
```
sn6uv: `d < $MinPrecision`?

mathics/builtin/lists.py
```diff
@@ -1531,7 +1531,10 @@ def apply_max(self, expr, imax, evaluation):
         '%(name)s[expr_, {imax_}]'
 
         index = 0
-        imax = imax.evaluate(evaluation).get_real_value()
+        imax = imax.evaluate(evaluation).numerify(evaluation)
+        if isinstance(imax, Number):
+            imax = imax.round()
+        imax = imax.get_float_value()
```
sn6uv: round_to_float

mathics/builtin/numeric.py
```diff
@@ -124,48 +160,52 @@ class N(Builtin):
...
-                    eval_range = list(range(1, len(expr.leaves)))
-                elif 'System`NHoldRest' in attributes:
-                    if len(expr.leaves) > 0:
-                        eval_range = (0,)
-                    else:
-                        eval_range = ()
+        if prec.get_name() == 'System`MachinePrecision':
+            d = None
+        else:
+            d = prec.round_to_float(evaluation)
```
sn6uv: `d < $MinPrecision`?

mathics/builtin/numeric.py
```diff
@@ -174,16 +214,22 @@ class MachinePrecision(Predefined):
...
-            return Real(dps(machine_precision), prec)
+        if prec == Symbol('MachinePrecision'):
+            # avoids recursion
+            return MachineReal(machine_precision * math.log10(2))
+        return Expression('N',
+            Expression('Times', Expression('Log', Integer(10), Integer(2)),
+                       Integer(machine_precision)), prec)
 
 
 class Precision(Builtin):
```
sn6uv: add note about only integer precision in Mathics

mathics/builtin/specialfunctions.py
```diff
@@ -702,12 +711,20 @@ class AiryAiZero(Builtin):
     def apply_N(self, k, precision, evaluation):
         'N[AiryAiZero[k_Integer], precision_]'
 
-        prec = get_precision(precision, evaluation)
+        if precision.get_name() == 'System`MachinePrecision':
+            prec = machine_precision
+            d = None
+        else:
+            d = precision.round_to_float(evaluation)
```
sn6uv: `d < $MinPrecision`?

mathics/builtin/specialfunctions.py
```diff
@@ -742,12 +762,20 @@ class AiryBiZero(Builtin):
     def apply_N(self, k, precision, evaluation):
         'N[AiryBiZero[k_Integer], precision_]'
 
-        prec = get_precision(precision, evaluation)
+        if precision.get_name() == 'System`MachinePrecision':
+            prec = machine_precision
+            d = None
+        else:
+            d = precision.round_to_float(evaluation)
```
sn6uv: `d < $MinPrecision`?

mathics/core/expression.py
```diff
@@ -709,14 +711,17 @@ def get_sort_key(self, pattern_sort=False):
                     name = leaf.get_name()
                     if leaf.has_form('Power', 2):
                         var = leaf.leaves[0].get_name()
-                        exp = leaf.leaves[1].get_real_value()
+                        if isinstance(leaf.leaves[1], Number):
+                            exp = leaf.leaves[1].round_to_float()
+                        else:
+                            exp = None
```
sn6uv: extend `round_to_float` to `BaseExpression`.

mathics/core/expression.py
```diff
@@ -1738,70 +1704,186 @@ def user_hash(self, update):
...
+    def is_zero(self):
+        return self.value == 0.0
+
+
+class PrecisionReal(Real):
+    '''
+    Arbitrary precision real number.
+
+    value is stored internally as a sympy.Float.
+    prec is the number of bits of precision (differs from Mathematica).
```
sn6uv: fix

bnjones:
I haven't had a chance to read through the diff yet, but I gave this a try and it looks great. Nice work! I'd suggest adding an example to the `N` docs to illustrate how the input precision affects the result, e.g. `N[Exp[0.1], 100]` vs `N[Exp[1/10], 100]`

mathics/builtin/arithmetic.py
```diff
@@ -571,21 +599,35 @@ def apply(self, items, evaluation):
...
+                else:
+                    with mpmath.workprec(prec):
+                        numbers = [item.to_mpmath() for item in numbers]
+                        number = mpmath.fprod(numbers)
+                        number = Number.from_mpmath(number, dps(prec))
+            else:
+                number = sympy.Mul(*[item.to_sympy() for item in numbers])
+                number = from_sympy(number)
+        else:
+            number = Integer(1)
```
bnjones: Is it worth factoring this `if numbers:` block out into a function so it can be shared with Plus?

mathics/builtin/arithmetic.py
```diff
@@ -1477,10 +1536,14 @@ class Gamma(_MPMathMultiFunction):
...
      = 1.242169344504305404913070252268300492431517240992022966055507541481863694148882652446155342679460339
     #> N[N[Gamma[24/10],100]/N[Gamma[14/10],100],100]
-     = 1.4
+     = 1.400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
     #> % // Precision
      = 100.
 
+    #> Gamma[1.*^20]
+     : Overflow occurred in computation.
+     = Overflow[]
```
bnjones: It took me a few tries to figure out how to do this:
```
In[36]:= Gamma[1.0000000000000*^20]
 : Overflow occurred in computation.
Out[36]= Overflow[]

In[37]:= Precision[1.0000000000000*^20]
Out[37]= MachinePrecision

In[38]:= N[MachinePrecision]
Out[38]= 15.9546

In[39]:= Gamma[1.`16*^20]
Out[39]= 1.932849514310098*^1956570551809674817225

In[40]:= Precision[%]
Out[40]= 16.
```
Since this brings us much closer to the MMA behaviour, is it worth adding some documentation to explain how overflows work and how to input numbers with different precisions? I don't think it should hold this PR up, but it'd be nice to have.

poke1024:
Looks very nice to me, and I like the "get real number" function design.

The only thing I noticed: maybe some check at the end of `core/numbers.py/convert()` that we're really dealing with a non-float/integer?

mathics/builtin/arithmetic.py
sn6uv: Probably. I tried a couple of times to clean this up / subclass `_MPMathFunction` but it just got a bit messy.

mathics/builtin/arithmetic.py
sn6uv: In mathematica these should all overflow. The reason for the different behaviour is that mpmath has unbounded exponents.

mathics/builtin/numeric.py
```diff
@@ -199,46 +261,133 @@ class Precision(Builtin):
...
-    def apply_complex(self, x, evaluation):
-        'Precision[x_Complex]'
 
-        if x.is_inexact():
-            return Real(dps(x.get_precision()))
-        else:
-            return Symbol('Infinity')
+class MinPrecision(Builtin):
+    '''
+    >> $MinPrecision
```
bnjones: It's a really minor nitpick, but it'd be great to have `<dl>` blocks for these.

PR #464, opened 2016-07-29
Faster insert_rule

poke1024:
Makes Mathics startup faster, by making the time in `contribute` in the `Definitions` constructor go to roughly 50% (from 2.5 to 1.1 s).

### DIFF
--- a/mathics/core/definitions.py
+++ b/mathics/core/definitions.py
@@ -10,6 +10,7 @@ import six.moves.cPickle as pickle
 import os
 import base64
 import re
+import bisect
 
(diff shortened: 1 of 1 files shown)


sn6uv:
Looking at the changes it's not clear what's going on. Possibly we're relying python's sort being stable?

poke1024:
Interesting. Didn't see that coming. I'll look into it.

sn6uv:
We also have a system to cache the definitions (not sure why it's currently disabled). Try adding `builtin_filename='mathics_defn_cache.txt'` to the `Definitions` constructor in `main.py`.

PR #466, opened 2016-07-29
Cleanup power

sn6uv:
Some cleanups refactored out of #463.


sn6uv:
I've disabled the failing tests for now (complex formatting discrepancies between python2 and python3). Plan to fix these up in #463.

PR #468, opened 2016-08-02
Cleanup of Style definitions

poke1024:
This is in preparation of adding `Arrowheads` and `FontSize` soon.

The current code makes adding new `Style` directives cumbersome, as several places of definition and reference in the code need to be changed.

This PR reorganizes and unifies the places where style heads are declared into one central place, such that adding a new style means adding a new entry to `styles` at the end of the file and nowhere else.

I didn't reorganize `element_heads` (apart from making it a `set`) as some funky mixing of `Graphics` and `Graphics3D` objects is going on there, that I didn't understand.

I'm not sure what the `# noqa` comment in `_GraphicsElements.__init__` means, so I just left it. Is it 'no QA'?

### DIFF
--- a/mathics/builtin/graphics.py
+++ b/mathics/builtin/graphics.py
@@ -1144,8 +1144,7 @@ class _GraphicsElements(object):
                 if item.get_name() == 'System`Null':
                     continue
                 head = item.get_head_name()
-                if (head in style_heads or  # noqa
-                    head in ('System`EdgeForm', 'System`FaceForm')):
+                if head in style_heads or head in ('System`EdgeForm', 'System`FaceForm'):
                     style.append(item)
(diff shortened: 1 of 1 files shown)


sn6uv:
The noqa is to disable pep8 warnings. I run pep8/flake8 on the codebase occasionally and clean things up but it's fine to remove it.

mathics/builtin/graphics.py
```diff
@@ -241,6 +229,10 @@ def init(self, graphics, item=None, style=None):
         self.style = style
         self.is_completely_visible = False  # True for axis elements
 
+    @staticmethod
+    def create_as_style(klass, graphics, item):
+        return klass(graphics, item)
```
sn6uv: German language strikes again :+1:.

PR #469, opened 2016-08-02
AtomBuiltin and decentral Atom box handling

poke1024:
A cleaner version of the `Atom` implementation used in `Image[]`, that makes it easy to add new `Atom`s, such as `Image[]`, `Graph[]`,  etc. without having to change central code in `MakeBoxes`.

Two changes:

(1) new `AtomBuiltin` (from #378); needed as a mechanism to register a builtin-like symbol that is actually an `Atom` (declaring an `Atom` directly does not register a name that could be used for constructing it in an `Expression`, so there always has to be an accompanying `Builtin`).

(2) moved out logic from `class MakeBoxes(Builtin)` into a new `Atom.atom_to_boxes` method, so that new `Atom`s need not add their logic (and import their symbols) to `MakeBoxes` but must instead provide an implementation of `atom_to_boxes`.

In the current implementation, `MakeBoxes` returns `None` on unknown `Atom` types. This PR changes this to raising an `NotImplementedError` for `Atom`s not implementing `atom_to_boxes`. As implementations are provided for `Symbol`, `Integer`, `Rational`, `Real`, `Complex`, `String`, this might only happen if new `Atom`s are added that do not properly implement `atom_to_boxes`.

### DIFF
--- a/mathics/builtin/base.py
+++ b/mathics/builtin/base.py
@@ -236,6 +236,16 @@ class InstancableBuiltin(Builtin):
         pass
 
 
+class AtomBuiltin(Builtin):
+    # allows us to define apply functions, rules, messages, etc. for Atoms
+    # which are by default not in the definitions' contribution pipeline.
+    # see Image[] for an example of this.
+
+    def get_name(self):
+        name = super(AtomBuiltin, self).get_name()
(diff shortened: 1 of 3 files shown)


sn6uv:
Great!

PR #478, opened 2016-08-11
Replace[]

poke1024:
An implementation for `Replace[]`.

The changes in `apply_rules` were necessary since `Replace` replaces depth first, whereas `ReplaceAll` replaces breadth first.


poke1024:
Seems like another Travis problem. Any simple way to restart this?

sn6uv:
I've restarted the errored build. I think you need to be a maintainer to do it.

sn6uv:
Nice!

PR #480, opened 2016-08-12
Changes for imathics kernel

poke1024:
I strongly suggest we merge the changes necessary to run Mathics as an imathics kernel into `master`. The current imathics branch is outdated and it's a pain to keep it updated. This is a proposal to add all necessary changes to master.

I've tested this with the experimental version in https://github.com/mathics/IMathics/pull/11 under Jupyter 5.

### DIFF
--- a/mathics/builtin/files.py
+++ b/mathics/builtin/files.py
@@ -795,8 +795,7 @@ class Write(Builtin):
         expr = expr.get_sequence()
         expr = Expression('Row', Expression('List', *expr))
 
-        evaluation.format = 'text'
-        text = evaluation.format_output(from_python(expr))
+        text = evaluation.format_output(from_python(expr), 'text')
         stream.write(six.text_type(text) + '\n')
         return Symbol('Null')
 
--- a/mathics/builtin/files.py
+++ b/mathics/builtin/files.py
@@ -795,7 +795,8 @@ class Write(Builtin):
         expr = expr.get_sequence()
         expr = Expression('Row', Expression('List', *expr))
 
-        text = evaluation.format_output(from_python(expr), 'text')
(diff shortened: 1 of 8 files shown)


sn6uv:
Great idea

poke1024:
Done.

Basically two changes:

(1) the callbacks in `Evaluation` are now gathered in a single class `Output`
(2) all image outputs (i.e. svgs, imgs) now output `mglyphs`

(2) has the advantage of working with Jupyter without additional JavaScript transformations. On the other hand, in order to not break any behaviours in the classic Mathics web frontend, I added JS code to convert `mglyph` to `svg` or `img`, so that there's really no change for the frontend JS logic.

I'm not sure about the max_allowed_packet size, which was changed in the imathics branch. Is it something that should always be unlimited for Jupyter kernels?

sn6uv:
I was running into trouble with the `max_allowed_packet_size` when working with large images. I don't remember exactly what it was sorry.

sn6uv:
This is breaking plots on Firefox for me.

poke1024:
I cannot reproduce this using Firefox 48.0.1 under OS X; I added a `svg.setAttribute('data-info', 'generated through convertMathGlyphs()');` in `web/media/js/mathics.js:convertMathGlyphs()` to make sure I really tested the code, but it ends up in my Firefox (see attached image) and it works. Which OS/Firefox versions are you using? Any extensions? JavaScript is enabled?
firefox

sn6uv:
Cleared my cache and it seems to be fixed now. Not sure what the problem was, half of the SVG wasn't present in the HTML document.

poke1024:
Maybe the new JavaScript didn't get loaded properly.

PR #482, opened 2016-08-16
Image[] (again)

poke1024:
This is a rebase of https://github.com/mathics/Mathics/pull/378 to the Mathics branch.

It works perfectly in the classic Mathics web interface.

Combined with https://github.com/mathics/Mathics/pull/480 and a small change, this will also work in Jupyter kernels.


sn6uv:
I've closed #378 in favour of this one.

poke1024:
in its current form, this now depends on https://github.com/mathics/Mathics/pull/480

mathics/builtin/image.py
```diff
@@ -1134,6 +1134,16 @@ def _image_pixels(matrix):
...
 
 
+class ImageQ(Builtin):
+    def apply_image(self, image, evaluation):
+        'ImageQ[image_Image]'
+        return Symbol('True')
+
+    def apply_no_image(self, array, evaluation):
+        'ImageQ[Image[array_]]'
+        return Symbol('False')
```
sn6uv: What about other non-images, e.g. `ImageQ[x]`?

mathics/builtin/image.py
```diff
@@ -1238,15 +1239,19 @@ def make_boxes(self, form):
                 scaled_height = int(scale * height)
                 pixels = skimage.transform.resize(pixels, (scaled_height, scaled_width), order=0)
 
-            stream = BytesIO()
-            skimage.io.imsave(stream, pixels, 'pil', format_str='png')
-            stream.seek(0)
-            contents = stream.read()
-            stream.close()
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore")
```
sn6uv: What warnings are we ignoring here? Are they important?

poke1024: Not really. During the conversion of floating point to byte components (which are necessary for the png encoding), PIL outputs a warning that precision might be lost. Since this code only handles the display for the frontend, it doesn't matter.

PR #485, opened 2016-08-19
FilledCurve[], BezierCurve[], BezierFunction[]

poke1024:
These work with svg and asy (I've tested both with the accompanying test cases). One test case specifically demonstrates that the points generated by `BezierFunction` match the curve which defined through the svg/asy commands.

There are two changes to `Piecewise` in here needed for `BernsteinBasis` (and because I was too confused and committed this to this PR instead of a new one). `Piecewise` now holds all expressions, as some might not be evaluable, and only evaluates conditions.


sn6uv:
Nice, this looks really nice. Especially the example using #494 changes.

PR #489, opened 2016-08-21
Fix #488

### LINKED ISSUE #488: #473 breaks web interface
#473 creates the Django database with root privileges and so prevents write access by normal users, which prevents the server from returning responses to input.

I'll fix this shortly. I would self-assign but apparently I can't.

jaycech3n:
Turns out all that was needed to fix save/load issue was to get the correct version Django 1.8.*.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,6 @@ import sys
 import platform
 import os
 from setuptools import setup, Command, Extension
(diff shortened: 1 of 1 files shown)


sn6uv:
Thanks for following up on this!

PR #499, opened 2016-08-24
BarChart[]

poke1024:
chart-example

still needs some error checks for a basic version.

still lacks legends.

### DIFF
--- a/mathics/builtin/plot.py
+++ b/mathics/builtin/plot.py
@@ -23,6 +23,12 @@ from mathics.builtin.scoping import dynamic_scoping
 from mathics.builtin.options import options_to_rules
 from mathics.builtin.numeric import chop
 
+try:
+    import palettable
+    palettable_installed = True
+except ImportError:
+    palettable_installed = False
+
 
 try:
     from mathics.builtin.compile import _compile, CompileArg, CompileError, real_type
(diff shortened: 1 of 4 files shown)


mathics/builtin/plot.py
```diff
@@ -543,6 +634,390 @@ def find_excl(excl):
...
+            multiple_colors = False
+            groups = [points]
+
+        def to_number(x):
+            if isinstance(x, Integer):
+                return float(x.get_int_value())
+            y = x.get_float_value()
+            if y is None:
+                return 0.
+            return y
```
sn6uv: I think this would be better implemented with `round_to_float(evaluation=evaluation)`. This one doesn't handle non-machine reals and symbolic arguments.

sn6uv:
Also, is `palettable` strictly necessary? I think `BarChart` is 'standard enough' that we should try to make it work by default (without having to install additional libs). Another option would be just to install it by default but maybe we could replicate the functionality easily enough?

poke1024:
We could add our own color tables, but it's not that easy to get good palettes and copying these or the `palettable` palettes always poses license issues. `palettable` is quite small (the color palette values plus helper methods plus some docs and tests, see https://github.com/jiffyclub/palettable) and it doesn't bring in any additional dependencies (the only completely optional dependency is matplotlib, but it only interfaces if you install matplotlib manually). It is documented to work Python 2 and Python 3 and I used it under PyPy without any problems. It solves the license brouhaha we'd have to solve if we copied their (or other) palettes into our code. The other option would obviously be our own original palettes, but I'm not sure where got get those or how to produce them. So I'm quite in favour of making `palettable` a default install, as it gives us very nice palettes without any headaches.

sn6uv:
Okay. Good points. Let's add it as a default then.

GarkGarcia:
See https://github.com/mathics/Mathics/pull/855 for updates.

PR #472, opened 2016-08-04
Functions for natural language processing

poke1024:
Opening a PR for this since I believe this is pretty complete, apart from some final QA passes.

Depends heavily on third party libraries, but thanks to `require` will handle this gracefully if packages are not installed.

This whole thing is obviously somewhat experimental, but so is most of this functionality in MMA.

Everything tries to resemble MMA functions. Noteworthy functions:

`Experimental`WordSimilarity` does not exist in MMA. Yet, it's extremely powerful for semantic analysis so I couldn't help including it, though putting it in this `Experimental` context.

`TextStructure` gives slightly different results than MMA (it's not the same parser after all), and it currently only supports the `ConstituentString` mode, since everything else relies of `Assocation`s which we do not support (yet).

I'm not sure what a good place for the installation notes (regarding `pattern` and `omw`) in the top of the file would look like. Maybe somewhere in the docs?

PR #500, opened 2016-08-24
TextRecognize[]

poke1024:
These language codes are the same ones as in https://github.com/mathics/Mathics/pull/472. Some common place to put them would be good.

### DIFF
--- a/mathics/builtin/image.py
+++ b/mathics/builtin/image.py
@@ -1381,23 +1381,45 @@ class TextRecognize(Builtin):
         'pyocr',
     )
 
+    messages = {
+        'tool': 'No text recognition tools were found in the paths available to the Mathics kernel.',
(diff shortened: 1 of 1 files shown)


mathics/builtin/image.py
```diff
@@ -1347,6 +1347,9 @@ class TextRecognize(Builtin):
 
     languages = {
         'English': 'eng',
+        'Spanish': 'spa',
+        'French': 'fra',
+        'German': 'deu',
```
Piruzzolo: Could you add Italian language too?

poke1024: Definitely, we just need to find a better place for the ISO language codes

sn6uv: Merged #520 so we can use that now.

PR #502, opened 2016-08-25
Faster StringSplit[]

poke1024:
Avoids the use of `from_python` and thus gains 10% speed.


sn6uv:
There are actually quite a few places `from_python` can be replaced with the (faster) explicit conversion. I started looking into this a while back as part of #463 but forgot about it. In particular, the `from_python` call in `Expression.__new__`.

poke1024:
Yep, the one in `Expression.__new__` is quite a hit, I tried to get rid of it once, but then I gave up because the plot code relies on it all over the place.

PR #507, opened 2016-08-25
Cache for Definitions

poke1024:
This continues https://github.com/mathics/Mathics/pull/501 with a more general approach.

Depending on the situation, the performance gain with this seems to be between 5% and 15%:

[before.txt](https://github.com/mathics/Mathics/files/437771/before.txt)
[after.txt](https://github.com/mathics/Mathics/files/437772/after.txt)

### DIFF
--- a/mathics/core/definitions.py
+++ b/mathics/core/definitions.py
@@ -12,7 +12,9 @@ import base64
 import re
 import bisect
 
-from mathics.core.expression import Expression, Symbol, String, fully_qualified_symbol_name
+from collections import defaultdict
+
+from mathics.core.expression import Expression, Symbol, String, fully_qualified_symbol_name, strip_context
 from mathics.core.characters import letters, letterlikes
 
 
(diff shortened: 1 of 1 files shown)


sn6uv:
rather than throwing out the whole cache each time, could we pass a name argument to `clear_cache` and only clear the conflicting items?

poke1024:
So I tried to limit the cache clears. It's a bit tricky since clearing `A` also needs to clear ```X`A``` and clearing ```X`A``` also needs to clear `A`, and so on.

The new approach is still very on the conservative side and is clearing more than needed (namely a stripped symbol's name across all contexts) and it doesn't try to identify a minimal set. We could try to be still smarter about it, but it's tricky, and since the number of gets should be drastically higher than the number of sets, the occasional wrong clears are, I believe, rare, and I'm a bit too lazy to think through all the border cases right now.

This new implementation is never slower than the previous cache implementation, and it's another 10% faster in some cases:

[caching-new.txt](https://github.com/mathics/Mathics/files/446628/caching-new.txt)

sn6uv:
Did some testing and got quite a significant speedup: 70s -> 47s. I'm not completely happy with how complicated the `Definitions` object is becoming but in this case I think the trade-off is worthwhile.

Issue #68, opened 2012-11-28
Undo and Redo in the web interface


sn6uv:
Something notoriously bad in Mathematica is the limited undo/redo. I think it would be great if we could add this feature to the web interface somehow. 

P.S. how do I tag this as an enhancement?

GarkGarcia:
Fixed on https://github.com/mathics/Mathics/commit/543596d2e6ac3192bea1c61211d4fe89c6a3ed45 (apparently)

PR #511, opened 2016-08-27
Implement undo/redo functionality

jaycech3n:
Undo/redo functionality implementation for the web interface (refer #68).

The input cells are implemented using `textarea` elements which have native browser undo/redo; however the specific way this works differs quite dramatically among browsers (e.g. Chrome vs. Firefox), and switching between cells can cause the undo history to be forgotten.
A custom implementation is much better.

**Tested on Firefox and Chrome; help testing on Safari appreciated.**

Notes:
-  No ability to undo deletion of cells.
- Undo limit can be set in `mathics.js`.

### DIFF
--- a/mathics/web/media/js/mathics.js
+++ b/mathics/web/media/js/mathics.js
@@ -7,6 +7,13 @@ var clickedQuery;
 
 var lastFocus = null;
 
+// Undo-redo functionality
+const UNDO_LIMIT = 250;
+var previousState = {};
+var saveNextState = false;
+var undoStates = [];
+var redoStates = [];
+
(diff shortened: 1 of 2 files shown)


sn6uv:
Just tested this out and it seems to work well. It would be nice if the undo could remove created cells (i.e. results) but I don't think that should block this. Some undo/redo buttons might also be a nice addition.

jaycech3n:
I've begun work on bringing the notebook interface up to date and have undo/redo buttons in the pipeline.

Cell handling is a little harder but potentially doable; I'll see what can be done.

PR #523, opened 2016-09-01
Quit and Exit

sn6uv:
I'm not quite sure what the implications of changing the functionality of `Quit` are but this is more consistent with Mathematica.


sn6uv:
Does not work in the web interface.

PR #541, opened 2016-09-08
Overhaul of image.py

poke1024:
- Removed dependency on skimage everywhere where it's not strictly necessary; everything except for morphology and edge detection now only relies on PIL and numpy
- Removed matplotlib dependency
- Extended Colorize[] to use arbitrary ColorData[] and color functions (also works on images now)
- Added support for reading JPEG EXIF data via Import["image.jpg", "RawExif"]
- Extended ImageTake[] for general cropping
- Added ColorCombine[]
- Added DominantColors[]
- Added ImageConvolve[]
- Reimplemented BoxMatrix, DiskMatrix, DiamondMatrix to match MMA
- Added docs

I've added the scipy license to README.rst, since ImageConvolve[] is basically a port of their code, and we might want to adapt some other of their stuff in the future; I'm not sure if this is the right way, though.

### DIFF
--- a/mathics/builtin/image.py
+++ b/mathics/builtin/image.py
@@ -145,7 +145,7 @@ class ImageImport(_ImageBuiltin):
         is_rgb = len(pixels.shape) >= 3 and pixels.shape[2] >= 3
         exif = Expression('List', *list(_Exif.extract(pillow, evaluation)))
 
-        atom = Image(pixels, 'RGB' if is_rgb else 'Grayscale', exif)  # FIXME
+        atom = Image(pixels, 'RGB' if is_rgb else 'Grayscale')
         return Expression(
             'List',
             Expression('Rule', String('Image'), atom),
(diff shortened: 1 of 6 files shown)


Piruzzolo:
Great job! Just note that, while our implementation of DominantColors uses k-means, MMA uses k-medoids, which may lead to slightly different behaviour on outliers.

poke1024:
This PR actually uses a bottom-up agglomerative clustering for `DominantColors`, namely `clusters.py:agglomerate()`, as I thought I read this approach out of the MMA docs: `"With MinColorDistance->d, clusters represented by colors c1 and c2 are merged if ColorDistance[c1,c2]<d. The color with the larger coverage is assigned to the new cluster."` (https://reference.wolfram.com/language/ref/DominantColors.html). This, to me, sounded like a bottom-up agglomerative clustering approach.

Piruzzolo:
These are the options I get after a bit of spelunking: 
`<|"Subsampling" -> True, Resampling -> "Nearest", "SubsamplingSize" -> {100, 100}, "NewClusteringFun" -> True,  "MinClusterNumber" -> 80,   "ClusteringMethod" -> "KMedoids",  "SameColorMerging" -> True, "SameColorDistance" -> 0.05,  "IterativeMerging" -> True,  "MinClusterDistance" -> 0.2, "LocalThresholding" -> True,  "ClusterCoverage" -> {0.03, Infinity},  "MinClusterStrength" -> 0.2|>`
Here `KMedoids` seems to be used. But it's just a guess, I don't know the exact meaning of each property.

poke1024:
It's not obvious to me how to use a KMedoids based method with a merging limit as in `MinColorDistance`, as the flavors of KMeans or KMedoids I came across do not involve any limits of that sort; a very obvious option would be to replace the initial step of using PIL's color palette reduction with a KMedoids method. Maybe I'll try this.

mathics/builtin/image.py
```diff
@@ -1056,22 +1405,196 @@ def _linearize(a):
...
+                return
+
+        if isinstance(color_coverage, Symbol) and color_coverage.get_name() == 'System`Automatic':
+            py_min_color_coverage = 0.05
+            py_max_color_coverage = 1.
+        elif color_coverage.has_form('List', 2):
+            py_min_color_coverage = color_coverage.leaves[0].round_to_float()
+            py_max_color_coverage = color_coverage.leaves[1].round_to_float()
+        else:
+            py_min_color_coverage = color_coverage.round_to_float()
```
sn6uv: should `py_max_color_coverage` also be set here?

mathics/builtin/image.py
```diff
@@ -1085,38 +1608,84 @@ def apply(self, image, stype, evaluation):
...
+            pixels = image.pixels[:py_n]
+        elif py_n < 0:
+            pixels = image.pixels[py_n:]
+        return Image(pixels, image.color_space)
+
+    def apply_rows(self, image, r1, r2, evaluation):
+        'ImageTake[image_Image, {r1_Integer, r2_Integer}]'
+        py_r1 = max(r1.get_int_value() - 1, 0)
+        py_r2 = max(r2.get_int_value() - 1, 0)
+        return Image(image.pixels[py_r1:1 + py_r2], image.color_space)
```
sn6uv: also need to check the indices are valid `0 <= py_r1 <= py_r2 < len(image.pixels)`. At the moment it looks like we're just checking `py_ri >= 0`.

mathics/builtin/image.py
```diff
@@ -1085,38 +1608,84 @@ def apply(self, image, stype, evaluation):
...
+
+    def apply_rows_cols(self, image, r1, r2, c1, c2, evaluation):
+        'ImageTake[image_Image, {r1_Integer, r2_Integer}, {c1_Integer, c2_Integer}]'
+        py_r1 = max(r1.get_int_value() - 1, 0)
+        py_r2 = max(r2.get_int_value() - 1, 0)
+        py_c1 = max(c1.get_int_value() - 1, 0)
+        py_c2 = max(c2.get_int_value() - 1, 0)
+        return Image(image.pixels[py_r1:1 + py_r2, py_c1:1 + py_c2], image.color_space)
```
sn6uv: indices here too

PR #544, opened 2016-09-11
XML Import

poke1024:
Still a work in progress, but basically working already.


poke1024:
The one serious problem with this it that `Import["ExampleData/InventionNo1.xml", "XMLObject"]` freezes and never seems to return, since it gets stuck seemingly indefinitely in the `MakeBoxes` phase (I didn't measure it, I always gave up). Also see https://github.com/mathics/Mathics/pull/385.

PR #557, opened 2016-09-16
Two bug fixes for images

poke1024:
- Images with alpha channels are now displayed correctly
- Loading jpegs with missing exif data no longer crashes


sn6uv:
Could you add a couple of test for the (previously failing) behaviour?

PR #559, opened 2016-09-16
a basic version of Nearest[]

poke1024:
Only the `Scan` method is implemented here, but it's still quite useful, I think.

### DIFF
--- a/mathics/builtin/lists.py
+++ b/mathics/builtin/lists.py
@@ -4230,6 +4230,9 @@ class Nearest(Builtin):
 
             py_distances = [(_to_real_distance(d), i) for i, d in enumerate(distances.leaves)]
 
+            if py_r is not None:
+                py_distances = [(d, i) for d, i in py_distances if d <= py_r]
+
             def pick():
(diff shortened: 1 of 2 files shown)


mathics/builtin/lists.py
```diff
@@ -4097,3 +4105,129 @@ def apply_manual_k(self, p, k, evaluation, options):
...
+            py_distances = [(_to_real_distance(d), i) for i, d in enumerate(distances.leaves)]
+
+            def pick():
+                if py_n is None:
+                    candidates = sorted(py_distances)
+                else:
+                    candidates = heapq.nsmallest(py_n, py_distances)
+
+                for d, i in candidates:
+                    if py_r is not None and d > py_r:
```
sn6uv: I think we should filter before finding the `py_n` smallest:

1. Avoids comparison with `None` in heapq.nsmallest
2. If k elements are within the limit distance the runtime is (very) slightly better: O(n) + O(min(k, m) log(k)) vs O(m log (n)) for m nearest elements.

PR #560, opened 2016-09-16
ByteCount using asizeof

Piruzzolo:
Implements ByteCount using the BSD-licensed asizeof module, which scans through the object tree and gives the total byte count of it. The results may differ between Python implementations and versions.

### DIFF
--- a/mathics/builtin/structure.py
+++ b/mathics/builtin/structure.py
@@ -15,7 +15,6 @@ from mathics.builtin.lists import (python_levelspec, walk_levels,
 from mathics.builtin.functional import Identity
 import six
 from six.moves import range
-from sys import getsizeof
 
 
 class Sort(Builtin):
@@ -1262,13 +1261,3 @@ class Through(Builtin):
         for leaf in args.get_sequence():
             items.append(Expression(leaf, *x.get_sequence()))
         return Expression(p, *items)
-
-
-
-def ByteCount(Builtin):
-
-
-
-    def apply(self, expression, evaluation):
-        'ByteCount[expression_]'
-        return Integer(getsizeof(expression))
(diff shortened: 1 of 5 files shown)


Piruzzolo:
see #561 for memory issues found

Piruzzolo:
@sn6uv There seems to be a problem with pypy build. Could you please take a look? Maybe pypy is not compatible with CPython in this case?

mathics/builtin/structure.py
```diff
@@ -1276,5 +1281,10 @@ class ByteCount(Builtin):
 
     def apply(self, expression, evaluation):
         'ByteCount[expression_]'
-
-        return Integer(count_bytes(expression))
+        if bytecount_support == False:
```
sn6uv: I thought it was just PyPy that didn't support bytecount?
In that case:
```
if platform.python_implementation == 'PyPy':
    bytecount_supporty = False
else:
    import ...
    bytecount_support = True
```
and then
```
if not bytecount_support:
    return evaluation.message('ByteCount', 'pypy')
```
would be simpler.

Piruzzolo: As of january 2016, Jython [has](https://mail.python.org/pipermail/jython-checkins/2016-January/001669.html) getsizeof() support too. I think it's just PyPy.

Piruzzolo: I think it's just PyPy. Jython too [supports](https://mail.python.org/pipermail/jython-checkins/2016-January/001669.html) sys.getsizeof().

TiberiumPY:
PyPy will not support getsizeof in future - because this is hard to implement and there's no good way to get REAL memory usage (because of custom GC and JIT)

mathics/builtin/structure.py
```diff
@@ -16,11 +16,12 @@
 import six
 import platform
 from six.moves import range
-try:
+
+if platform.python_implementation == 'PyPy':
```
sn6uv: I think this needs to be a function call, .i.e.
```
>>>> platform.python_implementation()
'PyPy'
```

PR #566, opened 2016-09-18
Make GatherBy efficient

poke1024:
In contrast to `Gather`, `GatherBy` had quadratic runtime until now. This changes it to linear.


Piruzzolo:
[Here](http://mathematica.stackexchange.com/questions/30328/underlying-algorithms-for-list-manipulation-functions) there's an interesting discussion about list function implementation. It May be useful to further optimize GatherBy, as well as the other ones :)

PR #569, opened 2016-09-20
optimise predetermined pattern lengths

sn6uv:
Early pattern matching exit if the length of the match is wrong. Currently the pattern matcher will keep trying to match even though it's clear from the number of arguments that no match exists. This patch fixes the simple case when the pattern length is precisely determined.

### Setup:
```
In[1]:= SetAttributes[g, Orderless]
In[2]:= g[a_, b_, c_Integer, d_Integer, e_, f_, g_, h_List] := {c, d}
```
### Before:
```
In[3]:= g[1,2,3,4,5,6,7,8,9,{}]; // AbsoluteTiming
Out[3]= {0.332413, Null}
```
### After:
```
In[3]:= g[1,2,3,4,5,6,7,8,9,{}]; // AbsoluteTiming
Out[3]= {0.000965118, Null}
```

### DIFF
--- a/mathics/core/pattern.py
+++ b/mathics/core/pattern.py
@@ -163,8 +163,13 @@ class ExpressionPattern(Pattern):
             #    next_leaves = self.leaves[1:]
 
             # early exit if the number of leaves does not match the predetermined length
-            if self._fixed_nargs is not None and len(expression.get_leaves()) != self._fixed_nargs:
-                return
+            if self._fixed_nargs is not None:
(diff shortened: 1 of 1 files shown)


sn6uv:
Looks like the overhead actually makes things worse, there's a 3% slowdown on the benchmarks.

sn6uv:
Although this fixes one pathological case, I don't think it's worth the overall slowdown.

PR #572, opened 2016-09-21
Fixes Import[] with XML files

poke1024:
Fixes e.g. `Import["http://ota.ox.ac.uk/text/5700.xml"]`, which was somehow broken.


poke1024:
replaced by https://github.com/mathics/Mathics/pull/573

PR #455, opened 2016-07-15
implement NumberForm

sn6uv:
**TODO:**
 - [x] fix passing functions as options e.g. `ExponentFunction`.
 - [x] `Quotient`
 - [x] handle `Integer` expr.
 - [x] zero case.
 - [x] replace `Real.make_boxes`.
 - [ ] `ScientificForm`, `AccountingForm`, `EngineeringForm`?

PR #574, opened 2016-09-22
Web interface for release 1.0

jaycech3n:
Release 1.0 should probably have a spiffier interface. In this PR:

- Cleaned up CSS
- Styling with CSS instead of GIFs
- Layout changes
- Improved welcome area UI
- Font Awesome
- Favicons
- New logo

Probably a couple other things I've forgotten. Check the logs for deets.

**NOTE** that this fork was based on `jaycech3n:undo` (refer #511), but the undo code has been reverted in this PR until more bugs are smoothed out.

---

Some screenshots:
screenshot from 2016-09-22 14 37 36
screenshot from 2016-09-22 14 18 21
screenshot from 2016-09-22 14 13 13


sn6uv:
I really like this! The only complaint I have is that the docs now require 2 clicks before you can begin searching. Could we either:

1. put the cursor in the search bar immediately upon clicking the question mark
2. keep the search bar at the top

jaycech3n:
Went with option 1.

Piruzzolo:
Is there a way to display numbers grouping each third digit? In MMA big numbers look far more readable than in Mathics, so far. Of course, formatting very big numbers would also require some optimization to make the loading faster (something like incremental loading)

sn6uv:
@Piruzzolo the work for digit grouping was added in #455 with `NumberForm`, e.g.
```
In[1]:= NumberForm[20!, DigitBlock -> 3]
Out[1]= 2,432,902,008,176,640,000
```
I think it's just a matter of turning it on by default for the web interface.

PR #582, opened 2016-09-25
Option to use sans serif fonts for MathML

poke1024:
I recently found that I like sans serif MathML fonts much better as they match the general Jupyter theme I use:

jupyter1
jupyter2

Maybe this is also interesting for https://github.com/mathics/Mathics/pull/574, as this seems to use sans serif fonts for the code input? Here's what it looks like in the non-spiffified `mathicsserver`:

mathicsserver

I wonder where this should be configured;`settings.py` doesn't seem a good place to me.

### DIFF
--- a/mathics/builtin/inout.py
+++ b/mathics/builtin/inout.py
@@ -1783,6 +1783,8 @@ class MathMLForm(Builtin):
             xml = ''
         # mathml = '<math><mstyle displaystyle="true">%s</mstyle></math>' % xml
         # #convert_box(boxes)
+        if True:  # configuration to be defined
+            xml = '<mstyle mathvariant="sans-serif">%s</mstyle>' % xml
         mathml = '<math>%s</math>' % xml  # convert_box(boxes)
         return Expression('RowBox', Expression('List', String(mathml)))
 
--- a/mathics/builtin/inout.py
+++ b/mathics/builtin/inout.py
@@ -658,7 +658,7 @@ class GridBox(BoxConstruct):
      = \begin{array}{cc} a & b\\ c & d\end{array}
 
     #> MathMLForm[TableForm[{{a,b},{c,d}}]]
-     = <math><mstyle mathvariant="..."><mtable columnalign="center">
+     = <math display="block"><mstyle mathvariant="..."><mtable columnalign="center">
      . <mtr><mtd columnalign="center"><mi>a</mi></mtd><mtd columnalign="center"><mi>b</mi></mtd></mtr>
      . <mtr><mtd columnalign="center"><mi>c</mi></mtd><mtd columnalign="center"><mi>d</mi></mtd></mtr>
      . </mtable></mstyle></math>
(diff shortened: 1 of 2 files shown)


Piruzzolo:
I like it too. We may use it as default output and use the current serif font for output in TraditionalForm.

Piruzzolo:
I think the fraction doesn't look very good in Jupyter, unlike the traditional interface. It should have the same font size as the test of the expression. In the traditional interface, the fraction bar seems a bit too think. Is there any option to control these behaviours in MathML?

poke1024:
Turns out the fraction size depends on the `display` type of the `<math>` tag; if it's `inline` the fractions are vertically compressed, if it's `block` they are as they are in the traditional interface. As we do not specify `display` for `<math>`, I suppose Jupyter's default is different from the Mathics traditional default. So I now added `display="block"` to all `<math>` output to fix this:

jup2

Piruzzolo:
Looks better now 👍

mathics/builtin/inout.py
```diff
@@ -1783,7 +1783,9 @@ def apply_mathml(self, expr, evaluation):
             xml = ''
         # mathml = '<math><mstyle displaystyle="true">%s</mstyle></math>' % xml
         # #convert_box(boxes)
-        mathml = '<math>%s</math>' % xml  # convert_box(boxes)
+        if True:  # configuration to be defined
+            xml = '<mstyle mathvariant="sans-serif">%s</mstyle>' % xml
```
sn6uv: can we add this as a settings option. Perhaps `XML_FONT = None` by default.

GarkGarcia: @poke1024 I belive this is the only thing preventing us from merging the PR (?)

rocky: @GarkGarcia @sn6uv From a user standpoint, what is a "settings option" and how does one add this? 

I like this feature.

mmatera:
Would be possible to choose the output format as a command from inside Mathics? 
Let's say
$UseSansserif=True;
from mathics turn the feature on, and

$UseSansserif=False;
brings to the default behavior.

I will try to take this on the weekend.

rocky:
Good. The Mathematica way seems to be to do this as part of the web interface: https://community.wolfram.com/groups/-/m/t/402640

mathics/builtin/inout.py
mmatera: Please see PR #908

axkr:
Not sure, but it seems that here:
https://github.com/mathics/Mathics/blob/a2b7f78184bf971e92e9ebac05febb6ab533a8f4/mathics/web/media/js/mathics.js#L233

```
mstyle.setAttribute('mathvariant', 'sans-serif');
```

is also necessary?

rocky:
@axkr It would be helpful if you could upload before and after screenshots.

@mmatera  I've been thinking about what you were saying about adding a `PyMathics` context and was wondering if $SansSerif too shouldn't be in some context. For example we could add a `Mathics` context for those things that are specific to Mathics and not in WL. I don't know. What do you think?

PR #585, opened 2016-09-27
Fix for natlang

mmatera:


mathics/builtin/assignment.py
```diff
@@ -1031,23 +1031,37 @@ class Quit(Builtin):
...
 
         evaluation.definitions.set_user_definitions({})
         return Symbol('Null')
 
+    def apply2(self, evaluation):
+        'Quit'
+
+        evaluation.definitions.set_user_definitions({})
+        return Symbol('Null')
+
```
sn6uv: Looks like there are two things being changed in this PR. Can you split this up into just the natlang changes please?

mathics/builtin/natlang.py
```diff
@@ -119,6 +119,9 @@ def generator(doc):
         'Adjective': [nltk.corpus.wordnet.ADJ, nltk.corpus.wordnet.ADJ_SAT],
         'Adverb': [nltk.corpus.wordnet.ADV],
     }
+except LookupError:
+    _wordnet_pos_to_type = {}
+    _wordnet_type_to_pos = {}
```
sn6uv: We should probably print a warning here so the user can fix the issue by installing the missing corpus.

mmatera:
Sorry, it seems I'm still a newby on github. Yes, the pr was about the second change. I think the first one leaked from an older version of my repo. I will add the warning and fix the pr.
Cheers 

M

Enviado desde mi teléfono inteligente Sony Xperia™

---- Angus Griffith escribió ----

PR #588, opened 2016-09-27
LogisticSigmoid

Piruzzolo:
Basic implementation of LogisticSigmoid


mathics/builtin/exptrig.py
```diff
@@ -1120,3 +1053,27 @@ def apply_xy_dx(self, x, y, dx, dy, steps, evaluation):
...
+    >> LogisticSigmoid[0.5 + 2.3 I]
+     = 1.06475 + 0.808177 I
+
+    >> LogisticSigmoid[{-0.2, 0.1, 0.3}]
+     = {0.450166, 0.524979, 0.574443}
+    """
+
+    attributes = ('Listable', 'NumericFunction',)
+    
+    rules = {'LogisticSigmoid[z_?NumberQ]': 'Power[Plus[1,Power[E,Times[-1,z]]],-1]'}
```
sn6uv: Note: it should be possible to write these rules in regular `InputForm` syntax, i.e. `1 / (1 + Exp[-z])`, which I think is a bit clearer.

Also, I think we should only apply this rule when the argument is real, `LogisticSigmoid[I Pi]` should remain unevaluated (good idea to add this as a test).

mathics/builtin/exptrig.py
```diff
@@ -3,7 +3,6 @@
 
 """
 Exponential, trigonometric and hyperbolic functions
-
```
sn6uv: Looks like a bunch of whitespace changes got applied by your editor. Can you fix that up please?

mathics/builtin/exptrig.py
```diff
@@ -1120,3 +1053,27 @@ def apply_xy_dx(self, x, y, dx, dy, steps, evaluation):
...
+    
+    >> LogisticSigmoid[0.5]
+     = 0.622459
+
+    >> LogisticSigmoid[0.5 + 2.3 I]
+     = 1.06475 + 0.808177 I
+
+    >> LogisticSigmoid[{-0.2, 0.1, 0.3}]
+     = {0.450166, 0.524979, 0.574443}
+    """
```
sn6uv: examples / docs look good

mathics/builtin/exptrig.py
Piruzzolo: `LogisticSigmoid[0.5+2.3 I]` returns `1.06475 +0.808177 I` in MMA, but `LogisticSigmoid[I Pi]`, as well as `LogisticSigmoid[1+I Pi]`, remain unevaluated. So I think the check here should be is_exact or something like that.

sn6uv: Okay, I think how it's implemented now is fine. We can always add inexact complex numbers later.

mathics/builtin/exptrig.py
```diff
@@ -1070,10 +1133,13 @@ class LogisticSigmoid(Builtin):
 
     >> LogisticSigmoid[{-0.2, 0.1, 0.3}]
      = {0.450166, 0.524979, 0.574443}
+
+    #> LogisticSigmoid[LogisticSigmoid[I Pi]]
```
sn6uv: typo here?

PR #591, opened 2016-09-27
Fixes nested Modules (again)

poke1024:
Yesterday's fix for nested Modules (https://github.com/mathics/Mathics/pull/584) was not really correct. I added `evaluation` to the `replace_vars` pipeline, which is not a bad thing in itself, I guess. The problem is, though, that `Module` should not actually evaluate at that point, but only later when it gets to `new_expr.evaluate(evaluation)`. If it evaluates earlier, nested modules get evaluated, which is not what we want. I added this rather funky test case to demonstrate what happens if evaluation happens too early (it calls `Abort[]`).

### DIFF
--- a/mathics/builtin/scoping.py
+++ b/mathics/builtin/scoping.py
@@ -24,7 +24,8 @@ def get_scoping_vars(var_list, msg_symbol='', evaluation=None):
         if var.has_form('Set', 2):
             var_name = var.leaves[0].get_name()
             new_def = var.leaves[1]
-            new_def = new_def.evaluate(evaluation)
+            if evaluation:
+                new_def = new_def.evaluate(evaluation)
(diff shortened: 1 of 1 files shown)


sn6uv:
Should I revert the changes in #584 or are they still needed?

poke1024:
The changes in https://github.com/mathics/Mathics/pull/584 are actually not needed. `replace_vars` will never need an `evaluation` for this fix.

sn6uv:
I've reverted e7e51ddbaabb345a8466d7f843286b5824eeca06 but kept the rest.

mmatera:
I proposed the change because Mathics crashes on my system without it
M

Enviado desde mi teléfono inteligente Sony Xperia™

---- Angus Griffith escribió ----

PR #593, opened 2016-09-27
Fix for natlang v0.2 + warning

mmatera:


mathics/builtin/natlang.py
```diff
@@ -119,6 +119,10 @@ def generator(doc):
         'Adjective': [nltk.corpus.wordnet.ADJ, nltk.corpus.wordnet.ADJ_SAT],
         'Adverb': [nltk.corpus.wordnet.ADV],
     }
+except LookupError:
+    print("warning: nltk dictionary not installed\n")
```
sn6uv: print isn't quite the correct way to handle this. We need to use `evaluation.message` to get the messaging appearing in all the Mathics interfaces. Furthermore, it looks like the parts of the code using these dicts don't expect them to be empty (the `ImportError` case is okay because `requires` prevents that code from running). Does this change fix the issue you're having?

poke1024:
@mmatera I tried to rewrite your fix with https://github.com/mathics/Mathics/pull/594, since it's a bit tricky as @sn6uv pointed out. Can you check if this fixes your test case? For me the `LookupError` seemed to occur if NLTK was installed, but the wordnet corpora were not found in the nltk_data folders.

mathics/builtin/natlang.py
sn6uv: To be more clear: What I think is going on is that the wordnet corpus is missing? Fixing the `LookupError` in this way will just move the error to a `KeyError` in the natural language functions. This is better but doesn't properly resolve the issue.

sn6uv:
fixed by #594

mmatera:
Hi,

I proposed that simple solution just to avoid the runtime error when mathics loads. Since this functionality depends on the proper  installation of a non basic library, maybe a cleaner way to solve the issue would be to  implement this functionality as a package, and load it just if the dependence is properly installed.

PR #571, opened 2016-09-21
more image docs and examples

sn6uv:
Fixes #492

PR #597, opened 2016-09-29
OutputSizeLimit terminal enhancements 

sn6uv:
Cleans up `$OutputSizeLimit` and also uses `boxes_to_text` over `boxes_to_xml` when using the terminal interface. Fixes some problems I was having with #571.

On second thought I decided to just use `boxes_to_text` everywhere since otherwise images etc. get hidden.


sn6uv:
@poke1024 do you mind jumping on slack? I'd like to chat with you about this.

sn6uv:
see #605

Issue #602, opened 2016-09-30
bug rendering equations in web interface


sn6uv:
see screenshot:
![2016-09-30-133602_2560x1440_scrot](https://cloud.githubusercontent.com/assets/1217643/19006165/095018ac-8713-11e6-9cee-39f49813a6cb.png)

first reported https://groups.google.com/forum/#!topic/mathics-users/3MO658UGaCo

sn6uv:
This is another `OutputSizeLimit` bug.

Piruzzolo:
Maybe worths introducing some kind of output percentual check or measurement to cut off not atomic or unsignificant parts? The problem is: how to tell a perceptually atomic expression from a generic one? LHS of rules should be cut only if very very big (the user expects to see them). <<1>>s could be all relaxed if doing so does not exceed an (adjustable) percent of the ouputlimit. The same holds for <<2>> and <<3>>. The adjustable percent may be set by looking at the expression form: if we have a list of compound expressions, we can get the relative weight of each and doing some sort of weighted mean. If they're all big numbers, then we can cut them all, showing the first few and last digits, at most. If it's a mixed list, I'd personally give priority to the expressions - if not relatively too long. Presumably, Mathematica does likewise (they say that expression formatting code is hundreds pages long). Of course, being Python 100x slower or so, we'd better find a compromise

sn6uv:
Complaints with the current system:
- should only omit middle elements (i.e. no omissions when <=2 leaves).
- should only omit when more than 1000 characters are to be printed (e.g. the above example had `$OutputSizeLimit = 1000`).
- doesn't count heads or spacing between items towards the limit.
- expressions with a large XML representation (e.g. Images) are almost always omitted.
- length of `<<1>>` isn't taken into account.

Perhaps we should revert #385 and #581 until this is sorted out.

sn6uv:
workaround #611

GarkGarcia:
This has now been fixed.

PR #605, opened 2016-10-01
Improvements to tackle wrong expression outputs

poke1024:
see https://github.com/mathics/Mathics/issues/602

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -1187,7 +1187,7 @@ class Expression(BaseExpression):
             total_cost = 2 + cost_of_leaves + separator_cost * n_separators  # {a, b, c}, [a, b, c]
 
             if name != 'System`List':
-                total_cost += self.head.output_cost() + separator_cost * n_separators
+                total_cost += self.head.output_cost()
 
             return total_cost
 
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -62,6 +62,15 @@ def system_symbols_dict(d):
     return {ensure_context(k): v for k, v in six.iteritems(d)}
 
 
+_layout_boxes = system_symbols(
+    'RowBox',
+    'SuperscriptBox',
+    'SubscriptBox',
+    'SubsuperscriptBox',
+    'FractionBox',
+    'SqrtBox')
+
+
 class BoxError(Exception):
     def __init__(self, box, form):
         super(BoxError, self).__init__(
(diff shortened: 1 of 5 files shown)


poke1024:
You can get write fancy with an `$OutputSizeLimit` of `1000` now:

example1

Eventually, the limit will still kick in:

example2

We could issue a warning via `evaluate.message` in these cases, telling the user to bump the `$OutputSizeLimit`. We could also use `MakeBoxes` to make the omit tags like `<<2>>` appear in a slightly color/background.

poke1024:
With warning:

example3

mathics/core/expression.py
```diff
@@ -1155,6 +1167,21 @@ def block(tex, only_subsup=False):
...
+    def output_cost(self):
+        name = self.get_head_name()
+
+        if name in ('System`ImageBox', 'System`GraphicsBox', 'System`Graphics3DBox'):
+            return 0  # always display
+
+        cost = sum(leaf.output_cost() for leaf in self.leaves)
+
+        if name == 'System`List':
+            return 2 + cost + len(self.leaves)  # {a, b, c}
```
sn6uv: would it be better to compute the sum cost of leaves? Also, what about the size of ", " between leaves?

mathics/core/expression.py
```diff
@@ -1155,6 +1167,21 @@ def block(tex, only_subsup=False):
...
+        elif name in _layout_boxes:
+            return cost
+        else:
+            return cost + 2 + self.head.output_cost() + len(self.leaves)  # XYZ[a, b, c]
```
sn6uv: here also

poke1024: `cost` is the sum cost of leaves (I renamed it now). `len(self.leaves)` measures the size of ",", which could be one or two (it assumes for now). we could pass the separator size into `make_boxes` and then pass it on to `output_cost`, but it's bulky and then should we really measure it that way? If `$OutputSizeLimit` is understood as number of printable characters (without spaces and layout elements like fractions, parentheses, ...), then it makes sense to always measure 1 for ",".

sn6uv: In that case I guess technically it should be `max(len(self.leaves) - 1, 0)`.

sn6uv:
It seems like MMA kernels don't even observe `$OutputSizeLimit` only the notebook interface does and the full boxes are still generated anyway.

Piruzzolo:
So our approach is more efficient! Good to know

mathics/core/expression.py
```diff
@@ -1173,15 +1176,20 @@ def output_cost(self):
...
+        if name in _layout_boxes:
             return cost_of_leaves
         else:
-            return cost_of_leaves + 2 + self.head.output_cost() + separator_cost * len(self.leaves)  # XYZ[a, b, c]
+            separator_cost = 1  # i.e. ","
+            n_separators = max(len(leaves) - 1, 0)
+            total_cost = 2 + cost_of_leaves + separator_cost * n_separators  # {a, b, c}, [a, b, c]
+
+            if name != 'System`List':
+                total_cost += self.head.output_cost() + separator_cost * n_separators
```
sn6uv: doesn't this count the separators twice now?

sn6uv:
This will need to be rebased on a revert of 8e14668.

mmatera:
I made a rebase in #1371. @poke1024 if you have the time and the will, look at it.

PR #614, opened 2016-10-05
Fix unicode decode problems

poke1024:
Using `Import` on a text file with wrong unicode sequences currently crashes Mathics. This works around the problem, allowing the file to be imported with the problematic characters being replaced by whitespace.


sn6uv:
Great!

PR #616, opened 2016-10-07
Speedups by avoiding inner classes

poke1024:
Turns out that, at least in CPython, declaring `class`es inside functions is quite slow.

Running `x = Range[50000]; First[Timing[Position[x, 10000]]]` without this PR gives `1.16326`, with this PR, it gives `0.461211`. This is only due to one inner `class` definition.

Running the profiler, one actually sees Python's class build function popping up. So this PR moves inner classes out of functions at several places.

### DIFF
--- a/mathics/builtin/patterns.py
+++ b/mathics/builtin/patterns.py
@@ -524,6 +524,10 @@ class Alternatives(BinaryOperator, PatternObject):
         return range
 
 
+class _StopGeneratorExcept(StopGenerator):
+    pass
+
+
 class Except(PatternObject):
     """
     <dl>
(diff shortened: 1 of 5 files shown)


mathics/builtin/patterns.py
```diff
@@ -1332,3 +1334,25 @@ def _match(leaf):
...
+    try:
+        form.match(yield_match, expr, {}, evaluation, fully=False)
+    except _StopGeneratorBaseExpressionIsFree as exc:
+        return exc.value
+
+    if expr.is_atom():
+        return True
+    else:
+        return expr.head.is_free(form, evaluation) and all(
+            leaf.is_free(form, evaluation) for leaf in expr.leaves)
```
sn6uv: is this function related to the change?

poke1024: I forgot to hook this up. It's done now. I needed to move this function into `patterns.py` as the declaration of `_StopGeneratorBaseExpressionIsFree` needs `StopGenerator` which cannot be imported into `core.expression.py` as `patterns.py` depends on `core.expression.py`.

PR #618, opened 2016-10-12
Information

mmatera:
Finally, I could add the prefix operators ? and ?? with its standard mea ning in mma: ? F is equivalent to Definition[F] and ??F calls Information[F].
I also added the command Information, which for built in symbols, shows the **doc** string of the correspondent class, and for user symbols, looks to the message ::usage.

Something that is missing is to show inside Information the output of Definition, but for some reason it didn't work. Anyway, I think the basic functionality is implemented.

### DIFF
--- a/mathics/builtin/assignment.py
+++ b/mathics/builtin/assignment.py
@@ -2,12 +2,13 @@
 # -*- coding: utf-8 -*-
 
 
-
+import mathics.builtin
 from mathics.builtin.base import (
     Builtin, BinaryOperator, PostfixOperator, PrefixOperator)
 from mathics.core.expression import (Expression, Symbol, valid_context_name,
-                                     system_symbols)
-from mathics.core.rules import Rule
+                                     system_symbols, String)
+from mathics.core.rules import Rule, BuiltinRule
+from mathics.builtin.patterns import RuleDelayed
 from mathics.builtin.lists import walk_parts
 from mathics.core.evaluation import MAX_RECURSION_DEPTH, set_python_recursion_limit
 
@@ -707,6 +708,7 @@ class Definition(Builtin):
     """
 
     attributes = ('HoldAll',)
+    precedence = 670
 
     def format_definition(self, symbol, evaluation, grid=True):
         'StandardForm,TraditionalForm,OutputForm: Definition[symbol_]'
@@ -787,8 +789,199 @@ class Definition(Builtin):
 
     def format_definition_input(self, symbol, evaluation):
         'InputForm: Definition[symbol_]'
-
         return self.format_definition(symbol, evaluation, grid=False)
+
+
+def _get_usage_string(symbol, evaluation, htmlout=False):
+    '''
+    Returns a python string with the documentation associated to a given symbol.
+    '''
+    builtins = mathics.builtin.builtins
+    # from mathics.builtin import builtins
+
+    definition = evaluation.definitions.get_definition(symbol.name)
+    ruleusage = definition.get_values_list('messages')
+    usagetext = None
+    import re
+    bio = builtins.get(definition.name)
+
+    if bio is not None:
+        from mathics.doc.doc import Doc
+        if htmlout:
+            usagetext = Doc(bio.__class__.__doc__).text(0)
+        else:
+            usagetext = Doc(bio.__class__.__doc__).text(0)
+        usagetext = re.sub(r'\$([0-9a-zA-Z]*)\$', r'\1', usagetext)
+
+    # Looks for the "usage" message. For built-in symbols, if there is an "usage" chain, overwrite the __doc__ information.
+    for rulemsg in ruleusage:
+        if rulemsg.pattern.expr.leaves[1].__str__() == "\"usage\"":
+            usagetext = rulemsg.replace.value
+    return usagetext
+
+
+class Information(PrefixOperator):
+    """
+    <dl>
+    <dt>'Information[$symbol$]'
+        <dd>Prints information about a $symbol$
+    </dl>
+    'Information' does not print information for 'ReadProtected' symbols.
+    'Information' uses 'InputForm' to format values.
+
+    >> a = 2;
+    >> Information[a]
+     = a = 2
+
+
+    >> f[x_] := x ^ 2
+    >> g[f] ^:= 2
+    >> f::usage = "f[x] returns the square of x";
+    >> Information[f]
+     = f[x] returns the square of x
+     .
+     . f[x_] = x ^ 2
+     .
+     . g[f] ^= 2
+
+    >> ? Table
+     = 
+     .   'Table[expr, {i, n}]'
+     .     evaluates expr with i ranging from 1 to n, returning
+     . a list of the results.
+     .   'Table[expr, {i, start, stop, step}]'
+     .     evaluates expr with i ranging from start to stop,
+     . incrementing by step.
+     .   'Table[expr, {i, {e1, e2, ..., ei}}]'
+     .     evaluates expr with i taking on the values e1, e2,
+     . ..., ei.
+     .      
+
+    >> Information[Table]
+     = 
+     .   'Table[expr, {i, n}]'
+     .     evaluates expr with i ranging from 1 to n, returning
+     . a list of the results.
+     .   'Table[expr, {i, start, stop, step}]'
+     .     evaluates expr with i ranging from start to stop,
+     . incrementing by step.
+     .   'Table[expr, {i, {e1, e2, ..., ei}}]'
+     .     evaluates expr with i taking on the values e1, e2,
+     . ..., ei.
+     .
+     . Attributes[Table] = {HoldAll, Protected}
+     .
+    """
+
+    operator = "??"
+    precedence = 670
+    attributes = ('HoldAll', 'SequenceHold', 'Protect', 'ReadProtect')
+    messages = {'notfound': 'Expression `1` is not a symbol'}
+    options = {'LongForm': 'True', }
+
+    def format_definition(self, symbol, evaluation, options, grid=True):
+        'StandardForm,TraditionalForm,OutputForm: Information[symbol_, OptionsPattern[Information]]'
+        from mathics.core.expression import from_python
+        lines = []
+
+        if isinstance(symbol, String):
+            evaluation.print_out(symbol)
+            evaluation.evaluate(Expression('Information', Symbol('System`String')))
+            return
+
+        if not isinstance(symbol, Symbol):
+            evaluation.message('Information', 'notfound', symbol)
+            return Symbol('Null')
+
+        # Print the "usage" message if available.
+        usagetext = _get_usage_string(symbol, evaluation)
+        if usagetext is not None:
+            lines.append(String(usagetext))
+
+        if self.get_option(options, 'LongForm', evaluation).to_python():
+            self.show_definitions(symbol, evaluation, lines)
+
+        if grid:
+            if lines:
+                return Expression(
+                    'Grid', Expression(
+                        'List', *(Expression('List', line) for line in lines)),
+                    Expression(
+                        'Rule', Symbol('ColumnAlignments'), Symbol('Left')))
+            else:
+                return Symbol('Null')
+        else:
+            for line in lines:
+                evaluation.print_out(Expression('InputForm', line))
+            return Symbol('Null')
+
+        # It would be deserable to call here the routine inside Definition, but for some reason it fails...
+        # Instead, I just copy the code from Definition
+
+    def show_definitions(self, symbol, evaluation, lines):
+        def print_rule(rule, up=False, lhs=lambda l: l, rhs=lambda r: r):
+            evaluation.check_stopped()
+            if isinstance(rule, Rule):
+                r = rhs(rule.replace.replace_vars(
+                        {'System`Definition': Expression('HoldForm', Symbol('Definition'))}))
+                lines.append(Expression('HoldForm', Expression(
+                    up and 'UpSet' or 'Set', lhs(rule.pattern.expr), r)))
+
+        name = symbol.get_name()
+        if not name:
+            evaluation.message('Definition', 'sym', symbol, 1)
+            return
+        attributes = evaluation.definitions.get_attributes(name)
+        definition = evaluation.definitions.get_user_definition(
+            name, create=False)
+        all = evaluation.definitions.get_definition(name)
+        if attributes:
+            attributes = list(attributes)
+            attributes.sort()
+            lines.append(Expression(
+                'HoldForm', Expression(
+                    'Set', Expression('Attributes', symbol), Expression(
+                        'List',
+                        *(Symbol(attribute) for attribute in attributes)))))
+
+        if definition is not None and 'System`ReadProtected' not in attributes:
+            for rule in definition.ownvalues:
+                print_rule(rule)
+            for rule in definition.downvalues:
+                print_rule(rule)
+            for rule in definition.subvalues:
+                print_rule(rule)
+            for rule in definition.upvalues:
+                print_rule(rule, up=True)
+            for rule in definition.nvalues:
+                print_rule(rule)
+            formats = sorted(definition.formatvalues.items())
+            for format, rules in formats:
+                for rule in rules:
+                    def lhs(expr):
+                        return Expression('Format', expr, Symbol(format))
+
+                    def rhs(expr):
+                        if expr.has_form('Infix', None):
+                            expr = Expression(Expression(
+                                'HoldForm', expr.head), *expr.leaves)
+                        return Expression('InputForm', expr)
+                    print_rule(rule, lhs=lhs, rhs=rhs)
+        for rule in all.defaultvalues:
+            print_rule(rule)
+        if all.options:
+            options = sorted(all.options.items())
+            lines.append(
+                Expression('HoldForm', Expression(
+                    'Set', Expression('Options', symbol),
+                    Expression('List', *(
+                        Expression('Rule', Symbol(name), value)
+                        for name, value in options)))))
+        return
+
+    def format_definition_input(self, symbol, evaluation, options):
+        'InputForm: Information[symbol_, OptionsPattern[Information]]'
+        return self.format_definition(symbol, evaluation, options, grid=False)
 
 
 class Clear(Builtin):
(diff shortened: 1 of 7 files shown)


mathics/builtin/assignment.py
```diff
@@ -792,10 +800,130 @@ def rhs(expr):
...
+def information_interpret_doc_string(ds):
+    # TODO: work a little bit the interpretation
+    ds = ds.split("\n")
+    out = ""
+    for s in ds:
+        if s[0:6]=="    >>":
+            # If we reach the test section, stop the loop.
+            break
+        out = out + "\n" + s
+    return out
```
sn6uv: This is not really a robust way to parse the docstrings. I think it's better to reuse the existing stuff in `mathics/doc`.

mathics/builtin/assignment.py
```diff
@@ -792,10 +800,130 @@ def rhs(expr):
...
+   
+    >> F[x_] := x^2
+    >> F::usage = "F[x] evaluates the square of x."
+    >> H[F[y_,u_]]^:=F[H[y],u]
+    >> Information[F]
+     |F[x] evaluates the square of x
+     = Null
+    """
+    operator="??"
+    precedence=5001
```
sn6uv: Is this correct?

mathics/core/parser/operators.py
```diff
@@ -17,6 +17,8 @@
     'Exists': 240,
     'NotExists': 240,
     'Not': 230,
+    'Information': 5001,
+    'Definition' : 5000,
```
sn6uv: What's the correct precedence of these operators?

mathics/core/parser/tokeniser.py
```diff
@@ -206,6 +206,9 @@
     ('Therefore', r' \u2234 '),
     ('Because', r' \u2235 '),
     ('Backslash', r' \u2216 '),
+
+    ('Information', r'\?\?'),
+    ('Definition', r'\?'),
```
sn6uv: conflicts with `PatternTest` token.

mathics/core/parser/tokeniser.py
```diff
@@ -230,7 +233,7 @@
     '<': ['Get', 'StringJoin', 'LessEqual', 'Less'],
     '=': ['SameQ', 'UnsameQ', 'Equal', 'Unset', 'Set'],
     '>': ['PutAppend', 'Put', 'GreaterEqual', 'Greater'],
-    '?': ['PatternTest'],
+    '?': ['Information', 'Definition','PatternTest'],
```
sn6uv: Rather than adding an additional token we should reuse the `PatternTest` token for `Information`.

mathics/main.py
```diff
@@ -246,6 +249,22 @@ def main():
...
+                query = evaluation.parse_feeder(feeder)
+                if query is None:
+                    continue
+                evaluation.evaluate(query, timeout=settings.TIMEOUT)
+        except (KeyboardInterrupt):
+            print('\nKeyboardInterrupt')
+        
+        definitions.set_line_no(0)
+
+
```
sn6uv: Could you split this `initfile` stuff out into a separate PR please.

mathics/builtin/assignment.py
```diff
@@ -9,8 +9,10 @@
 from mathics.builtin.base import (
     Builtin, BinaryOperator, PostfixOperator, PrefixOperator)
 from mathics.core.expression import (Expression, Symbol, valid_context_name,
-                                     system_symbols)
-from mathics.core.rules import Rule
+                                     system_symbols,String)
+
```
sn6uv: Can you fix the formatting here. Should be: `system_symbols, String`

mathics/builtin/assignment.py
```diff
@@ -625,6 +627,11 @@ class Definition(Builtin):
     >> Definition[a]
      = a = 2
 
+    >> a = 2;
+    >> ? a
+     = a = 2
+
+
```
sn6uv: `Definition` is not related to `Information`:

- `?a` is `Information["a", Rule[LongForm, False]]`
- `??a` is `Information["a", Rule[LongForm, True]]`

mathics/builtin/inout.py
```diff
@@ -1820,6 +1820,7 @@ def apply_tex(self, expr, evaluation):
             # Replace multiple newlines by a single one e.g. between asy-blocks
             tex = MULTI_NEWLINE_RE.sub('\n', tex)
 
+
```
sn6uv: there are a few whitespace changes that don't need to be included.

sn6uv (dismissed):

mathics/builtin/assignment.py
mmatera: I do not know if there is an standard value for the precedence of these operators. I just choose a very large one, because I supposed that in the context where these operators make sense, they should be evaluated with very high priority.

mathics/builtin/inout.py
mmatera: I notice that, but I do not know how to ask git to revert just a file, or ignore this kind of changes.

mathics/builtin/assignment.py
mmatera: OK

mathics/core/parser/tokeniser.py
sn6uv: My point is that `PatternTest` and `Information` need to share the ` \? ` token.

mathics/core/parser/tokeniser.py
sn6uv: So there are two tokens `('SingleQM',  ' \? ')` and `('DoubleQM', '  \?\? ')`. This part of the tokeniser looks at a single character and tries to determine what token will match.

Let's say the tokeniser sees `? next. The problem is we don't know which token to emit without looking at the next character. The next character is either `?` or it's something else. The solution is to first try matching `DoubleQM` and then try `SingleQM`.

Does that make sense?

mathics/builtin/assignment.py
sn6uv: True. It looks like they don't have a builtin precedence. `Precedence[Information]` gives the default of 670. You can experimenting to compare with other operators. You can wrap things in `Hold` to investigate.

For example:
```
In[2]:= Hold[a ? b ; c] // FullForm
Out[2]//FullForm= Hold[CompoundExpression[PatternTest[a, b], c]]
```
so the precedence is higher than `CompoundExpression`.

mathics/builtin/assignment.py
sn6uv: Which version of MMA are you using?
```
In[1]:= $Version
Out[1]= 11.0.1 for Linux x86 (64-bit) (September 21, 2016)

In[2]:= Hold[??a]
Out[2]= Hold[Information[a, LongForm -> True]]

In[3]:= Hold[?a]
Out[3]= Hold[Information[a, LongForm -> False]]
```

mmatera: I use the version 9, and Hold[?a] also returns Hold[Information[a, LongForm -> False]]
,  but the help says something different: https://reference.wolfram.com/language/ref/Definition.html

mathics/builtin/assignment.py
mmatera: Indeed, it seems to have smaller priority than Times, but larger than +...

In[1]:= Hold[??a + b]//FullForm
Out[1]//FullForm= Hold[Times[Information["a", Rule[LongForm, True]], Plus[b]]]

In[2]:= Hold[??a *b]//FullForm
Out[2]//FullForm= Hold[Information["a * b", Rule[LongForm, True]]]

mathics/core/parser/tokeniser.py
mmatera: OK, but then how we should implement it? I tried to imitate the behavior of @@ and >>, in a way that the tokeniser first look at the double operator, and afterward to the simple one.  This is why the change at line 236. What I do not realise is how and when the tokeniser distinguishes a prefix operator from an infix binary one...

mathics/builtin/assignment.py
mmatera: Now it calls Doc().__str__() from mathics.doc.doc to parse the docstrings.

mathics/main.py
mmatera: I moved this chunk of code to a new branch.

mathics/builtin/assignment.py
sn6uv: Not quite, it could still have a higher precedence than `*` just the second example has no possibility of a unary `Times`.

mathics/core/parser/parser.py
```diff
@@ -236,6 +236,12 @@ def parse_postfix(self, expr1, token, p):
     # Called with one Token and return a Node.
     # Used for prefix operators and brackets.
 
+    def p_PatternTest(self,token):
```
mmatera: This point was confusing for me: the name of the token has nothing to do with the prefix op...

mathics/core/parser/tokeniser.py
mmatera: I think I have solved this issue now.

mathics/builtin/assignment.py
sn6uv: I think the point in the docs is (it's unclear) that under the hood `Information` is calling `Definitions` to construct its output.

mathics/core/parser/tokeniser.py
sn6uv: The tokeniser does not distinguish between prefix and binary operators at all, that's the parsers job. For example, `1 + 2` will tokenise as a sequence `[Number, Plus, Number]` but `+3` would just be `[Plus Number]`.

The parser (in `core/parser/parser.py`) knows whether it's at the start of an expression or part-way through.

At the start of an expression it calls `parse_p` for parsing prefix expressions. It looks for a manual rule of the form `p_TOKENNAME` (in this case `p_Plus`). Alternatively it looks for the token in `prefix_ops` (defined in `core/parser/operators.py`) which is handled directly. If neither case applies it emits a syntax message.

In the current master `+3` finds `p_Plus` but `?Sin` can't find either a `p_PatternTest` nor find it in the `prefix_ops` dict so it emits the syntax error message that we currently see.

Does that make sense?

mathics/core/parser/parser.py
sn6uv: Yes, one of the quirks of the parser is that is uses the token name (called `tag` in the code) to construct the Node (see the `parse_p` method). This works well for most cases unless the token is reused across different functions with different names.

mathics/core/parser/parser.py
```diff
@@ -413,6 +419,18 @@ def p_Decrement(self, token):
         q = prefix_ops['PreDecrement']
         return Node('PreDecrement', self.parse_exp(q))
 
+
+    def p_Definition(self, token):
+        self.consume()
+        q = prefix_ops['Definition']
+        return Node('Definition', self.parse_exp(q))
```
sn6uv: I think we can remove this rule now right? It's been replaced with `p_PatternTest`.

mathics/core/parser/parser.py
```diff
@@ -413,6 +419,18 @@ def p_Decrement(self, token):
...
+
+    def p_Information(self, token):
+        self.consume()
+        q = prefix_ops['Information']
+        return Node('Information', self.parse_exp(q))
```
sn6uv: what about the `LongForm` option?

sn6uv:
One thing that's missing is the `LongForm` options. I'm not sure what the best way to do this is, probably just construct a sub node e.g. `Node('Rule', Symbol('LongForm'), Symbol('True'))` in the `p_` methods.

mathics/builtin/assignment.py
mmatera: Now ? F calls Information[F, LongForm->False]

mmatera:
Now the operators works as they should, without introduce further conflicts, and Information now accepts the option LongForm.  

The only problem now seems to be in the test strings. For some reason, the tester neglects some spaces in the test string, and then it see a mismatch between the expected and the obtained output. Maybe you can give me a clue about how to fix it.

sn6uv:
It looks like a bunch of the jupyter kernel stuff got committed here.

mathics/core/parser/operators.py
mmatera: Still not sure about this. What's the best? To assign the default precedence or to leave it as it is now? In principle, the last option seems to work.

mathics/core/parser/parser.py
```diff
@@ -459,6 +459,20 @@ def p_Decrement(self, token):
         q = prefix_ops['PreDecrement']
         return Node('PreDecrement', self.parse_exp(q))
 
+    def p_PatternTest(self,token):
+        self.consume()
+        q = prefix_ops['Definition']
+        child = self.parse_exp(q)
+        return Node('Information', child, Node('Rule',Symbol("LongForm"), Symbol("False")))
```
wolfv: nitpick: some formatting (space after comma)

mathics/core/parser/parser.py
```diff
@@ -459,6 +459,20 @@ def p_Decrement(self, token):
...
+
+    def p_Information(self, token):
+        self.consume()
+        q = prefix_ops['Information']
+        child = self.parse_exp(q)
+        if not child.__class__ is Symbol:
+            raise InvalidSyntaxError()
+        return Node('Information', child, Node('Rule',Symbol("LongForm"), Symbol("True")))
```
wolfv: same here :)

mathics/core/parser/tokeniser.py
```diff
@@ -228,7 +232,7 @@
     '<': ['Get', 'StringJoin', 'LessEqual', 'Less'],
     '=': ['SameQ', 'UnsameQ', 'Equal', 'Unset', 'Set'],
     '>': ['PutAppend', 'Put', 'GreaterEqual', 'Greater'],
-    '?': ['PatternTest'],
+    '?': ['PatternTest','Definition','Information'],
```
wolfv: spaces

wolfv:
let me know once this is done!

mmatera:
Now I am trying to decouple the patch from other changes that had the
version of Mathics over which I had implemented it. I tell you as soon I
have finished.

mathics/core/parser/parser.py
mmatera: I think all my changes now are mostly pep8 compliant, excepting for a few long lines.

mathics/builtin/assignment.py
mmatera: Now, it parses ? and ?? before + and *:

In[1]:= Hold[??a + b]
Out[1]= Hold[Null + b]

In[2]:= Hold[?a + b]
Out[2]= Hold[Null + b]

In[3]:= Hold[?? a * b]
Out[3]= Hold[Null b]

In[4]:= ? a
Out[4]= Null

I do not know if this current behavior makes more or less sense than the wolfram's

In[1]:= Hold[??a + b]'
Out[1]= Hold[Information[a, LongForm -> True]*Plus[b]]

In[2]:= Hold[?a + b]'
Out[2]= Hold[Information[a, LongForm -> False]*Plus[b]]

 In[3]:= Hold[?? a * b]
 Out[3]= Hold[Information[a * b, LongForm -> True]]

In[4]:= ? a

Information::notfound: Symbol a not found.


Notice that Wolfram's does not show Null Symbol as a return, so actually Information returns Null:

$ wolframscript -c "? Plus"
x + y + z represents a sum of terms. 
Null

Anyway, any of these behaviors should be fine, if there is no practical context in which they will be used...

mathics/core/parser/tokeniser.py
mmatera: In[1]:= a ? b//FullForm
Out[1]= PatternTest[a, b]

In[2]:= ? a b //FullForm
Out[2]= Times[b, Information[a, Rule[LongForm, False]]]


In Wolfram's

 In[1]:= a ? b //FullForm
 Out[1]//FullForm= FullForm[a?b]

 In[2]:= ? a b //FullForm
Global`a
Global`b 
 Out[2]//FullForm= FullForm[Null]

mathics/core/parser/tokeniser.py
mmatera: Done

mathics/core/parser/operators.py
mmatera: Seems to work as it is... we can try to change it to behave more like the weird way Wolfram's does it in weird cases, but at a practical level, it seems that does not matter.

mathics/builtin/assignment.py
mmatera: Done

mathics/builtin/inout.py
mmatera: Done

mathics/main.py
mmatera: Done

mathics/core/parser/tokeniser.py
mmatera: Fixed

mathics/builtin/assignment.py
mmatera: Indeed,

Mathics:

In[1]:= Hold[?? a + b] //FullForm
Out[1]= Hold[Plus[Information[a, Rule[LongForm, True]], b]]

In[2]:= Hold[? a + b]//FullForm
Out[2]= Hold[Plus[Information[a, Rule[LongForm, False]], b]]

In[3]:= Hold[?? a * b] //FullForm
Out[3]= Hold[Times[Information[a, Rule[LongForm, True]], b]]

In Wolfram's 
In[1]:= Hold[?? a + b] //FullForm
Out[1]//FullForm= FullForm[Hold[Information[a, LongForm -> True]*Plus[b]]]

In[2]:= Hold[? a + b]//FullForm
Out[2]//FullForm= FullForm[Hold[Information[a, LongForm -> False]*Plus[b]]]

In[3]:=  Hold[?? a * b] //FullForm
Out[3]//FullForm= FullForm[Hold[Information[a * b, LongForm -> True]]]


I don't know if it is really relevant to fix the precedence to reproduce this weird behavior.

mathics/core/parser/operators.py
mmatera: It works as it is

mmatera:
I think now it is ready for merging.

mmatera:
Done. And also I fixed the precedence of ? and ?? operators according to
wolfram's specification (=670).

M

GarkGarcia:
Great, it passed all tests! @mmatera Have the issues pointed out by @wolfv been addressed?

mmatera:
I guess I did...

PR #619, opened 2016-10-13
Fix quadratic runtimes in Mathics

poke1024:
This is continuation of #575 and #615. It's my cleaned up proposal for fixing quadratic runtimes in Mathics.

The following plots show the results of the accompanying two commands, run once using `master` (yellowish) and once using this PR (blue). Don't ask me where the spurious regular dots come from (garbage collection?).

`First[Timing[Fold[#1+#2&, Range[#]]]]& /@ Range[750]`

plot-1

```
MyTotal[x0_] := Module[{x = x0, y = 0}, While[Length[x] > 0, y = y + First[x]; x = Rest[x]]; y];
First[Timing[MyTotal[Range[#]]]]& /@ Range[500]
```

plot2

The changes are twofold:
- First, a new class called a `Structure` was introduced to enable list operations like `Most` or `Take` not to trigger an evaluation of every leaf again (this used to be #615).
- Second, the cache check is now based on a set of symbols rather than a full tree traversal (this used to be #575).

The quadratic runtimes are only fixed when both measures are applied, which is why I merged everything into this PR now.

All the information regarding cached `Expression` information (including what is `Expression._sequences` in `master`) is now bundled in `ExpressionCache`, which makes the code much cleaner I think.

I experimented with quite a variety of data structures, including `dict`s instead of `set`s for the symbol caching, but a `set` and a `sequences` list seems to be the best thing to do.

### DIFF
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -266,8 +266,8 @@ class Plus(BinaryOperator, SympyFunction):
                     leaves.append(last_item)
                 else:
                     if last_item.has_form('Times', None):
-                        last_item.leaves.insert(0, from_sympy(last_count))
-                        leaves.append(last_item)
+                        leaves.append(Expression(
+                            'Times', from_sympy(last_count), *last_item.leaves))
                     else:
                         leaves.append(Expression(
                             'Times', from_sympy(last_count), last_item))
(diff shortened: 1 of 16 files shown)


GarkGarcia:
This seems related to https://github.com/mathics/Mathics/issues/844

mmatera:
I am trying to do a rebase of this PR, but there are something that I do  not completely understand: in Expression, in the PR, leaves are represented by tuples. This originates some errors in operations that try to modify the leaves of an expression. I do not understand the motivation for using tuples instead of lists. Is about safety in the code? or about performance? @poke1024 could you give me a hand with this?

rocky:
#936 should be the same thing rebased on current sources.

PR #621, opened 2016-10-15
CharacterEncoding and more

poke1024:
This adds the `CharacterEncoding` option to text and csv `Import` and `Export`, as well as to `FromCharacterCode` and `ToCharacterCode`.

Along the way, I encountered quite a number of things I needed to add/change, specifically:

- importers and exporters may now specify `Options` and they are passed down to the respective calls
- options are now parsed very leniently; it doesn't matter whether you write `X -> 1` or `"X" -> 1`; it also doesn't matter if X has been declared a `System` symbol or not (i.e. it doesn't matter if `X` resolves to ```System`X``` or ```Global`X```.
- `FilterRules`
- `AllTrue`, `AnyTrue`, `NoneTrue`

### DIFF
--- a/mathics/builtin/base.py
+++ b/mathics/builtin/base.py
@@ -8,6 +8,7 @@ import re
 import sympy
 from functools import total_ordering
 import importlib
+from itertools import chain
 
 from mathics.core.definitions import Definition
 from mathics.core.rules import Rule, BuiltinRule, Pattern
@@ -192,12 +193,16 @@ class Builtin(object):
 
     @staticmethod
     def get_option(options, name, evaluation, pop=False):
-        # we do not care, whether an option X is given as
-        # System`X, Global`X, or "X". we always handle it
-        # as the same option. this matches Wolfram Language
+        # we do not care whether an option X is given as System`X,
+        # Global`X, or with any prefix from $ContextPath for that
+        # matter. Also, the quoted string form "X" is ok. all these
+        # variants name the same option. this matches Wolfram Language
         # behaviour.
 
-        for variant in ('System`%s', 'Global`%s', '"%s"'):
+        contexts = (s + '%s' for s in
+                    evaluation.definitions.get_context_path())
+
+        for variant in chain(contexts, ('"%s"',)):
             resolved_name = variant % name
 
             if pop:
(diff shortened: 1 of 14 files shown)


mathics/builtin/base.py
```diff
@@ -190,13 +190,25 @@ def get_functions(self, prefix='apply'):
...
-        else:
-            return None
+    @staticmethod
+    def get_option(options, name, evaluation, pop=False):
+        # we do not care, whether an option X is given as
+        # System`X, Global`X, or "X". we always handle it
+        # as the same option. this matches Wolfram Language
+        # behaviour.
+
+        for variant in ('System`%s', 'Global`%s', '"%s"'):
```
sn6uv: Should we look at all the entries in `$ContextPath` or even just drop the context entirely?

mathics/builtin/importexport.py
```diff
@@ -458,6 +485,8 @@ class Import(Builtin):
      = ...
     #> Import["ExampleData/Middlemarch.txt"];
      : An invalid unicode sequence was encountered and ignored.
+    #> StringTake[Import["ExampleData/Middlemarch.txt", CharacterEncoding -> "ISO8859-1"], {21, 69}]
+     = Le sentiment de la fausseté des plaisirs présents
```
sn6uv: Great!

mathics/builtin/logic.py
```diff
@@ -137,3 +138,120 @@ class False_(Predefined):
...
+    <dt>'AllTrue[$test$]'
+        <dd>gives an operator that may be applied to expressions.
+    </dl>
+
+    >> AllTrue[{2, 4, 6}, EvenQ]
+     = True
+
+    >> AllTrue[{2, 4, 7}, EvenQ]
+     = False
+    """
```
sn6uv: can you add some 0-arg tests to these functions.

mathics/builtin/importexport.py
```diff
@@ -683,6 +724,20 @@ class Export(Builtin):
      | 1 + x + y
     #> DeleteFile[%%]
 
+    #> Export["abc.txt", "ä", CharacterEncoding -> "Latin-1"];
```
sn6uv: should this be `ISOLatin1` to match what's in `$CharacterEncodings`.

PR #622, opened 2016-10-15
Changes about logic

hengruoz:
logic.py:
1. Added three functions: Implies, Equivalent, Xor
2. Added rules for 'Not': In Mathematica, when you input '!!A' it will return 'A', but Mathics will return '!!A'. So I added a rule to simplify double negation.

characters.py:
1. change the unicode of symbol 'Implies' and 'Equivalent'. The original unicodes are wrong.


mathics/builtin/logic.py
```diff
@@ -116,9 +116,139 @@ class Not(PrefixOperator):
...
+     >> Equivalent[a, b, c]
+      = a \u21D4 b \u21D4 c
+      Otherwise, 'Equivalent' returns a result in DNF
+      >> Equivalent[a, b, True, c]
+       = a && b && c
+     """
+
+    operator = '\u21D4'
+    precedence = 205
+    attributes = ('Flat', 'HoldAll', 'OneIdentity')
```
sn6uv: likewise here, I think the only attribute should be `Orderless`.

mathics/builtin/logic.py
```diff
@@ -116,9 +116,139 @@ class Not(PrefixOperator):
...
+      = False
+
+     If all expressions do not evaluate to 'True' or 'False', 'Equivalent'
+     returns a result in symbolic form:
+     >> Equivalent[a, b, c]
+      = a \u21D4 b \u21D4 c
+      Otherwise, 'Equivalent' returns a result in DNF
+      >> Equivalent[a, b, True, c]
+       = a && b && c
+     """
```
sn6uv: can you also add a test for `Equivalent[]` (should be `True`) and `Equivalent[x]`, also `True`. Note that you can add hidden tests with `#>` (they won't show up in the docs).

mathics/builtin/logic.py
```diff
@@ -116,9 +116,139 @@ class Not(PrefixOperator):
...
+
+    If an expression does not evaluate to 'True' or 'False', 'Xor'
+    returns a result in symbolic form:
+    >> Xor[a, False, b]
+     = a \u22BB b
+    """
+
+    operator = '\u22BB'
+    precedence = 215
+    attributes = ('Flat', 'HoldAll', 'OneIdentity')
```
sn6uv: also `Orderless`.

mathics/builtin/logic.py
```diff
@@ -116,9 +116,139 @@ class Not(PrefixOperator):
...
+                return Expression('Xor', *leaves)
+        elif leaves and not flag:
+            if len(leaves) == 1:
+                return Expression('Not', leaves[0])
+            else:
+                return Expression('Not', Expression('Xor', *leaves))
+        else:
+            return Symbol(repr(not flag))
+
+
```
sn6uv: Looks correct but I'd like to see some more tests. At least: `Xor[a]`, `Xor[]`, `Xor[False]`, `Xor[True]`, `Xor[a, b]`.

mathics/core/characters.py
```diff
@@ -339,7 +339,7 @@
     'Equal': '\uF431',
     'EqualTilde': '\u2242',
     'Equilibrium': '\u21CC',
-    'Equivalent': '\u29E6',
+    'Equivalent': '\u21D4',
```
hengruoz: It's true that in mathematica "Equivalent" is "\u29E6" and "Implies" is "\uF523", but in Mathics they look like wrong characters. I searched in the internet and found "\u29E6" was "⧦" and "\uF523" was in the private use area. I think Mathematica uses its own font to show math symbols. That's an awkward problem... I will change characters.py back.

mathics/builtin/logic.py
```diff
@@ -116,9 +116,139 @@ class Not(PrefixOperator):
...
+    If an expression does not evaluate to 'True' or 'False', 'Implies'
+    returns a result in symbolic form:
+    >> Implies[a, Implies[b, Implies[True, c]]]
+     = a \u21D2 b \u21D2 c
+    """
+
+    operator = '\u21D2'
+    precedence = 200
+    attributes = ('HoldAll', 'OneIdentity')
```
hengruoz: Thanks a lot. I didn't know about Attributes function... I will correct it and the following problems.

mathics/core/characters.py
```diff
@@ -510,7 +510,8 @@
     'ImaginaryI': '\uF74E',
     'ImaginaryJ': '\uF74F',
     'ImplicitPlus': '\uF39E',
-    'Implies': '\uF523',
+# Implies' unicode in Mathematica is '\uF523', but it will show nothing without the specific font.
+    'Implies': '\u21D2',
```
sn6uv: The problem with the change here is that it breaks `a \[Implies] b` (we should add a test).

mathics/builtin/logic.py
```diff
@@ -116,9 +116,158 @@ class Not(PrefixOperator):
...
+    >> Implies[True, a]
+     = a
+
+    If an expression does not evaluate to 'True' or 'False', 'Implies'
+    returns a result in symbolic form:
+    >> Implies[a, Implies[b, Implies[True, c]]]
+     = a \u21D2 b \u21D2 c
+    """
+
+    operator = '\u21D2'
```
sn6uv: If we use this character then `Implies[a, b]` is formatted as `a \u21D2 b` which should actually parse as `DoubleRightArrow[a, b]`.

sn6uv:
Great!

PR #623, opened 2016-10-15
In-memory sessions

poke1024:
I'm wondering if we really want to store definitions data in the database for the classic mathics server. If there's a lot of data, the interface gets really sluggish since it needs to unpickle and repickle all definitions for each query.

Here's a proposal to keep everything in memory, which is always fast (but loses its data, as soon as you quit the mathicsserver process).

### DIFF
--- a/mathics/core/evaluation.py
+++ b/mathics/core/evaluation.py
@@ -294,22 +294,10 @@ class Evaluation(object):
         return result
 
     def get_stored_result(self, result):
-        from mathics.core.expression import Symbol
-
         # Remove outer format
         if result.has_form(FORMATS, 1):
             result = result.leaves[0]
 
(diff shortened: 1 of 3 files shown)


Piruzzolo:
Maybe adding an autosave in idle time?

sn6uv:
I guess this is a holdover from when we supported multiple client sessions properly in the frontend (I think it's somewhat broken now). The solution I envisage for multiple clients is to give each client their own kernel (or even multiple if they want different workspaces) and store all the definitions in memory. At idle time (or when we inevitably run out of RAM) the definitions should be saved to disk. When the user returns we load their state from disk.

I'm +1 on merging this.

rocky:
@poke1024 @GarkGarcia Although I did test this, please double check. If this is addressed it would be nice to close this issue.

PR #625, opened 2016-10-17
Faster leading blank patterns

poke1024:
I came across this while using `GatherBy`; without this PR, `x = Range[5000]; First[Timing[GatherBy[x, OddQ]]]` evaluates to `29.0664`, with this PR this changes to `0.517357`.

The basic problem is described in the comment in `yield_choice`: sub patterns that are very expensive to check are tested, even though the Blank match should be obvious.

Seems to give a very slight but measurable improvement in runtime.

[master.txt](https://github.com/mathics/Mathics/files/534476/master.txt)

[fastblank.txt](https://github.com/mathics/Mathics/files/534477/fastblank.txt)

### DIFF
--- a/mathics/builtin/patterns.py
+++ b/mathics/builtin/patterns.py
@@ -756,6 +756,9 @@ class Pattern_(PatternObject):
             self.error('patvar', expr)
         self.pattern = Pattern.create(expr.leaves[1])
 
+    def __repr__(self):
+        return '<Pattern: %s>' % repr(self.pattern)
+
     def get_match_count(self, vars={}):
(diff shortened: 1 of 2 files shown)


sn6uv:
Is this valid for `Flat` attributes?

rocky:
> Is this valid for `Flat` attributes?

I take it this is about improving coverage or improving performance furhter, not about whether this causes a bug. 

Assuming this is an overall improvement, it doesn't bother me that it isn't perfect. We can deal with perfection another time. 

If this was about being buggy, let us know. Or alternatively we'll realize that and deal with it down the line.

PR #631, opened 2016-10-19
Fixes Manipulate[x,{x,{a,b}}]

poke1024:
This got broken with the `imathics` branch merge into master. There is no `format_all_outputs` anymore. Works now.

### DIFF
--- a/mathics/builtin/manipulate.py
+++ b/mathics/builtin/manipulate.py
@@ -168,8 +168,8 @@ class _WidgetInstantiator():
     def _add_options_widget(self, symbol, options, default, label, evaluation):
         formatted_options = []
         for i, option in enumerate(options.leaves):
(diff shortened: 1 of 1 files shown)


sn6uv:
This kind of thing seems to be happening frequently. The motivation for putting imathics into a separate repository was it's experimental status. I think we should consider moving it into the main repository. I'd be more confident if we had some passing tests for imathics though. Thoughts @poke1024?

poke1024:
Having imathics in the main repo would be a big advantage in my opinion. Jupyter has automated tests so we should be able to do automated tests for imathics, though it's not that simple either. If we want one quality tested Mathics version that also has Jupyter support, one repo is the way to go. I'll take a look at the test thing.

poke1024:
see https://github.com/mathics/Mathics/pull/636

PR #632, opened 2016-10-19
Fixes Travis

poke1024:
Travis CI currently seems to break due to a change in spacy's `en` corpus. This PR seems to fix it by installing a specific version of the corpus using `sputnik`.

Another (new) problem is that `DominantColors` fails. I think this might be due to some changed `PIL` version. Therefore, this PR adds print outs of versions of all installed modules to enable looking into this further now (as well as in the future).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,15 +49,15 @@ before_install:
   - if [[ "$ALL_MODULES" == "true" ]]; then
       sudo apt-get update -qq &&
       sudo apt-get install -qq python-enchant &&
-      pip install numpy wordcloud ipywidgets ipykernel IPython nltk langid pycountry pyenchant lxml matplotlib networkx &&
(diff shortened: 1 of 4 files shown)


poke1024:
Two things seemed to have happened:

(1) spacy got updated. this (a) broke the Travis download, which I fixed via introducing a fixed version download via `sputnik` in `travis.yml`. it also broke `DeleteStopwords` as the API for `is_stop` has changed. This is now changed and works again.

(2) `DominantColors` was suddenly broken. I can only hypothesize to the reason; the difference between my local test cases and the ones on Travis was one pixel that was counted or not counted as a certain color. Since the base image was jpg, I can only assume that Travis maybe updates their `lipjpeg` that then handled a border case in reading `sunflower.jpg` differently and ended up with one pixel slightly different than before. This is the only explanation I have. I now changed the complete test case to `lena.tif`, and since `tif` is lossless, there should really be no more changes in the future due to library updates.

sn6uv:
FYI I've restarted the tests on #631 and #630.

sn6uv:
Looks like spacy just released a 1.0 https://explosion.ai/blog/spacy-deep-learning-keras

PR #644, opened 2016-11-02
Support for importing HTML

poke1024:
Only `FullData` is not ready for a test case yet, but I consider that a border case.


mmatera:
Rebased in #1368

PR #645, opened 2016-11-03
Give warnings when using options not supported by a Builtin

poke1024:
Currently, if options are given for commands that are not supported, they are simply ignored. This can be very confusing for novice users, especially if they mistype an option (e.g. `FieldSeperetors` instead of `FieldSeparators`).

This PR adds a new meta option called `$OptionSyntax` that may be either `Strict`, `Warn` or `Ignore`, and defines what should happen, if the user specifies an option that does not show up in  the `Builtin`'s `options` dictionary.

The default (also used if no `$OptionSyntax` is specified) is to give a warning but continue with evaluation. `Strict` gives a warning and stops the evaluation (this happens for `Plot`). `Ignore` is the old behavior. 

In addition to this, this PR adds custom checks for `Import` and `Export` that gives warnings for options that are not supported by the import/export plugin.

In this current form, `Import`, `Export` and other commands will warn about unused options, which might differ from MMA. Still, I think these kinds of warnings are so useful that we should accept their general existence in Mathics. In addition, some commands, like `Plot`, really want to abort evaluation, if non-supported options are given, and this PR offers a solution for not having to code this checks for each single Builtin that wants to behave in that way.

### DIFF
--- a/mathics/builtin/base.py
+++ b/mathics/builtin/base.py
@@ -18,6 +18,34 @@ from mathics.core.expression import (BaseExpression, Expression, Symbol,
 import six
 
 
+def get_option(options, name, evaluation, pop=False, evaluate=True):
+    # we do not care whether an option X is given as System`X,
+    # Global`X, or with any prefix from $ContextPath for that
+    # matter. Also, the quoted string form "X" is ok. all these
+    # variants name the same option. this matches Wolfram Language
+    # behaviour.
+
+    contexts = (s + '%s' for s in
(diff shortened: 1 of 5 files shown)


sn6uv:
This sounds like a great idea. It hits a pain point I've hit before which is trying to use an MMA option that isn't implemented in Mathics and not knowing what's gone wrong.

GarkGarcia:
Hey @poke1024, would you mind rebasing this too? 😅️

PR #646, opened 2016-11-03
Adding default streams for stdin, stdout and stderr

mmatera:


mmatera:
I realized that this was still

PR #647, opened 2016-11-07
Iohooks

mmatera:
In MMA, the main loop includes some hooks which allow us to modify the way in which expressions are readed from the input, and postprocessed after evaluation:

""
$PreRead	a function applied to each input string before being fed to Mathematica
$Pre	a function applied to each input expression before evaluation
$Post	a function applied to each expression after evaluation
$PrePrint	a function applied after Out[n] is assigned, but before the result is printed
$SyntaxHandler	a function applied to any input line that yields a syntax error
""
This PR is (maybe a raw)  proposal to implement this functionality. By now, just $Pre, $Post and $PrePrint hooks are implemented.

### DIFF
--- a/mathics/builtin/__init__.py
+++ b/mathics/builtin/__init__.py
@@ -103,6 +103,30 @@ def get_module_doc(module):
     return title, text
 
 
+inout_hooks = {'$PreRead':"$PreRead is a global variable whose value, if set, is applied to the \
+text or box form of every input expression before it is fed to the parser. ",
+               '$Pre': "$Pre is a global variable whose value, if set, is applied to every \
+input expression.",
+               '$Post': "$Post is a global variable whose value, if set, is applied to every \
+output expression. ",
+               '$PrePrint': "$PrePrint is a global variable whose value, if set, is applied to \
+every expression before it is printed. ",
+               '$SyntaxHandler': "$SyntaxHandler is a global variable which, if set, is applied to any \
+input string that is found to contain a syntax error. ",
+           }
+
+def load_InOut_hooks(definitions):
+    from mathics.core.definitions import Definition
+    from mathics.core.expression import  Symbol, String
+    from mathics.core.rules import Rule
+    from mathics.core.pattern import Pattern_create
+    from mathics.builtin.inout import MessageName
+    for hook in inout_hooks:
+        fullname = "System`" + hook
+        definitions.builtin[fullname] = Definition(name = fullname )
+        usage = Rule(MessageName(Symbol(fullname), String("usage")), String(inout_hooks[hook]))
+        definitions.builtin[fullname].set_values_list('messages',[usage])
+
 def contribute(definitions):
     # let MakeBoxes contribute first
     builtins['System`MakeBoxes'].contribute(definitions)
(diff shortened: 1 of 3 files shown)


mathics/builtin/__init__.py
```diff
@@ -103,13 +103,39 @@ def get_module_doc(module):
...
+    from mathics.core.definitions import Definition
+    from mathics.core.expression import  Symbol, String
+    from mathics.core.rules import Rule
+    from mathics.core.pattern import Pattern_create
+    from mathics.builtin.inout import MessageName
+    for hook in inout_hooks:
+        fullname = "System`" + hook
+        definitions.builtin[fullname] = Definition(name = fullname )
+        usage = Rule(MessageName(Symbol(fullname), String("usage")), String(inout_hooks[hook]))
+        definitions.builtin[fullname].set_values_list('messages',[usage])
```
sn6uv: Why do the hooks need a special loading mechanism? Does the standard `Builtin` method not work?

mathics/core/evaluation.py
```diff
@@ -253,12 +257,23 @@ def evaluate(self, query, timeout=None):
 
         result = None
         exc_result = None
+        
+        def check_io_hook(hook):
+            return  len(self.definitions.get_definition(hook).ownvalues)>0
```
sn6uv: Better to use `get_ownvalues(hook)`.

mathics/core/evaluation.py
```diff
@@ -270,6 +285,9 @@ def evaluate():
                 self.definitions.add_rule('Out', Rule(
                     Expression('Out', line_no), stored_result))
             if result != Symbol('Null'):
+                if check_io_hook('System`$PrePrint'): 
+                    print(self.definitions.get_definition('System`$PrePrint').__repr__())
```
sn6uv: extra print statement here

mathics/core/evaluation.py
```diff
@@ -193,6 +193,10 @@ def display(self, data, metadata):
         raise NotImplementedError
 
 
+
+
+    
+
```
sn6uv: extra whitespace here

mathics/builtin/__init__.py
```diff
@@ -121,3 +147,9 @@ def contribute(definitions):
         if not definitions.have_definition(ensure_context(operator)):
             op = ensure_context(operator)
             definitions.builtin[op] = Definition(name=op)
+
+    
+#    definitions.builtin['System`$PrePrint'].set_values_list('messages',
+#                                                            Rule('usage',"$PrePrint is a global variable whose value, if set, is applied to \
+#every expression before it is printed. ",True))
+
```
sn6uv: extra noise here.

mathics/builtin/__init__.py
mmatera: My problem here is that we can not start a class name with $ in python ... This is also a problem for defining tests as a doc string. Is there a better way to deal with this?

mathics/builtin/__init__.py
mmatera: I left this because it was another possibility to check if the hook was defined. I removed the comment now.

mathics/core/evaluation.py
mmatera: Ok

PR #648, opened 2016-11-07
Adding the "initfile" option to the command line interpreter

mmatera:


sn6uv:
What is the difference between init files and standard files? I'm a bit confused by the MMA docs on this.

mmatera:
An init.m is just a script that runs before the interpreter ask for the first line. In mma, there are always a default init.m file, which is empty. In this PR, the aim is to imitate the functionality in the mma command line interpreter, which is equivalent to call the program with a filename and the option "--persist"

mmatera:
I think this is also ready now...

PR #652, opened 2016-11-16
A mechanism to load python modules defining mathics symbols from inside the interpreter and from  the command line.

mmatera:
As the number of functionalities increases, the number of   python libraries dependencies and the general testing time increases more and more, doing the travis testing fail. 
   
This PR tries to address this problem by providing a dynamical mechanism to load python modules defined in different small packages.   The function 

LoadPyMathicsModule["module"]

loads the python package "module", provided "module" has defined the variable pymathics_version_data". In such a case, it loads the symbols defined in the package. 

It would also be useful to accelerate the development of new functionalities like - for example - importers/exporters which could be defined by independent packages. Other example would be modules like NaturalLanguage or the implementation of  a linear algebra package based on Scipy which would not fit on the main built tree of Mathics.   

The external package loading can also be done from the command line, using the parameter "--pyextensions"


GarkGarcia:
Hey @mmatera, this is a very interesting idea. I was just wondering, is there a Mathematica equivalent for this? In other words, does Mathematica have a module system? We should probably try to emulate it (if it exists).

mmatera:
Wolfram Mathematica has MathLinks, but it does not have support for python, so it would be a distinctive feature of Mathias. On the other hand, I think that Mathlinks is less intuitive than my proposal (because it is an interface to C++, which is a strongly typed language).

GarkGarcia:
> Wolfram Mathematica has MathLinks, but it does not have support for python, so it would be a distinctive feature of Mathias. On the other hand, I think that Mathlinks is less intuitive than my proposal (because it is an interface to C++, which is a strongly typed language).

Makes sense, Python is quite a good language for writing software extensions. However, I think we should call `LoadPyMathicsModule` in a similar fashion to how MathLinks calls the equivalent function.

mmatera:
The problem is that there is no equivalent functionality. There are similar things, but not the same thing:

https://library.wolfram.com/infocenter/TechNotes/174/
https://reference.wolfram.com/language/guide/CallingExternalPrograms.html

GarkGarcia:
> The problem is that there is no equivalent functionality. There are similar things, but not the same thing:
> 
> https://library.wolfram.com/infocenter/TechNotes/174/
> https://reference.wolfram.com/language/guide/CallingExternalPrograms.html

I see. Anyway, if we're going to implement a module system we should think about what's the most ergonomic way of doing so. In my understanding, you proposal is similar to how header files work in C, right? Ie. by importing a module you necessarily imports all definitions contained in the module.

It would be cool if we could import specific definitions from a module, just like we can in Python (`from my_module import stuff`). Also, your proposal would introduce an interface for importing definitions written in Python, but we should also consider implementing an interface for importing definitions written in the Mathics language (which is a strict yet imcomplete Mathematica implementation, as far as I can tell).

mmatera:
> > The problem is that there is no equivalent functionality. There are similar things, but not the same thing:
> > https://library.wolfram.com/infocenter/TechNotes/174/
> > https://reference.wolfram.com/language/guide/CallingExternalPrograms.html
> 
> I see. Anyway, if we're going to implement a module system we should think about what's the most ergonomic way of doing so. In my understanding, you proposal is similar to how header files work in C, right? Ie. by importing a module you necessarily imports all definitions contained in the module.
> 
> It would be cool if we could import specific definitions from a module, just like we can in Python (`from my_module import stuff`). Also, your proposal would introduce an interface for importing definitions written in Python, but we should also consider implementing an interface for importing definitions written in the Mathics language (which is a strict yet imcomplete Mathematica implementation, as far as I can tell).

Indeed, the idea is following the C++ standard, which is the same than the W-MMA standard. What you always can do is to define packages loading a certain subset of symbols (from the python side), and load the reduced package. But it is important to check that you are loading all the dependences properly.
On the other hand, to load native MMA packages, you can use the Keyword "Get", and define packages and contexts using Begin and BeginPackage Keywords, which are currently implemented in mathics.

What makes that loading extensions from python can be simpler that loading generic (compiled) libraries, is that all the information about connecting a python function with a mma expression can be defined directly in the docstring, avoiding the need of the compilers, and the infrastructure of mathlink.

GarkGarcia:
> On the other hand, to load native MMA packages, you can use the Keyword "Get", and define packages and contexts using Begin and BeginPackage Keywords, which are currently implemented in mathics.

Ohh, I see. Sorry, I'm not that experienced with MMA, didn't know about that.

> What makes that loading extensions from python can be simpler that loading generic (compiled) libraries, is that all the information about connecting a python function with a mma expression can be defined directly in the docstring, avoiding the need of the compilers, and the infrastructure of mathlink.

Makes sense. 

Alright, my current understanding is that the only thing preventing us from merging the PR already is the discussion on how should `LoadPyMathicsModule` be called. I think we should avoid mentioning implementation details in it's name (in this case, the words "Py" and "Mathics"). What about `LoadModule` or `LoadExtension`?

Another thing worth considering is creating CLI options to call `LoadPyMathicsModule` on startup. The idea is that running

```
$ mathics --module "module"
```

would start Mathics and run `LoadPyMathicsModule["module"]` before prompting the user for an expression.

mmatera:
If you want, we could change the name of the load function. On the other hand, the idea would be to use this function inside packages written in native mma language, in a way that the final user do not have to deal with it.

GarkGarcia:
> If you want, we could change the name of the load function. On the other hand, the idea would be to use this function inside packages written in native mma language, in a way that the final user do not have to deal with it, and leaving it for developers.

I'd prefer if we chose a more generic name.

mmatera:
OK, I think that now the PR is ready.

mathics/builtin/evaluation.py
```diff
@@ -459,27 +459,35 @@ class Exit(Builtin):
...
-    '''
-    <dl>
-    <dt>'Quit[]'
-      <dd>terminates the Mathics session.
-    <dt>'Quit[n]'
-      <dd>terminates with exit code $n$.
-    </dl>
-    '''
-
+    # I need to recover the standard behaviour to make the tests run properly.
```
GarkGarcia, line 462: You need to undo this.

PR #672, opened 2017-01-24
Adding some checking functions

larionvn:
Add BooleanQ, DigitQ, LetterQ

This is my first pull request. It is great if you guys can spend sometime to feedback on this. Thank in advance

### DIFF
--- a/mathics/builtin/strings.py
+++ b/mathics/builtin/strings.py
@@ -641,6 +641,9 @@ class DigitQ(Builtin):
 
     >> DigitQ["-123456789"]
      = False
+    
+    #> DigitQ[""]
+     = True
 
     #> DigitQ["."]
      = False
@@ -654,7 +657,7 @@ class DigitQ(Builtin):
 
     rules = {
         'DigitQ[string_]': (
-            'If[StringQ[string], StringMatchQ[string, DigitCharacter..], False, False]'),
+            'If[StringQ[string], StringMatchQ[string, DigitCharacter...], False, False]'),
     }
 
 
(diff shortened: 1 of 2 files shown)


mathics/builtin/strings.py
```diff
@@ -623,6 +623,70 @@ class HexidecimalCharacter(Builtin):
...
+    #> DigitQ[1==2]
+     = False
+
+    #> DigitQ[a=1]
+     = False
+    """
+
+    rules = {
+        'DigitQ[string_]': (
+            'If[StringQ[string], StringMatchQ[string, DigitCharacter..], False, False]'),
```
sn6uv: I think it should be `DigitCharacter...` for compatibility.  Could you also add a test for this `DigitQ[""]` (`True`).

mathics/builtin/strings.py
```diff
@@ -623,6 +623,70 @@ class HexidecimalCharacter(Builtin):
...
+    >> LetterQ["Welcome to Mathics"]
+     = False
+
+    #> LetterQ["\[Alpha]\[Beta]\[Gamma]\[Delta]\[Epsilon]\[Zeta]\[Eta]\[Theta]"]
+     = True
+    """
+
+    rules = {
+        'LetterQ[string_]': (
+            'If[StringQ[string], StringMatchQ[string, LetterCharacter..], False, False]'),
```
sn6uv: same as `DigitQ` for an empty string.

mathics/builtin/strings.py
larionvn: Just updated with a new commit to incorporate this
Thank for your feedback

PR #678, opened 2017-03-17
Add some new polynomial functions

hailongnk:


sn6uv:
Failed test is due to one of the optional libraries. Investigating

mathics/builtin/algebra.py
```diff
@@ -178,6 +178,26 @@ def find_vars(e, e_sympy):
     return variables
 
 
+def exponent(expr, var):
```
sn6uv: Could you add a docstring explaining this function? It's not immediately clear what it does.

mathics/builtin/algebra.py
```diff
@@ -178,6 +178,26 @@ def find_vars(e, e_sympy):
...
+    f = expr.to_sympy()
+    x = var.to_sympy()
+    list = [i.as_coeff_exponent(x) for i in f.expand(power_exp=False).as_ordered_terms()]
```
sn6uv: This could be cleaned up a bit, no need to allocate a new list here. Also, `list` isn't a good variable name.

mathics/builtin/algebra.py
```diff
@@ -312,6 +332,148 @@ def apply(self, expr, evaluation):
...
+    <dl>
+    <dt>'FactorTermsList[poly]'
+        <dd>returns a list in which the first element is the overall numerical factor in $poly$,
+        and the second element is the polynomial with the overall factor removed.
+    <dt>'FactorTermsList[poly,{x1, x2, ...}]'
+        <dd>returns a list of factors of $poly$.
+        The first element in the list is the overall numerical factor.
+        The second element is a factor that does not depend on any of the $xi$.
+        Subsequent elements are factors which depend on progressively more of the $xi$.
+    </dl>
```
sn6uv: Can we rewrite these descriptions a little so they're not verbatim identical to MMA: https://reference.wolfram.com/language/ref/FactorTermsList.html

mathics/builtin/algebra.py
```diff
@@ -312,6 +332,148 @@ def apply(self, expr, evaluation):
...
+     = {7 - I, 1}
+    #> FactorTermsList[(x - 1) (1 + a), {c, d}]
+     = {1, -1 - a + x + a x}
+    #> FactorTermsList[(x - 1) (1 + a), {c, x}]
+     = {1, 1 + a, -1 + x, 1}
+    #> FactorTermsList[(x - 1) (1 + a), {}] == FactorTermsList[(x - 1) (1 + a)]
+     = True
+    
+    #> FactorTermsList[x]
+     = {1, x}
```
sn6uv: What about `FactorTermList["a"]`? Crashes.

mathics/builtin/algebra.py
```diff
@@ -312,6 +332,148 @@ def apply(self, expr, evaluation):
...
+        if expr == Integer(0):
+            return Expression('List', Integer(1), Integer(0))
+        elif isinstance(expr, Number):
+            return Expression('List', expr, Integer(1))
+        
+        for x in vars.leaves:
+            if not(isinstance(x, Atom)):
+                return evaluation.message('CoefficientList', 'ivar', x)
+        
+        sympy_expr = sympy.together(expr.to_sympy())
```
sn6uv: Note: `to_sympy()` can return None.

mathics/builtin/algebra.py
```diff
@@ -312,6 +332,148 @@ def apply(self, expr, evaluation):
...
+        sympy_vars = [x.to_sympy() for x in vars.leaves if isinstance(x, Symbol) and sympy_expr.is_polynomial(x.to_sympy())]
```
sn6uv: here too: `to_sympy()` returns None

mathics/builtin/arithmetic.py
```diff
@@ -1134,6 +1134,55 @@ class Abs(_MPMathFunction):
...
+    messages = {
+        'argx':  'Sign called with `1` arguments; 1 argument is expected.',
+    }
+    
+    def apply(self, x, evaluation):
+        'Sign[x_]'
+        if isinstance(x, Complex):
+            return Expression('Times', x, Expression('Power', Expression('Abs', x), -1))
+        
+        sympy_x = x.to_sympy()
```
sn6uv: might be None.

mathics/builtin/algebra.py
hailongnk: no crash here because there is a check that x must be Symbol

PR #682, opened 2017-03-29
Adding some functions in numbertheory and arithmetic

larionvn:
- This PR includes the implementation of 3 functions:
  + MantissaExponent
  + FractionalPart
  + CubeRoot

### DIFF
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -909,13 +909,28 @@ class CubeRoot(Builtin):
      
     #> CubeRoot[-0.5]
      = -0.793701
+     
+    #> CubeRoot[3 + 4 I]
+     : The parameter 3 + 4 I should be real valued.
+     = (3 + 4 I) ^ (1 / 3)
     """
     
     attributes = {'Listable', 'NumericFunction', 'ReadProtected'}
     
-    rules = {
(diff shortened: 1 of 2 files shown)


mathics/builtin/arithmetic.py
```diff
@@ -887,6 +887,37 @@ class Sqrt(SympyFunction):
...
+     = b ^ (1 / 3)
+     
+    #> CubeRoot[-0.5]
+     = -0.793701
+    """
+    
+    attributes = {'Listable', 'NumericFunction', 'ReadProtected'}
+    
+    rules = {
+        'CubeRoot[n_]': 'If[NumericQ[n] == True, If[n > 0, Power[n, Divide[1, 3]], Times[-1, Power[Times[-1, n], Divide[1, 3]]], Power[n, Divide[1, 3]]], Power[n, Divide[1, 3]]]',
```
sn6uv: What about Complex numbers (which are `NumericQ`). It looks like `CubeRoot` only acts on reals.  I think this can be cleaned up a bit.

larionvn: Thanks for your feedback @sn6uv . The tests were failed due to some problems on the Travis_ci side. I've passed all the testcases (on Travis_ci) with the same source code before.  For the Complex numbers case, I've checked on the Mathematica and  see that the CubeRoot still work with Complex number except they throw a warning message. Anyway, I will push the new source code for handling that warning message soon.

wolfv:
rebased and will be merged once tests pass.

PR #693, opened 2017-05-18
Add RealDigits

larionvn:
In this pull request, the RealDigits[] function has been added to the Mathics.
In order to pass all the TravisCI checks, I've also updated the .travis.yml file by adding the package Requests and specifying the version of Ipython to 5.0.0.

PR #695, opened 2017-05-19
Improved worksheet management

balthild:
Demo: http://math.qaq.cat/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ before_install:
   - if [[ "$ALL_MODULES" == "true" ]]; then
       sudo apt-get update -qq &&
       sudo apt-get install -qq python-enchant &&
-      pip install numpy ipywidgets ipykernel IPython nltk langid pycountry pyenchant lxml matplotlib networkx &&
+      pip install numpy ipywidgets ipykernel requests IPython==5.0.0 nltk langid pycountry pyenchant lxml matplotlib networkx &&
       python travis.py &&
       if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then
           pip install pattern;
--- a/mathics/web/forms.py
+++ b/mathics/web/forms.py
@@ -43,6 +43,15 @@ class SaveForm(AjaxForm):
     )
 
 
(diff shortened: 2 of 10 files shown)


balthild:
Could you please tell me why the build failed?
Jobs with `ALL_MODULES="true"` fail at `python: No module named requests`. Seems that `request` should be installed before build (#693), but some other PRs built successfully without this.

larionvn:
@balthild  I'm not sure about the changing details but the TravisCI is updating its environments so it now requires more modules to run. So try to install "requests" module before build to fix your issue.

sn6uv:
The only thing I'm not sure about is, design wise, is it best to present the users with a list of worksheets or preload one (as we do today)?

balthild:
I agree. Displaying the worksheet last used is much better.

BTW, the sharing feature is broken. I'm fixing it.

suhr:
I tried to rebase you PR against the current master: https://github.com/mathics/Mathics/pull/782

It runs, but /ajax/getworksheets/ throws `sqlite3.OperationalError: no such table: web_worksheet`.

Would you like to update your PR to current master branch (using django 1.11)?

balthild:
The `/ajax/getworksheets/` route hits this function: https://github.com/mathics/Mathics/pull/695/commits/3609988e16c4709297bd75b676823bb767e3f050#diff-c53f701d935994cfef35be4b90836f1cR346
Seems that it's not modified by this PR. It might be a bug that appears in the current master branch too. We'd better fix it individually.

wolfv:
@balthild do you have some time to investigate this? I am hoping to make a new release soon (at least on conda-forge).

balthild:
Sure. I've created a new PR: #783

wolfv:
we still need to take care of this one, correct?

suhr:
Yes. Right now worksheet saving does not work.

It would be nice if @balthild updates his PR and fixed remain issues.

mmatera:
@balthild,  @suhr, I do not know if this still makes sense, but in any case, this should be pushed to MathicsDjango. I close this PR here.

PR #710, opened 2017-07-23
Add matrix trace as Tr

Mego:
Implements the basic form of [Tr](http://reference.wolfram.com/language/ref/Tr.html.en) (trace of a square matrix).


sn6uv:
Looks great. Thanks for the PR

PR #722, opened 2017-11-01
graphics.py: Remove duplicate imports

numbermaniac:
On line 12, `sin, cos, pi` are already imported from `math`, but are then imported again on line 19.


numbermaniac:
Is there something I need to do for this? The only reason the CI build failed is the pypy tests seem to have randomly segmentation faulted.

GarkGarcia:
> Is there something I need to do for this? The only reason the CI build failed is the pypy tests seem to have randomly segmentation faulted.

Nope, you're all good. Thanks for contributing!

PR #755, opened 2018-10-07
Fix tests

wolfv:
This needs some additional work for PyPy. Not sure what's going on there yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,19 +46,20 @@ matrix:
 before_install:
   - "export DISPLAY=:99.0"
   - "sh -e /etc/init.d/xvfb start"
+  - pip install numpy
   - if [[ "$ALL_MODULES" == "true" ]]; then
       sudo apt-get update -qq &&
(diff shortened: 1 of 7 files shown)


wolfv:
This PR updates spacy to the latest version and fixes the way we download + install the NLP data in travis for spacy 2.0

There were some other fixes (namely since the model changed, some values changed).

I added a way to exclude a test from running by addign `X>` instead of `>>` so we can still display the NLP commands in the docs and have a set of hidden tests.

PyPy has some weird bugs: one where it segfaults because of some internal exception, one where it raises an error because of lxml https://bitbucket.org/pypy/pypy/issues/2666/lxml-40-fails-to-create-xmlsyntaxerror-on

And there was a problem in the ColorDistance function when numpy wasn't installed (noticed when debugging pypy).

wolfv:
@poeschko this fixes the tests.

poeschko:
Great, thanks!

PR #759, opened 2018-10-10
Iohooks

mmatera:
In MMA, the main loop includes some hooks which allow us to modify the way in which expressions are readed from the input, and postprocessed after evaluation:

""
$PreRead a function applied to each input string before being fed to Mathematica
$Pre a function applied to each input expression before evaluation
$Post a function applied to each expression after evaluation
$PrePrint a function applied after Out[n] is assigned, but before the result is printed
$SyntaxHandler a function applied to any input line that yields a syntax error
""
This PR is (maybe a raw) proposal to implement this functionality. By now, just $Pre, $Post and $PrePrint hooks are implemented.


wolfv:
Hey @mmatera can you squash the commits? 

If you do a `git rebase --interactive origin/master` the history will read nicer.

mmatera:
I just did it. I hope I did it properly...

PR #760, opened 2018-10-10
Base64

mmatera:
Adding support to Export to and Import from Base64 binary to text encoding (see https://reference.wolfram.com/language/ref/format/Base64.html).


wolfv:
Nice! looks much better when it's just one commit! 

I would love to merge it, but we need @poeschko for this! ...

mmatera:
Yep... It is just that I didn't know how to do the rebase before... :)

PR #776, opened 2019-06-03
Add some types

suhr:
The situation with static typing in Python is very far from perfect, but I still hope that some types will make Mathics more maintainable.

### DIFF
--- a/mathics/core/definitions.py
+++ b/mathics/core/definitions.py
@@ -10,6 +10,8 @@ import bisect
 
 from collections import defaultdict
 
+import typing
+
 from mathics.core.expression import Expression, Symbol, String, fully_qualified_symbol_name, strip_context
 from mathics.core.characters import letters, letterlikes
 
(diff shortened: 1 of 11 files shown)


mathics/core/definitions.py
```diff
@@ -19,14 +21,14 @@
 full_names_pattern = r'(`?{0}(`{0})*)'.format(base_names_pattern)
 
 
-def get_file_time(file):
+def get_file_time(file) -> float:
     try:
         return os.stat(file).st_mtime
     except OSError:
         return 0
```
wolfv: this is an integer

suhr: But `st_mtime` is a float:

```
>>> os.stat('.zshrc').st_mtime
1548286918.3710866
```

wolfv: and mypy doesn't complain about returning the wrong number type?

suhr: Yep. I suppose integer literals can be both `int` and `float`.

wolfv: ok :)

GarkGarcia:
Hey @suhr, thanks for rebasing the PR! I'll have a look at the CI logs, but I imagine the failed tests have nothing to do with this PR.

PR #780, opened 2019-06-05
Bump sympy version in .travis.yml

suhr:


suhr:
Looks like I have to fix some things...

wolfv:
Could also be that sympy has introduced some different results...

suhr:
> {{f -> (Function[{x}, C[1] + Integrate[ArcSin[f[2 x]], x]])}, {f -> (Function[{x}, C[1] + Integrate[Pi - ArcSin[f[2 x]], x]])}}

Well, looks like dsolve in sympy is broken.

> Exception Unknown SymPy expression: Matrix

But *that* is indeed a bug.

suhr:
So matrix power is now MatPow.

I wonder what should I do with this tests for DSolve:

```
    #> Attributes[f] = {HoldAll};
    #> DSolve[f[x + x] == Sin[f'[x]], f, x]
     : To avoid possible ambiguity, the arguments of the dependent variable in f[x + x] == Sin[f'[x]] should literally match the independent variables.
     = DSolve[f[x + x] == Sin[f'[x]], f, x]

    #> Attributes[f] = {};
    #> DSolve[f[x + x] == Sin[f'[x]], f, x]
     : To avoid possible ambiguity, the arguments of the dependent variable in f[2 x] == Sin[f'[x]] should literally match the independent variables.
     = DSolve[f[2 x] == Sin[f'[x]], f, x]
```

suhr:
```
Test failed: MatrixPower in Reference of built-in symbols / Linear algebra
MatrixPower[{{0, x}, {0, 0}}, n]
Output:
Wanted:
::: Matrix det == 0; not invertible
```

Why.

suhr:
I'm not sure if this is the right way to fix things, but it works.

mathics/builtin/linalg.py
```diff
@@ -761,7 +757,6 @@ class MatrixPower(Builtin):
      = {{169, -70}, {-70, 29}}
 
     #> MatrixPower[{{0, x}, {0, 0}}, n]
```
wolfv: Is the result givne by sympy correct?
Mathematica seems to give something like 

image

suhr: If n is negative, then there's no valid result since the matrix is not invertible. So sympy returns the expression as is.

PR #783, opened 2019-06-11
Try fixing the 'no such table' error

balthild:
The new version of Django is not automatically migrating models in `web/models.py`. Seems that it requires creating migration scripts manually by `python mathics/manage.py makemigrations web` now.


suhr:
I suppose you should add it to `setup.py` as well.

balthild:
According to the statements in the [Django Documentation](https://docs.djangoproject.com/en/2.2/topics/migrations/), database migrations are made once by developers whenever models definition changes, and treated as a part of the codebase. Therefore, we should not run the `makemigrations` command automatically in `setup.py`.

suhr:
Setup.py does not install `mathics/web/migrations/`, so mathicsserver returns the same error as before: `sqlite3.OperationalError: no such table: web_worksheet`

balthild:
I understood... The migrations is created once, but should be run every time we start `mathicsserver`. Currently it's only run at the first run (when there're no `~/.local/var/mathics/mathics.sqlite` file).

suhr:
Well, there's still `sqlite3.OperationalError: no such table: web_worksheet`, even after a successful migration.

suhr:
This commit fixes the issue: https://github.com/suhr/Mathics/commit/7e0dbbf49e8c754b72fe07771009fd38cae5dc3a

balthild:
That's strange. I cannot reproduce it.

image

balthild:
I cherry picked your commit. Thanks!

suhr:
@wolfv

wolfv:
thanks for the contribution!

PR #784, opened 2019-07-10
Added Delete command

mrtwistor:
Added missing Delete command.

### DIFF
--- a/mathics/builtin/lists.py
+++ b/mathics/builtin/lists.py
@@ -1752,6 +1752,7 @@ class Delete(Builtin):
                 raise PartRangeError
         else:
             return self.del_one(cur, rest[0])
+
     def del_part(self, expr,indices,evaluation):
         if indices.is_atom():
             return self.del_one(expr,indices.get_int_value())
--- a/mathics/builtin/lists.py
+++ b/mathics/builtin/lists.py
@@ -1729,7 +1729,7 @@ class Delete(Builtin):
     def del_rec(self, cur, rest):
         if cur.is_atom():
             raise PartDepthError
-        if len(rest)>1:
+        if len(rest) > 1:
             pos = rest[0]
             try:
                 if pos > 0:
--- a/mathics/builtin/lists.py
+++ b/mathics/builtin/lists.py
@@ -1726,7 +1726,8 @@ class Delete(Builtin):
             return cur
         else:
             raise PartRangeError
-    def del_rec(self, cur, rest):
+
+def del_rec(self, cur, rest):
         if cur.is_atom():
             raise PartDepthError
         if len(rest) > 1:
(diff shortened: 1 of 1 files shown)


mathics/builtin/lists.py
```diff
@@ -1683,6 +1683,94 @@ def callback(level):
 
         return Expression('List', *results)
 
+def del_part(expr,indices,evaluation):
+    def del_one(cur,pos):
```
suhr: Is it really necessary to put a function inside a function?

mrtwistor: I wouldn't say it's "necessary", no.  But these are just helper functions that are not intended to be called outside the function del_part, so they are local to that scope.

mathics/builtin/lists.py
```diff
@@ -1684,6 +1684,96 @@ def callback(level):
...
+            return cur
+        elif pos == 0:
+            cur.head = Symbol('System`Sequence')
+            return cur
+        elif pos >= -l:
+            cur.leaves = cur.leaves[:l+pos] + cur.leaves[l+pos+1:]
+            return cur
+        else:
+            raise PartRangeError
+    def del_rec(self, cur, rest):
```
suhr: Add a newline.

mathics/builtin/lists.py
```diff
@@ -1684,6 +1684,96 @@ def callback(level):
...
+        if cur.is_atom():
+            raise PartDepthError
+        if len(rest)>1:
```
suhr: Spaces around `>`.

mathics/builtin/lists.py
suhr: > Is that better?

I believe so.

mathics/builtin/lists.py
mrtwistor: ok

mathics/builtin/lists.py
mrtwistor: oops, I missed this.

mathics/builtin/lists.py
```diff
@@ -1684,6 +1684,97 @@ def callback(level):
...
+                    part = self.del_rec(part,rest[1:])
+                    cur.leaves = cur.leaves[:l+pos] + [part] + cur.leaves[l+pos+1:]
+                    return cur
+                else:
+                    raise PartRangeError
+            except IndexError:
+                raise PartRangeError
+        else:
+            return self.del_one(cur, rest[0])
+    def del_part(self, expr,indices,evaluation):
```
suhr: And here should be a newline too.

suhr:
@wolfv

wolfv:
Great! Fantastic contribution.

Issue #795, opened 2019-11-26
Can't compute eigenvalues/eigenvectors of matrix with complex elements


nan-af:
You can follow this link to reproduce the problem:

https://mathics.angusgriffith.com/#cXVlcmllcz1KeSUzRCgoJTVDJTVCSW90YSU1RCkoJTVDJTVCSEJhciU1RCklMkZzcXJ0JTVCMiU1RCkoJTdCJTdCMCUyQy0xJTJDMCU3RCUyQyU3QjElMkMwJTJDLTElN0QlMkMlN0IwJTJDMSUyQzAlN0QlN0QpJnF1ZXJpZXM9TWF0cml4Rm9ybSU1Qkp5JTVEJnF1ZXJpZXM9RWlnZW52YWx1ZXMlNUJKeSU1RCZxdWVyaWVzPUVpZ2VudmVjdG9ycyU1Qkp5JTVEJnF1ZXJpZXM9RWlnZW5zeXN0ZW0lNUJKeSU1RA==

GarkGarcia:
This is very curious. I'm trying to understand why is `object.__nonzero__` called in [`linalg.py`](https://github.com/mathics/Mathics/blob/master/mathics/builtin/linalg.py#L999):

```
    System::exception:

    Exception raised: cannot determine truth value of Relational
     
    Traceback (most recent call last):
     
    File "/home/pablo/.local/lib/python3.6/site-packages/mathics/web/views.py", line 120, in query
    result = evaluation.evaluate(expr, timeout=settings.TIMEOUT)
     
    File "/home/pablo/.local/lib/python3.6/site-packages/mathics/core/evaluation.py", line 243, in evaluate
    result = run_with_timeout(evaluate, timeout)
     
    File "/home/pablo/.local/lib/python3.6/site-packages/mathics/core/evaluation.py", line 63, in run_with_timeout
    return request()
     
    File "/home/pablo/.local/lib/python3.6/site-packages/mathics/core/evaluation.py", line 226, in evaluate
    result = query.evaluate(self)
     
    File "/home/pablo/.local/lib/python3.6/site-packages/mathics/core/expression.py", line 941, in evaluate
    return result.evaluate(evaluation)
     
    File "/home/pablo/.local/lib/python3.6/site-packages/mathics/core/expression.py", line 871, in evaluate
    eval_range(range(len(leaves)))
     
[...]

PR #804, opened 2020-08-24
Fixed the behaviour of Eigenvalues and Eigenvectors

GarkGarcia:
This is a follow up to https://github.com/mathics/Mathics/issues/795. Although Mathics had no problem computing the eigenvalues and eigenvectors of complex matrices, it tried sorting them before returning the results, which caused errors. This PR makes it so that the results are only sorted whenever it's possible to do so.


wolfv:
Thanks!

PR #808, opened 2020-08-26
Information functions

mmatera:


mathics/doc/doc.py
```diff
@@ -850,6 +850,21 @@ def __init__(self, doc):
     def __str__(self):
         return '\n'.join(str(item) for item in self.items)
 
+    def text(self, detail_level):
+        # used for introspection
```
GarkGarcia, line 854: We should work in this before we merge the PR

GarkGarcia: Besides this, I think we're all done in here. II haven't looked at the tests though.

GarkGarcia:
Hey @mmatera, isn't this a duplicate of #618?

mmatera:
It was an attempt to make a rebase of the other version. At this point, I thought it was a bit messy to make a rebase of the other version. The problem is that I will not be able to work on the project during the next few weeks. After that, I will try to fix the incompatibilities and merge both versions in a single pull request.

GarkGarcia:
> It was an attempt to make a rebase of the other version. At this point, I thought it was a bit messy to make a rebase of the other version. 

I see. Could I close the old PR then?

> The problem is that I will not be able to work on the project during the next few weeks. After that, I will try to fix the incompatibilities and merge both versions in a single pull request.

No problem, take your time 😋️

PR #809, opened 2020-08-27
Fixed the annoying 'Eigenvalues' bug that is making builds fail

GarkGarcia:


GarkGarcia:
The PR passed all relevant tests. The only tests which are not passing are related to image-processing in certain platforms.

PR #815, opened 2020-08-30
Made it so that N computes approximation of polynomial roots

GarkGarcia:
Also fixed a bunch of errors in `Root`'s `to_sympy` implementation


GarkGarcia:
Essentially, now `N[Root[f, i], d]` works properly.

Issue #796, opened 2019-12-04
#name (slot associativity) not implemented yet


ghost:
See: https://reference.wolfram.com/language/ref/Slot.html

For example: `f[#u, #v, #u] &[<|"u" -> x, "v" -> y|>]` should output `f[x, y, z]`, but outputs the error:

> Syntax::sntxf:
> 
> "f[#u, #v, #u] &[" cannot be followed by "<|"u" -> x, "v" -> y|>]" (line 1 of "<notebook>").

in Mathics instead. Seems like `#name` for associations is not implemented yet. Possible related pull request: https://github.com/mathics/Mathics/pull/705

GarkGarcia:
Alright, currently Mathics doesn't take name slots into account. This is something that we should fix.

This fix will involve redesigning the way associations are handled.

GarkGarcia:
Fixed https://github.com/mathics/Mathics/commit/34a1d2a8bb92ef5232a7a0fbe76523ae25594a93

PR #818, opened 2020-09-03
Fixed the bahaviour of `Expression.replace_slots`

GarkGarcia:
Now `Expression.replace_slots` takes associations into account


GarkGarcia:
This is a follow up to https://github.com/mathics/Mathics/issues/796#issuecomment-686613576

PR #819, opened 2020-09-04
Fixed the behaviour of `Quit`

GarkGarcia:
There was a bug that caused the exit code in `Quit[n]` to be ignored. It's now fixed, but the `Quit` shortcut for `Quit[]` doesn't work anymore.

### DIFF
--- a/mathics/builtin/evaluation.py
+++ b/mathics/builtin/evaluation.py
@@ -451,7 +451,7 @@ class Out(Builtin):
 class Exit(Builtin):
     '''
     <dl>
-    <dt>'Exit'
+    <dt>'Exit[]'
       <dd>terminates the Mathics session.
     <dt>'Exit[n]'
       <dd>terminates with exit code $n$.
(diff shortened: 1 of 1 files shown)


GarkGarcia:
@wolfv @suhr I'd like to get a position on the `Quit` shortcut situation from you. Essentially, in Mathematica `Quit` is a shortcut for `Quit[]` (which is a shortcut for `Quit[0]`) when `Quit` is the entirety of the expression being evaluated. I'd like to emulate this behaviour, but I don't think it's possible in our current implementation.

The issue is that Mathics evaluates a function before applying arguments to it, so if `Quit` exits the application then `Quit[n]` will quit Mathics before it has a change to process the `n` argument and correctly exit with `n` as the exit code.

suhr:
Personally, I never used `Quit`, since it's easier to do Ctrl+D. But I think the shortcut is less important than the correct behavior of `Quit[n]`.

PR #822, opened 2020-09-06
Remove deprecation warnings, & lint things

rocky:
* Remove deprecation warnings involving \
* remove some unused imports in the above files
* remove trailing blanks in the above files
* mathics/builtin/pympler/asizeof.py updated to remove internal-unicode
  deprecation warning

### DIFF
--- a/mathics/builtin/diffeqns.py
+++ b/mathics/builtin/diffeqns.py
@@ -8,7 +8,7 @@ Differential equation solver functions
 import sympy
 from mathics.builtin.base import Builtin
 from mathics.core.expression import Expression
-from mathics.core.convert import sympy_symbol_prefix, from_sympy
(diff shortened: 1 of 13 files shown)


rocky:
Given the failures, I will be breaking this up into smaller pieces.

GarkGarcia:
@rocky Not sure why the builds are failing, could you take a look at them?

rocky:
@GarkGarcia Some of them are due to trailing blanks that Emacs removed that seem *necessary* in for the testing system. 

IMHO such either the need for trailing blanks would be removed, or they would be marked in a way that makes it clear where and when they are needed. (And it would be nice if GNU Emacs had an easy way to allow some lines to allow a trailing blank.

PR #829, opened 2020-09-06
Show chapter/sections in test.py

rocky:
Also show number within section.

Before:

```
   1. TEST 1 + 2
   2. TEST 1 - 2 * (3 + 5) / 4
   3. TEST 1 - 2 (3 + 5) / 4
...
```

Now:

```
**********  Language tutorials / Basic calculations **********
   1 ( 1): TEST 1 + 2
   2 ( 2): TEST 1 - 2 * (3 + 5) / 4
...
**********  Language tutorials / Symbols and assignments **********
  25 ( 1): TEST x
  26 ( 2): TEST x + 2 x
```

### DIFF
--- a/mathics/test.py
+++ b/mathics/test.py
@@ -24,7 +24,7 @@ class TestOutput(Output):
         return None
 
 
-sep = '-' * 70 + '\n'
+sep = "-" * 70 + "\n"
 
 
 def compare(result, wanted):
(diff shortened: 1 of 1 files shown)


rocky (commented):
The motivation here is that running tests takes a long time 5 minutes or so just for CI. 

You can filter by either section or skip number, but filtering by skip number is slow. So filtering by section seems to be a big win. However in order to do that, you need to know *what* the section number is!

Independent of this, I find showing the organization of the tests helpful. 

Things are still not ideal because section names do not have to be unique. In particular "Graphics' appears in more than one chapter. Therefore the numbering within the Graphics section will be off in absolute sense for the subsequent "Graphics" sections, and this is relevant if you give the `--skip` option. 

I tried addressing this but the logic is a bit involved and couldn't get this easily working. 

Since this isn't a user-facing issue, but a developer-facing *enhancement* and since just this is in my opinion much much better than was there before, I'd like to do this step, and come back to the other down the line if it becomes an annoyance. In *most* cases the section name is unique.

Also, I separated the "blacken" commit from the function-changes so it would be easier to understand what has changed.

GarkGarcia:
Awesome! Great work!

PR #830, opened 2020-09-06
Remove escape char deprecation warnings

rocky:
Remove deprecation warning associated with `\`.
    
In older Python's backslash (\) followed by a symbol that is not an by one of the acceptable escape characters was okay and would be the two symbols. For example `"\["` would represent the string `"\["`.

In newer Pythons the escape is expected to be escaped. e.g. `"\\[` represents `\[`'.


rocky:
A change to mathics/test.py slipped in and I didn't intend that. I haven't been able to undo this, so I'll do another .

PR #833, opened 2020-09-07
Remove assertRegexpMatches DeprecationWarning

rocky:
It becomes assertRegex instead.

For the most part, I think this is the last of the deprecation warnings.

Again. running `black` format is a separate commit in case that is not desired.


GarkGarcia:
I'm not sure why the build failed, but it has nothing to do with the PR. Apparently `git clone` failed (?)

rocky:
And it worked on https://travis-ci.org/github/rocky/Mathics/builds

PR #834, opened 2020-09-07
Add $Machine, $MachineName, $SystemID ...

rocky:
and `$SystemWordLength`


rocky:
Looks like I have to remove the test example since this is not fixed. I am working on that now.

PR #835, opened 2020-09-07
Update to Pytho 3.8

GarkGarcia:
Added Python 3.8 to the CI config.

### DIFF
--- a/mathics/builtin/numpy_utils/with_numpy.py
+++ b/mathics/builtin/numpy_utils/with_numpy.py
@@ -330,7 +330,10 @@ class _ConditionalTransformer(ast.NodeTransformer):
 
         tree = self.visit(tree)
         tree = ast.fix_missing_locations(tree)
-        tree.posonlyargs = []
+        
+        for node in tree.body:
+            node.args.posonlyargs = []
+
         code = compile(tree, '<conditional:%s>'% func_name, 'exec')
 
         data = {}
--- a/mathics/builtin/numpy_utils/with_numpy.py
+++ b/mathics/builtin/numpy_utils/with_numpy.py
@@ -330,6 +330,7 @@ class _ConditionalTransformer(ast.NodeTransformer):
 
         tree = self.visit(tree)
         tree = ast.fix_missing_locations(tree)
+        tree.posonlyargs = []
         code = compile(tree, '<conditional:%s>'% func_name, 'exec')
 
         data = {}
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 
 language: python
 python:
+  - 3.8
   - 3.7
   - 3.6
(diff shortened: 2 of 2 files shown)


GarkGarcia:
@rocky I'm not sure how I'm supposed to fix the "posonlyargs" thing. Could you take a look at the relevant code in the diffs?

mathics/builtin/numpy_utils/with_numpy.py
```diff
@@ -330,7 +330,10 @@ def transform(self):
 
         tree = self.visit(tree)
         tree = ast.fix_missing_locations(tree)
-        tree.posonlyargs = []
+        
+        for node in tree.body:
+            node.args.posonlyargs = []
```
rocky, line 335: I believe node.args already has a posonlyargs field when in Python3.8+ (At least that's what my debugger told me when I last looked a while ago.)

GarkGarcia: So why is `compile` raising an error?

rocky:
@GarkGarcia sure, I'd be happy to look at when I get a chance. 

As best as I can tell, that point in the program is exactly where thing start to get messed up with respect to `compile`. Why, other than in general terms I don't understand yet. 

The thought though that I had around this was to take a step back and add better tracing and printing of the ASTs so that it is clearer what is going on. This would also start to address the concern in https://groups.google.com/g/mathics-devel/c/ODff1xp1o2k and somewhere in the list of issues (which I can't find now) there is mention of expanding the API.

rocky:
@GarkGarcia Changes are now in  https://github.com/mathics/Mathics/pull/839

Python's AST node creation are a bit weird and I guess some of this is intentional. You can create incomplete AST nodes like `ast.FunctionDef()` that are incomplete, and that I guess is intentional: you don't need to fill in all the information initally, just _eventually_. H

owever *what* needs to get filled in is not indicated in the docstrings. What is ther seems to be somewhat generic 

This is is from the [Python 3.8 docs](https://docs.python.org/3/library/ast.html)

```
 arguments = (arg* posonlyargs, arg* args, arg? vararg, arg* kwonlyargs,
                 expr* kw_defaults, arg? kwarg, expr* defaults)
```

And since `posonlyargs` is not optional (ends in "?") it is *required*. 

In digging around in this, I believe there are other subtle AST errors or things that are less than ideal if this is to be exposed more and I think it needs to be, especially if you want to show Python equivalances which I think is a good idea. 

Originally I thought I might have to clear some of this up in order to figure out how to get 3.8 working, but for now what is in the PR seems to be the bulk if it. The other stuff will require much larger changes. (But again, in my opinion, all for the good.)

GarkGarcia:
See https://github.com/mathics/Mathics/pull/839 for updates

PR #839, opened 2020-09-07
Python 3.8, 3.9 and Pypy 3.6 compatability

rocky:
and with this mathics can get to a 3.8 `In[]` prompt.

This is not the most elegant or efficient code, but it works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,16 @@
 
 language: python
 python:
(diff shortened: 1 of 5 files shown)


rocky:
With these changes both py3.6 and 3.9 now seem to pass.

GarkGarcia:
Great work!

Issue #847, opened 2020-09-08
"TemplateDoesNotExist" error when running Mathics from PyPy3


mmatera:
I was trying to compile and run mathicsserver after compiling with pypy3, but I found the error in the web browser
"""
TemplateDoesNotExist at /

main.html
"""

rocky:
I am seeing it in 3.7.9 as well. 

```
$ mathicsserver
Migrating database /home/rocky/.local/var/mathics/mathics.sqlite
loading symbols
Operations to perform:
  Apply all migrations: auth, contenttypes, sessions, sites, web
Running migrations:
  No migrations to apply.

database initialized sucessfully

Mathics 1.1.dev0
on CPython 3.7.9 (default, Sep  5 2020, 09:56:01) 
using SymPy 1.6.2, mpmath 1.1.0, django 1.11.29

Copyright (C) 2011-2020 The Mathics Team.
This program comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under certain conditions.
See the documentation for the full license.

Quit by pressing CONTROL-C

Open the graphical user interface at
http://localhost:8000
in Firefox, Chrome, or Safari to use Mathics

loading symbols
Internal Server Error: /
Traceback (most recent call last):
  File "/home/rocky/.pyenv/versions/3.7.9/lib/python3.7/site-packages/django/core/handlers/exception.py", line 41, in inner
    response = get_response(request)
  File "/home/rocky/.pyenv/versions/3.7.9/lib/python3.7/site-packages/django/core/handlers/base.py", line 249, in _legacy_get_response
    response = self._get_response(request)
  File "/home/rocky/.pyenv/versions/3.7.9/lib/python3.7/site-packages/django/core/handlers/base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/home/rocky/.pyenv/versions/3.7.9/lib/python3.7/site-packages/django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/rocky/src/external-vcs/github/rocky/Mathics/mathics/web/views.py", line 67, in main_view
    return render(request, 'main.html', context)
  File "/home/rocky/.pyenv/versions/3.7.9/lib/python3.7/site-packages/django/shortcuts.py", line 30, in render
    content = loader.render_to_string(template_name, context, request, using=using)
[...]

PR #849, opened 2020-09-09
Fix django

suhr:
Fixes https://github.com/mathics/Mathics/issues/847.


rocky:
Speaking of Django, what would it take to upgrade from 1.0 to 3.0 and what benefit would that offer?

suhr:
We want to use the latest LTS release to avoid bit rot without updating too often. By the way, 1.11 is out of date, we will need to upgrade to 2.2.

PR #854, opened 2020-09-09
Add environment

rocky:
Will do the  more complex {G,S}etEnvironment as another PR and the adding the latter will make automatic testing easier.

However, I am of the camp of many small PRs.


rocky:
Superseded by #859

PR #856, opened 2020-09-09
Partial fix for documentation

mmatera:
This patch fixes a part of the problem in compiling the documentation. In a second patch, I hope to be able to fix some issues with the LaTeX code in the documentation. Most of the errors in the pdf compilation are related with empty tests, and unicode chars that should be replaced by latex commands in the documentation.


mathics/doc/doc.py
```diff
@@ -142,7 +142,7 @@ def repl_python(match):
 
     text = _replace_all(text, [
         ('\\', '\\\\'), ('{', '\\{'), ('}', '\\}'),
-        ('~', '\\~{ }'), ('&', '\\&'), ('%', '\\%'),
+        ('~', '\\~{ }'), ('&', '\\&'), ('%', '\\%'), ('#','\#')
```
suhr, line 145:
```suggestion
        ('~', '\\~{ }'), ('&', '\\&'), ('%', '\\%'), ('#','\\#')
```

mathics/doc/doc.py
```diff
@@ -190,7 +190,9 @@ def repl_list(match):
 
     text = _replace_all(text, [
         ('$', r'\$'), ('\u03c0', r'$\pi$'), ('≥', r'$\ge$'), ('≤', r'$\le$'), ('≠', r'$\ne$'),
-    ])
+        ('\u29E6','\equiv'), ('\u22BB','\xor'), ('\uF523',r'\Rightarrow'),
+        ('ç',r'\c{c}'),('é','\'e'),('ê','\^e'),('ñ','\~n'),
+         ('∫',r'\int'),('','d'),   ])
```
suhr, line 195:
```suggestion
        ('\u29E6', r'\equiv'), ('\u22BB', r'\xor'), ('\uF523', r'\Rightarrow'),
        ('ç', r'\c{c}'),('é', r'\'e'),('ê', r'\^e'),('ñ', r'\~n'),
        ('∫',r'\int'),('', r'd'),
    ])
```

rocky:
I will probably build off of this over the weekend in a top-level Makfile.

mmatera:
One problem that I am trying to face is about Unicode symbols. I tried to replace them by latex commands, but in some places, it results in new errors, due to the latex environment where the code is placed.
Another alternative is to compile with xelatex. I am going to see is that direction provides a better solution...

GarkGarcia:
> One problem that I am trying to face is about Unicode symbols. I tried to replace them by latex commands, but in some places, it results in new errors, due to the latex environment where the code is placed.
> Another alternative is to compile with xelatex. I am going to see is that direction provides a better solution...

Xetex is probably the way to go. It usually works much better than `inputec`

PR #860, opened 2020-09-11
Add Combinatorica Package

### LINKED ISSUE #733: Use existing non-copyrighted Mathematica packages that are written in Mathematica
On my machine, /usr/local/Wolfram/Mathematica/11.1/AddOns/Packages/Combinatorica/Combinatorica.m notes:

> This package may be copied in its entirety for nonprofit purposes only

and the code itself appears to be written in Mathematica (it's a standard ".m" file).

There are several such files, not just in Mathematica 11, but going back to older versions of Mathematica as well (plus several on the Internet).

Including these functions in Mathics should be easy.

rocky:
This works and is cool. More work will come down the line to remove various errors.
    
Fixes #733

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,6 +68,7 @@ else:
 INSTALL_REQUIRES += [
     "sympy>=1.6, < 1.7",
     "django >= 1.8, < 1.12",
+    "llvmlite",
     "mpmath>=1.1.0",
     "python-dateutil",
     "colorama",
--- a/setup.py
+++ b/setup.py
@@ -71,6 +71,7 @@ INSTALL_REQUIRES += [
     "mpmath>=1.1.0",
     "python-dateutil",
     "colorama",
+    "palettable",
 ]
 
 
@@ -233,6 +234,7 @@ setup(
         "Intended Audience :: Science/Research",
         "License :: OSI Approved :: GNU General Public License v3 (GPLv3)",
         "Programming Language :: Python",
+
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
--- a/setup.py
+++ b/setup.py
@@ -74,9 +74,7 @@ INSTALL_REQUIRES += [
     "palettable",
 ]
 
-# TODO Remove this check. This is only meant to prevent CI for failing entirelly
-#      due to incorrect configuration
-if (is_PyPy or not sys.version_info == (3, 8)) and (not is_PyPy or sys.version_info == (3, 6)):
+if not ((not is_PyPy and sys.version_info == (3, 8)) or (is_PyPy and sys.version_info == (3, 6))):
     INSTALL_REQUIRES.append("llvmlite")
 
 
--- a/setup.py
+++ b/setup.py
@@ -68,7 +68,6 @@ else:
 INSTALL_REQUIRES += [
     "sympy>=1.6, < 1.7",
     "django >= 1.8, < 1.12",
-    "llvmlite",
     "mpmath>=1.1.0",
     "python-dateutil",
     "colorama",
--- a/setup.py
+++ b/setup.py
@@ -74,6 +74,11 @@ INSTALL_REQUIRES += [
     "palettable",
 ]
 
+# TODO Remove this check. This is only meant to prevent CI for failing entirelly
+#      due to incorrect configuration
+if (is_PyPy or not sys.version_info == (3, 8)) and (not is_PyPy or sys.version_info == (3, 6)):
+    INSTALL_REQUIRES.append("llvmlite")
+
 
 def subdirs(root, file="*.*", depth=10):
     for k in range(depth):
--- a/setup.py
+++ b/setup.py
@@ -74,7 +74,7 @@ INSTALL_REQUIRES += [
     "palettable",
 ]
 
-if not ((not is_PyPy and sys.version_info == (3, 8)) or (is_PyPy and sys.version_info == (3, 6))):
+if not ((not is_PyPy and sys.version_info >= (3, 8)) or (is_PyPy and sys.version_info >= (3, 6))):
     INSTALL_REQUIRES.append("llvmlite")
 
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ matrix:
 services:
   - xvfb
 before_install:
-  - sudo apt-get update -qq && sudo apt-get install -qq liblapack-dev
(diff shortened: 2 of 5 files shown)


GarkGarcia:
Apparently there are unrelated bugs concerning  the CI configuration (see https://travis-ci.org/github/mathics/Mathics/jobs/726264405#L2848).

GarkGarcia:
> Apparently there are unrelated bugs concerning the CI configuration for Python 3.8 (see https://travis-ci.org/github/mathics/Mathics/jobs/726264405#L2848). I'll investigate if they were introduced by the changes in this PR or by the recent commits.

I image it's the latter.

rocky:
If adding llvm-dev doesn't work there is also llvm-3.8-dev but tying this to a version seems more fragile. 

If 2332a32 doesn't work then please continue. I have other stuff I need to do right now.

rocky:
@GarkGarcia It failed on my fork so I expect it will fail here too. If you can investigate - great! If not I guess this can wait. Thanks.

GarkGarcia:
> @GarkGarcia It failed on my fork so I expect it will fail here too. If you can investigate - great! If not I guess this can wait. Thanks.

I'll try to address this week. I'm not quite sure how adding a MMA package to the repository would involve `llvm`, but I'll do my be 😅️

GarkGarcia:
It's also very weird this only fails on Python 3.8. Maybe `llvmlite` is out of date?

GarkGarcia:
If it was a `llvmlite` thing I'd expect all builds from today to fail too. All builds passed except for this one.

setup.py
```diff
@@ -51,24 +51,33 @@
...
+    CMDCLASS = {"build_ext": build_ext}
+    INSTALL_REQUIRES += ["cython>=0.15.1"]
 
 # General Requirements
-INSTALL_REQUIRES += ['sympy>=1.6, < 1.7', 'django >= 1.8, < 1.12',
-                     'mpmath>=1.1.0', 'python-dateutil', 'colorama']
+INSTALL_REQUIRES += [
+    "sympy>=1.6, < 1.7",
+    "django >= 1.8, < 1.12",
+    "llvmlite",
```
GarkGarcia: I see. This is the cause of the issue.

GarkGarcia: @rocky Why did you add `llvmlite` to the install dependencies?

rocky: That was in the last force push. It to try to addresss the `..so` not getting found by the previous commit.

rocky:
> > @GarkGarcia It failed on my fork so I expect it will fail here too. If you can investigate - great! If not I guess this can wait. Thanks.
> 
> I'll try to address this week. I'm not quite sure how adding a MMA package to the repository would involve `llvm`, but I'll do my be


That package contains C code for efficiency. So my view is that what's going on is that by pulling this in we are now testing more complete usage of Mathics.

rocky (commented):
Note that the way we are making tests "pass" is by basically skipping any
tests that are compiled:


```
test_if_eq (test_compile.FlowControlTest) ... skipped 'No llvmlite detected. Skipping all compile tests'
test_if_int_real (test_compile.FlowControlTest) ... skipped 'No llvmlite detected. Skipping all compile tests'
test_if_real_int (test_compile.FlowControlTest) ... skipped 'No llvmlite detected. Skipping all compile tests'
test_if_return1 (test_compile.FlowControlTest) ... skipped 'No llvmlite detected. Skipping all compile tests'
test_if_return2 (test_compile.FlowControlTest) ... skipped 'No llvmlite detected. Skipping all compile tests'
test_if_return3 (test_compile.FlowControlTest) ... skipped 'No llvmlite detected. Skipping all compile tests'
```

Down the line I think we want to *fix* those failing tests (on 3.8 and pypy 3.6)

GarkGarcia:
> > > @GarkGarcia It failed on my fork so I expect it will fail here too. If you can investigate - great! If not I guess this can wait. Thanks.
> > 
> > 
> > I'll try to address this week. I'm not quite sure how adding a MMA package to the repository would involve `llvm`, but I'll do my be
> 
> That package contains C code for efficiency. So my view is that what's going on is that by pulling this in we are now testing more complete usage of Mathics.

Yeah, I just noticed some tests were skipped due to this. Well, it clearly works for Python <=3.7, so I might add a check for that in `setup.py`.

GarkGarcia:
> Down the line I think we want to fix those failing tests (on 3.8 and pypy 3.6)

Fair enough. This isn't strictly related to the PR, so I'll merge the relevant changes and create an issue for us to address the `llvmlite` situation. Is that ok for you?

rocky:
> Fair enough. This isn't strictly related to the PR, so I'll merge the relevant changes and create an issue for us to address the llvmlite situation. Is that ok for you?

If I understand correctly 

* we'll require llvmlite in setup.py except in Pyhton versions where we know it is broken. (pypy 3.6 also is in this camp). 
* For now, CI tests will continue to succeed where they would have failed were llvmlite installed, but
* we'll have an issue to track that this (eventually) gets fixed. 

Thanks.

GarkGarcia:
> * we'll require llvmlite in setup.py except in Pyhton versions where we know it is broken. (pypy 3.6 also is in this camp).

My idea was to revert the change that added `llvmlite` to the requirements entirely, bu sure, we could add the check you mentioned. I'll work on it and then I'll merge the PR.

rocky:
> My idea was to revert the change that added `llvmlite` to the requirements entirely, bu sure, we could add the check you mentioned. I'll work on it and then I'll merge the PR.

Thanks. A problem with including Combinatorica but not llvmlite is that anyone uses mathics with something that requires compilation will notice that they need llvmlite (as I did). 

There are failures in Combinatorica outside of needing compilation that fail, but I hope will focus Mathimatic on ways it should improve. So in my opinion given the the choice of imperfect Combinatorica or not because we don't support compilation, I think adding Combinatorica with flaws is better.

But the other Combinatorica failures are very small compared to not having llvlite or more generally having Mathics compilation broken unless you install something. 

A person installing mathics on an OS where llvmlite works, but doesn't have it previously installed will notice the failure and do the simpler steps to get this installed and that is preferrable in my opinion for the next release.

It does add a little "bloat" because, but that is because _by default_ want people to have a better experience. 

The modularity issues should be addressed after the next release. Breaking this up will cause a lot of havoc, although in the end will be much better. And people will have the next release, 1.1 as a fallback for the more ambitious refactoring if that doesn't work. And I don't think it wise to do the modularity here and now because 1.0 isn't a good fallback position, in my opinion.

PR #862, opened 2020-09-11
Fix documentation errors regarding UTF8

mmatera:
Well, 
I was able to fix the utf8 errors using xelatex instead of pdflatex, and the issue of the empty tests. There are still errors coming from the asy interface for graphics generation. I am not an expert on that package, so I will need some help from here. Otherwise, we could simply remove the tests that generate problematic graphics.

### DIFF
--- a/mathics/doc/tex/Makefile
+++ b/mathics/doc/tex/Makefile
@@ -1,10 +1,11 @@
-TEX = pdflatex
+#TEX = pdflatex
+TEX = xelatex
 LATEXMK = latexmk
 #-quiet
 
 mathics.pdf: mathics.tex documentation.tex
 	$(LATEXMK) -pdf -pdflatex=$(TEX) mathics
-	
+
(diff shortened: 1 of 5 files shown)


mathics/doc/tex/Makefile
```diff
@@ -1,10 +1,11 @@
-TEX = pdflatex
+#TEX = pdflatex
```
GarkGarcia, line 1: We should probably remove this comment. It may  cause confusion in the future.

GarkGarcia: This is the only issue I found in this commit.

mmatera: I have removed it. Shall I make a rebase first?

GarkGarcia: > It's not necessary. There are no conflicts in your branch 😁️

PR #866, opened 2020-09-12
Remove remnants of old Combinatorica code...

rocky:
Instead import the package with Needs["Combinatorica`"]

This code is more complete and up to date.


rocky:
Needs work. Will revise later.

PR #868, opened 2020-09-12
Fix documentation

suhr:


suhr:
Hm, for some reason pydot generates invalid `dot` file.

.travis.yml
```diff
@@ -56,7 +56,7 @@ before_install:
       python -m nltk.downloader wordnet omw;
     fi
   - if [[ "$DOC" == "true" ]]; then
-      sudo apt-get install -qq graphviz asymptote texlive-latex-extra texlive-latex3 texlive-fonts-recommended;
+      sudo apt-get install -qq python-pydot graphviz asymptote texlive-latex-extra texlive-latex3 texlive-fonts-recommended;
```
rocky: pydot should be in setup.py. So why not do something like `pip install -e .` instead?

.travis.yml
```diff
@@ -56,7 +56,7 @@ before_install:
       python -m nltk.downloader wordnet omw;
     fi
   - if [[ "$DOC" == "true" ]]; then
-      sudo apt-get install -qq python-pydot graphviz asymptote texlive-latex-extra texlive-latex3 texlive-fonts-recommended;
+      sudo apt-get install -qq python3-pydot graphviz asymptote texlive-latex-extra texlive-latex3 texlive-fonts-recommended;
```
rocky: See other comment about installing pydot via pip

.travis.yml
suhr: Replaced this with `pip install -e .`, but the issue still remains.

rocky: I think there is a missing a semicolon: 

```
"if [[ "$DOC" == "true" ]]; then ... sudo apt-get install -qq inkscape; pip install -e .;  fi" 
                                                                                        ^
```

suhr: See https://github.com/mathics/Mathics/pull/868/commits/eb7db57dded90c6b9bf8d9997221745c00308da2

rocky: Ok. This is better - it is a _different_ failure and is a failure in running dot:
```
1851 (12): TEST EdgeCount[g, _UndirectedEdge]
"dot" with args ['-Tdot', '/tmp/tmpb67g448g'] returned code: 1
stdout, stderr:
 b''
b"Error: /tmp/tmpb67g448g: syntax error in line 6 near '`'\n"
```

And this comes from https://github.com/mathics/Mathics/blob/master/mathics/builtin/graphs.py#L1122

So here dot is finding a syntax error on line 6 of the generated dot file near backtick at the end of the line.

When I try to  see this in something that shows graphs, I see get an error in this example. 

So I think the problem is not in Travis CI but a bug in the graph code. 

If you are interested in seeing a run, why not remove graphs.py and see if the rest works? (You don't have to put in a PR for that.) 

Down the line I think we'll figure out how to fix graphs.py so it can display graphs.

PR #877, opened 2020-09-13
Add Github Actions

suhr:


suhr:
For some reason the action does not start.

GarkGarcia:
@mmatera @rocky Any objections to this?

mmatera:
The only objection I have is that we would need to maintain two different test schemes. But until the next release, I think we could check both and then see...

rocky:
Let's try it. I like the dea of running both side by side to see which works better. 

One feature that Travis CI has that we are not using is that it can test MacOSX. (I use [Appveyor](https://www.appveyor.com/) to test on Windows, but it looks like it *also* works for MacOSX.

One reason CI takes so long is that we are running many simultainous tests. One way that Open Source projects increase parallelism is by running on different CI platforms (in parallel). So the doc testing could be done on say CircleCI or Github Actions if that is faster).

PR #881, opened 2020-09-13
Update README.rst

rocky:
Used wrong markup language syntax.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,9 +12,8 @@ Installing
 
 To install so that you run from the source tree:
 
-```
-   $ make develop
-```
+    $ make develop
+   
 
 More detailed information is available on the Mathics Wiki. `here <https://github.com/mathics/Mathics/wiki/Installing>`_.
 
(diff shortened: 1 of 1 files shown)


suhr:
Should we keep it in rst format, or maybe it's better to convert it to markdown?

mmatera:
We could mention here the possibility of using the iwolfram jupyter interface. I tested it on my machine, and the only problem is that I need to extend the standard timeout of jupyter for launching the kernel because mathics takes a lot to load...

GarkGarcia:
> Should we keep it in rst format, or maybe it's better to convert it to markdown?

I think we should keep using rst for now. We could change that in a later PR.

rocky:
And advantage of rst is that it can be used on PyPI in the description. 

Other than that, I don't have a preference.

rocky:
> We could mention here the possibility of using the iwolfram jupyter interface (maybe marked ad experimental...). I tested it on my machine, and the only problem is that I need to extend the standard timeout of jupyter for launching the kernel because mathics takes a lot to load...

Sure, mark it as experimental. I haven't been able to get iwolfram working yet. I do however plan at some point to spend time to work on it. 

I was/am thinking more strategically in the long run.

PR #882, opened 2020-09-13
Fix basic.yml

suhr:


rocky (commented):
@suhr I should have mentioned this before:

Now that we have a Makefile in place, `make install` is does most of the `pip install` commands
And `make test` runs test.

The reason this is preferred (aside from it being shorter) is that when things change, we just adjust the Makefile, not the install instructions and the CI commands and so on. Therefore down the line using the Makefile will reduce maintainance I believe. 

Are you up for changing this? 

```
       pip install ...

suhr:
Well, `make check` does not work.

rocky:
> Well, make check does not work

Yes. That's because pytest is not installed. I forget what the correct way in Python how to indicate dependencies for running tests as opposed to what is needed for install. Anyone? 

For now a workaround is to put in the `pip install pytest` before `make check`. Or if you know the Pythonic way to get pytest installed that's fine too. 

The thing I like to keep in mind is that development isn't necessary linear in that things always get incrementally better. The main point of trying not to put stuff in CI for installing that is relevant other places still applies in order to keep maintainance lower.

If you use the same setup code to test locally as on CI, then you are likely to catch the problems earlier and when you fix it locally you are also fixing it on CI.

rocky:
I note that github actions takes about the same amount of time as TravisCI for the stuff that it does. Note that it doesn't try to build with all modules, nor try to build docs.

In my experience CircleCI offers the fasted free CI.

suhr:
Well, since Python is not a language I'm experienced with, I'll just make it work instead of figuring out how to use `make`. Feel free to change it later.

suhr:
> I note that github actions takes about the same amount of time as TravisCI for the stuff that it does.

Except it builds everything in parallel, unlike Travis CI.

rocky:
> Except it builds everything in parallel, unlike Travis CI.

What does this mean? 3.6 and 3.7 and 3.8 builds go in parallel. I think up to 4 or 5 jobs go on parallel on the free plan. you want more parallelism, you have to pay. 

What's the deal with github workflows? Unlimited paralleism? 

Or do you mean somehting else?

suhr:
> What's the deal with github workflows? Unlimited paralleism?

Yes. It's very easy to run into parallelism limits with Travis CI, compared to Github actions.

rocky:
> Yes. It's very easy to run into parallelism limits with Travis CI, compared to Github actions.

Ok. If you would like to expand github actions for pypy3.6 and for building docs and for building with all of the module, that wouldl speed up to overall CI time.  Thanks.

rocky:
Does Github actions on test MacOS and Windows platforms?

GarkGarcia:
> Does Github actions on test MacOS and Windows platforms?

I don't think so.

GarkGarcia:
@suhr Can I merge the PR?

suhr:
No point to merge this since https://github.com/mathics/Mathics/pull/883 already fixes the CI.

Issue #840, opened 2020-09-08
Fix error in the Make target for mathics.pdf


GarkGarcia:
Apparently an error in the Make target for `mathics.pdf` is causing builds with `DOC="true"` to fail, as can be seen in https://travis-ci.org/github/mathics/Mathics/jobs/725204768#L6770.

mmatera:
Yes, I was working yesterday on fixing it. The first thing would be to include in .travis.yml a line

sudo apt-get install -qq inkscape texlive-font-utils;

to install epstopdf and inkscape.
However, after that, it still show errors, maybe because the position of the directory where documentation is built.

GarkGarcia:
> Yes, I was working yesterday on fixing it. 

Great! 

> However, after that, it still show errors, maybe because the position of the directory where documentation is built.

Yeah, TeX command-line tools tend to be behave awkwardly on regards to this sort of thing.

rocky:
I've been meaning add a top-level Makefile and I have been meaning to look at how the docs are made.

One annoyance I currently have is that my editor (Emacs) likes to remove trailing blanks and that can mess up the tests because sometimes in docstrings that is necessary. That is another thing on my list of things to address.

GarkGarcia:
> I've been meaning add a top-level Makefile

That would be much better than the current shell-scripts scattered around the repository.

GarkGarcia:
@mmatera Maybe the remaining errors are related to https://github.com/mathics/Mathics/issues/609?

mmatera:
> @mmatera Maybe the remaining errors are related to #609?

Indeed, the errors comes from #609 and the unicode chars. I think both could be filtered in the routines building the documentation. Maybe this is work for the next week...

GarkGarcia:
Apparently https://travis-ci.org/github/mathics/Mathics/jobs/727081135#L6576 fails because of a missing pickle file... I'm not sure what it's supposed to contain.

mmatera:
I am working with  @rocky to fix this.

GarkGarcia:
> I am working with @rocky to fix this.

Great! Are you guys working in a PR? Which PR is it?

GarkGarcia:
@rocky @mmatera I was able to fix the "No such file or directory error", but the docs won't compile still because of other things. Please take a look at #886 if you can.

mmatera:
It's green now... good!

GarkGarcia:
> It's green now... good!

[...]

PR #886, opened 2020-09-14
Fixed missing file error in write_latex()

GarkGarcia:
This is a follow up to https://github.com/mathics/Mathics/issues/840#issuecomment-692163808.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,6 +71,7 @@ script:
          python pymathics/test.py -t &&
          cd pymathics/doc/tex &&  make latex;
       else
+         python test.py -o
          make doc;
       fi
     else
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,8 +64,8 @@ install:
   - sed -i "s/'sympy==[0-9]\.[0-9]\.[0-9]', //" setup.py
   - if [[ "$DOC" == "true" ]]; then python setup.py develop; else python setup.py install; fi
 script:
-  - make check
   - if [[ "$DOC" == "true" ]]; then
+      make check output="--output";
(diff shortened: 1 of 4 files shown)


mathics/test.py
```diff
@@ -127,8 +127,15 @@ def test_tests(tests, index, quiet=False, stop_on_failure=False, start_at=0):
     return count, failed, skipped, failed_symbols, index
 
 
-def create_output(tests, output_xml, output_tex):
+def create_output(tests, output_xml=None, output_tex=None):
+    if output_xml is None and output_tex is None:
+        raise TypeError(
+                "At least one if output_xml and output_tex must be specified")
```
GarkGarcia, line 133: one of*

.travis.yml
```diff
@@ -71,6 +71,7 @@ script:
          python pymathics/test.py -t &&
          cd pymathics/doc/tex &&  make latex;
       else
+         python test.py -o
```
GarkGarcia, line 74: I think this line is redundant. In other words, we might be running the tests twice. I'll investigate this.

GarkGarcia: Yes, this line made it so that the tests were run twice. That wouldn't be fun, the tests already take up to 20 minutes to finish.

GarkGarcia:
@rocky @mmatera It's fixed! Finally!

GarkGarcia:
I'll push one more commit and merge the PR. The commit simply adds `-halt-on-error` to the flags passed to `xelatex`. This will stop `xelatex` from freezing the tests when some LaTeX package isn't properly installed.

PR #887, opened 2020-09-14
Various Makefile bug fixes and improvements

rocky:
We need to run python mathics/test.py *twice*
Once with -o, and *then* once with -t. We can't combine these options.

Add top-level "clean" target.
Add .PHONY targets and do it properly
Turn a make comment into a remake-style comment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ PYTHON ?= python3
 PIP ?= pip3
 RM  ?= rm
 
-PHONY: all build check clean develop dist doc pytest test rmChangeLog
+.PHONY: all build check clean develop dist doc pytest test rmChangeLog
 
 #: Default target - same as "develop"
 all: develop
@@ -35,7 +35,7 @@ clean:
 	   $(MAKE) -C $$dir clean; \
 	done
 
-#: Run py.test tests. You can environment variable "o" for pytest options
+#: Run py.test tests. You can set environment variable "o" for pytest options
 pytest:
 	py.test test $o
 
--- a/Makefile
+++ b/Makefile
@@ -7,9 +7,9 @@
 GIT2CL ?= admin-tools/git2cl
 PYTHON ?= python3
 PIP ?= pip3
-RM      ?= rm
+RM  ?= rm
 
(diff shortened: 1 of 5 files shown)


rocky (commented):
@mmatera  Did you delete (the buggy) pymathics/Makefile? The diff I saw looked like you just removed some spaces to setting the `RM` variable, but the file disappeared in master. If you need it deleted that's okay. 

In my rush I've been making too many errors. I need to stop working on this project for a little bit.

mmatera:
> @mmatera Did you delete (the buggy) pymathics/Makefile? The diff I saw looked like you just removed some spaces to setting the `RM` variable, but the file disappeared in master. If you need it deleted that's okay.
> 
> In my rush I've been making too many errors. I need to stop working on this project for a little bit.

No, I didn't erase it (voluntarily, at least). I couldn't work too much today so I just was doing small tests.

Makefile
```diff
@@ -28,7 +28,14 @@ install:
 
 check: pytest doctest
 
-# Run py.test tests. You can environment variable o for pytest options
+# FIXME More directories will be added as Makefiles get improved
+#: Remove derived files
+clean:
+	for dir in mathics/doc; do \
+	   $(MAKE) -C $$dir clean; \
```
GarkGarcia, line 35: We should place `"`s around `$$dir`.

pymathics/Makefile
```diff
@@ -0,0 +1,34 @@
...
+develop:
+	$(PIP) install -e .
+
+#: Set up to run from source
+install:
+	$(PYTHON) setup.py install
+
+#: Run tests. For now there's not much here
+check:
+	python test.py
```
GarkGarcia, line 30: We should pass `-o` to `test.py` (so that the `data` files are saved).

GarkGarcia: Never mind, we should definitively not do that.

Makefile
GarkGarcia: Fixed on https://github.com/mathics/Mathics/pull/887/commits/c7be29e0337a3a4a86d20d832f706b37c841f4ef.

PR #890, opened 2020-09-15
Trailing space workaround

rocky:
@GarkGarcia  Finally figured out how to make the doctests be compatibe with editors that don't like trailing blanks on a line. 

In this  Make variable `output` was changed to `opt`(ion). `-o` is an _option_ that is commonly use in `test.py`, but you could pass options other than `-o`, e.g. `-t`. 

Sorry for the confusion here. I was keeping the variable name short since to keep the `remake --tasks` description short. Also since this is user typed, some prefer short variable names. However that's all a matter of taste.


rocky:
Sigh. Needs work.

PR #894, opened 2020-09-15
Fix #861

### LINKED ISSUE #861: Fix CI configuration issues regarding llvmlite
This a followup to https://github.com/mathics/Mathics/pull/860#pullrequestreview-486937901. Essentially, there is an (unknown) issue in the CI configuration that causes `llvmlite` to fail to find some relevant `.so` file. This prevents us from testing some features on certain platforms (CPython 3.8 and PyPy 3.6).

Up to now, we've been dealing with this issue [by simply skipping the tests that require `llvmlite`](https://github.com/mathics/Mathics/pull/860#pullrequestreview-486937901). A recent patch (https://github.com/mathics/Mathics/pull/860/commits/cb080a1ed09bdf6a5ef610d39985199f8f06423a and https://github.com/mathics/Mathics/pull/860/commits/ca74b1ed7a98da4388ee35a641e5db2b371c80fd, https://github.com/mathics/Mathics/pull/860/commits/ca74b1ed7a98da4388ee35a641e5db2b371c80fd and https://github.com/mathics/Mathics/pull/860/commits/1923b042519877f9a35d23171da673dc5e6bd81d) made it possible to tests features that require `llvmlite` on older platforms. However, this needs to be properly addressed at some point.

See https://travis-ci.org/github/mathics/Mathics/jobs/726327847 for the logs of the failing builds.

GarkGarcia:
This is a follow up to https://github.com/mathics/Mathics/issues/861.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -75,11 +75,8 @@ INSTALL_REQUIRES += [
     "pydot", # For graphs
     "python-dateutil",
     "colorama",
+    "llvmlite",
 ]
-
-if not ((not is_PyPy and sys.version_info >= (3, 8)) or (is_PyPy and sys.version_info >= (3, 6))):
-    INSTALL_REQUIRES.append("llvmlite")
-
 
 def subdirs(root, file="*.*", depth=10):
     for k in range(depth):
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,8 @@ services:
 before_install:
   - sudo apt-get update -qq && sudo apt-get install -qq liblapack-dev llvm-10-dev llvm-dev
   - sudo apt-get install -qq gfortran
-  - pip install numpy
+  - pip install numpy conda
+  - conda install --channel=numba llvmlite
   - if [[ "$ALL_MODULES" == "true" ]]; then
       sudo apt-get install -qq &&
       pip install ipywidgets ipykernel requests IPython==5.0.0 nltk langid pycountry pyenchant lxml matplotlib networkx &&
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,8 +35,7 @@ services:
 before_install:
   - sudo apt-get update -qq && sudo apt-get install -qq liblapack-dev llvm-10-dev llvm-dev
   - sudo apt-get install -qq gfortran
-  - pip install numpy conda
-  - conda install --channel=numba llvmlite
+  - pip install numpy
   - if [[ "$ALL_MODULES" == "true" ]]; then
       sudo apt-get install -qq &&
       pip install ipywidgets ipykernel requests IPython==5.0.0 nltk langid pycountry pyenchant lxml matplotlib networkx &&
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ matrix:
 services:
   - xvfb
 before_install:
-  - sudo apt-get update -qq && sudo apt-get install -qq liblapack-dev llvm-dev
+  - sudo apt-get update -qq && sudo apt-get install -qq liblapack-dev llvm-10-dev llvm-dev
   - sudo apt-get install -qq gfortran
   - pip install numpy
   - if [[ "$ALL_MODULES" == "true" ]]; then


GarkGarcia:
@mmatera I'm curious if the recent switch from focal to bionic will help with this.

GarkGarcia:
@rocky Now the only errors come from PyPy. I'm not sure why.

rocky:
I am not sure I agree with the motivation behind this, given what I can determine from the code.

Here are the facts as I understand it. 

`plot.py` would like to use compiled code, because it is slow and would need speeding up. 

Therefore it issues:

```
from mathics.builtin.compile import _compile, CompileArg, CompileError, real_type
```

and `_compile` needs `llvmlite.`

On PyPy there is an error because the shared object for llvmlite is not found. It is not found because I guess it is not installed. It is not installed because we now don't ask for it to be installed.

Given this, here is my view. Because compiling is (or shoud be) a necessary feature, `llvmlite` is a *necessary* dependency. If it _happens_ to work on some OS's and versions of CPython, then you are lucky. But that doesn't change the fact that is _necessary_.  For some random user on some random OS, I believe the preferable thing to do is to fail at installation time, which basically says: "go install llvmlite", rather than succeed on setup and then be slow, and have failures down the line. 

While, sure, there is an advantage to allowing this to run on a system without llvmlite, to start out, I don't think we can or have the resources to find and check all of places where we need to disable Compile. 

So the more responsible thing in my opinion is that if Mathics needs llvmlite, it should state that up front.

setup.py
```diff
@@ -75,12 +75,9 @@
     "pydot", # For graphs
     "python-dateutil",
     "colorama",
+    "llvmlite",
 ]
 
-if not ((not is_PyPy and sys.version_info >= (3, 8)) or (is_PyPy and sys.version_info >= (3, 6))):
-    INSTALL_REQUIRES.append("llvmlite")
```
rocky, line 81: Removing this is the part I don't agree with. 

Probably should go the other way and *always*  require "llvmlite". See the other comment for more detail here.

rocky (changes requested):

setup.py
GarkGarcia: But we are _always_ requiring... See L78, I've added `"llvmlite"` to `INSTALL_REQUIRES` again.

rocky: Ah - Ok. Then this is just about simplifying the code - great! I don't have any suggestions then. Doing via conda sound fine.

mmatera:
OK, the error in the pypy3 test is here:

"""
 from llvmlite.binding import ffi

../../../virtualenv/pypy3.6-7.3.1/site-packages/llvmlite-0.34.0-py3.6.egg/llvmlite/binding/ffi.py:153: in <module>

    raise OSError("Could not load shared object file: {}".format(_lib_name))
"""
 pypy3 do now uses the ubuntu package for llvmlite library, but generates it when it makes the  Mathics setup:

"""
Installed /home/travis/virtualenv/pypy3.6-7.3.1/site-packages/Mathics-1.1.dev0-py3.6.egg
Processing dependencies for Mathics==1.1.dev0
Searching for llvmlite
Reading https://pypi.org/simple/llvmlite/
Downloading https://files.pythonhosted.org/packages/0b/96/07bfa93a103fb9e3e9ae7f9f7c6687ae714aee66b6f3000da3fad71e0aa2/llvmlite-0.34.0.tar.gz#sha256=f03ee0d19bca8f2fe922bb424a909d05c28411983b0c2bc58b020032a0d11f63
Best match: llvmlite 0.34.0
Processing llvmlite-0.34.0.tar.gz
Writing /tmp/easy_install-x2geitxl/llvmlite-0.34.0/setup.cfg
Running llvmlite-0.34.0/setup.py -q bdist_egg --dist-dir /tmp/easy_install-x2geitxl/llvmlite-0.34.0/egg-dist-tmp-gtyv3sqb
zip_safe flag not set; analyzing archive contents...
llvmlite.binding.__pycache__.ffi.pypy36: module references __file__
llvmlite.tests.__pycache__.customize.pypy36: module references __file__
creating /home/travis/virtualenv/pypy3.6-7.3.1/site-packages/llvmlite-0.34.0-py3.6.egg
Extracting llvmlite-0.34.0-py3.6.egg to /home/travis/virtualenv/pypy3.6-7.3.1/site-packages
Adding llvmlite 0.34.0 to easy-install.pth file
"""

For some reason, it seems it can not find llvmlite.so library at this point, and then fails.  Maybe we would need to check which version of the library is this package looking for...

mmatera:
This is the output when the python package llvmlite in installed in the standard python setup:

"""
Installed /home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/Mathics-1.1.dev0-py3.8.egg
Processing dependencies for Mathics==1.1.dev0
Searching for llvmlite
Reading https://pypi.org/simple/llvmlite/
Downloading https://files.pythonhosted.org/packages/2b/f8/104e35345995e2ca35acff0aa429cab79677344fd54cd025eecebd0b01b1/llvmlite-0.34.0-cp38-cp38-manylinux2010_x86_64.whl#sha256=05253f3f44fab0148276335b2c1b2c4a78143dfa78e6bafd7f937d6248f297cc
Best match: llvmlite 0.34.0
Processing llvmlite-0.34.0-cp38-cp38-manylinux2010_x86_64.whl
Installing llvmlite-0.34.0-cp38-cp38-manylinux2010_x86_64.whl to /home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages
Adding llvmlite 0.34.0 to easy-install.pth file
"""

mmatera:
This is a similar issue in another project:
https://github.com/numba/llvmlite/issues/525

It seems that this line before installing Mathics could help 

LLVM_CONFIG=/path/to/llvm-config CXXFLAGS=-fPIC  pypy3 -m pip3 install llvmlite

mmatera:
Please, see  PR #897. I think I have solved the issue.

rocky:
I think you have. Very nice.

GarkGarcia:
> Please, see PR #897. I think I have solved the issue.

You probably did, but we still need to remove the patch I mentioned in https://github.com/mathics/Mathics/issues/861, so that I gets tested.

PR #895, opened 2020-09-16
In-memory sessions

rocky:
See #623


rocky:
@GarkGarcia Should the PR that this is based on get closed too?

GarkGarcia:
> @GarkGarcia Should the PR that this is based on get closed too?

I think GitHub marks them as _merged_ too. This happens when every commit from a PR is integrated into master.

PR #896, opened 2020-09-16
Update CHANGES.rst

rocky:
Run "make ChangeLog", look at the output and then add to CHANGES.rst

Note: this is intended to be a _summary_ of all the changes, not a blow by blow description.  So it is sufficient to mention that some files were blacken formatted once rather than for every time it was done. 

Often a feature is added and then fixed several times. From the standpoint of the user, all that matters is that the feature was added. 

Administrative items such as fixing CI endlessly, I don't think those need to be mentioned.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,8 +33,9 @@ matrix:
 services:
   - xvfb
 before_install:
-  - sudo apt-get update -qq && sudo apt-get install -qq liblapack-dev llvm-dev
+  - sudo apt-get update -qq && sudo apt-get install -qq liblapack-dev llvm-10 llvm-10-dev
   - sudo apt-get install -qq gfortran
+  - LLVM_CONFIG=/usr/bin/llvm-config-10  CXXFLAGS=-fPIC python -m pip install llvmlite
   - pip install numpy
   - if [[ "$ALL_MODULES" == "true" ]]; then
       sudo apt-get install -qq &&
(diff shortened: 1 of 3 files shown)


GarkGarcia:
@rocky Do you consider this to be finished? It looks like `CHANGES.rst` is up-to-date now.

GarkGarcia:
There are some bug fixes that come to mind which are not mentioned in `CHANGES.rst`, such as https://github.com/mathics/Mathics/issues/795. @rocky Could I push a commit to your branch?

rocky:
Pluse amend as you see fit. This is just a start. Undoubtably there will be more updates.

GarkGarcia:
I couldn't find any other issue/feature that wasn't mentioned in `CHANGES.rst`. @rocky I guess we're done with this commit? We should enforce the practice of logging changes introduced by PRs before we merge them, but I'm not sure how this could be done.

rocky:
I think done for now. There will be updates in the future,. but this gets us closer to reducing the work for a release.

I don't know of an automated practice for summarizing and making sure that PR's get added. It could be added as part of some sort of developer process or contributing guide. 

But now that this is relatively up to date, as PR's get merged the PR should include a change to CHANGES if it is relevant.

GarkGarcia:
> as PR's get merged the PR should include a change to CHANGES if it is relevant

Yes. That's what I meant by _"We should enforce the practice of logging changes introduced by PRs before we merge them"_. Maybe we could somehow test this? The problem with that is that there are PRs which shouldn't be mentioned in `CHANGES.rst`.

I'll merge the PR for now.

PR #908, opened 2020-09-19
Sansserif

mmatera:
I took the @poke1024  proposal in PR  #582, and added a system variable
System`$UseSansSerif
to control the behavior. Notice that as a system variable, is protected, so to set its value, it first needs to be unprotected:
Unprotect[System`$UseSansSerif]; System`$UseSansSerif=True

This is how it looks in the django's interface:

imagen


rocky:
Can we get rid of the need to unprotect this? A variable that is user-changable shouldn't be protected in the first place. 

(Who are we "protecting" this from? )

GarkGarcia:
@mmatera Great work! @mmatera I'm inclined to wait to the 1.1 release to merge this, just to avoid introducing bugs, is that ok to you?

rocky:
I think this should go in on 1.1.  Personally, I am more concerned about broken Django being in there than this not being in there.

mmatera:
> @mmatera Great work! @mmatera I'm inclined to wait to the 1.1 release to merge this, just to avoid introducing bugs, is that ok to you?

For me it is fine. In any case, we could think in a better name for the variable. This is just a proof of concept.

Issue #867, opened 2020-09-12
Address failures in Combinatorica Package


rocky:
A number of routines fail in the recent Combinatorica Package. Go over that and fix them where possible. 

(I'll take this)

rocky:
@GarkGarcia This shouldn't be in a 1.1 milestone. The failures are numerous and pervasive However the package is still useful and many things work. 

My suggestion is to deal with those after 1.1.

GarkGarcia:
> @GarkGarcia This shouldn't be in a 1.1 milestone. The failures are numerous and pervasive However the package is still useful and many things work.
> 
> My suggestion is to deal with those after 1.1.

I'd like to try to address the failures before releasing the code. As you said, it's far from ideal to release buggy code. But I'm okay removing it if you feel like you won't have enough time to get everything working.

rocky:
We can remove. I'd suggest though just having it hang out in a branch somewhere.

rocky:
As I look at Graphs, although this may seem radical  I think that should get chucked too. It looks pretty straightforward to reimplement all of this (and do a better job) in networkx. 

As with Combinatorica it can hang out in a branch. 

Actually, now that I think of it, the code could all probably be there (in pymathics) but just not enabled.  So if someone does one to use it in its buggy state, there is something trivial like uncommenting out some code that will do the trick.

GarkGarcia:
> We can remove. I'd suggest though just having it hang out in a branch somewhere.

That works, sure. I image removing Combinatory from master won't cause any tests to fail, so I'm inclined to that. Could you work on this? I could send you an invitation that become a project collaborator in GitHub, which would allow you to create a separate branch for Combinatorica.

> I doubt there will ever be merge conflicts with this because it is all so self-contained.

Yes, well-written commits usually don't cause conflicts (your contributions are well-written) 😀️

GarkGarcia:
> As I look at Graphs, although this may seem radical I think that should get chucked too. It looks pretty straightforward to reimplement all of this (and do a better job) in networkx.
> 
> As with Combinatorica it can hang out in a branch.

I agree. I'm inclined to do that too. Storing Graphs in a separate branch shouldn't cause any conflicts with everything else.

> Actually, now that I think of it, the code could all probably be there (in pymathics and packages) but just not enabled.
> 
> So if someone does wants to use it in its buggy state, there is something trivial like uncommenting out some code that will do the trick.
> 
[...]

PR #918, opened 2020-09-21
Remove Combinatorica from the master branch

GarkGarcia:
This is a follow up to https://github.com/mathics/Mathics/issues/867#issuecomment-695356831. The old code is stored in another branch.


GarkGarcia:
@mmatera I'd like your input on this. @rocky and I discussed this, but I'm not sure you are aware of this and that you agree with this decision.

rocky:
With the addition of `Import@` I think this is much less important. Possibly as an example for `Import@` the example can be the right thing for importing Combinatorica from sources. 

However with Graph now, gone half of this package will be broken.

Issue #905, opened 2020-09-18
Gettting initial page hangs with django 2.2


rocky:
Sometimes connecting to http://localhost:8000 works and is slow to load, sometimes it can be fast. Sometimes it hangs. 

I've seen this both on firefox and chrome. Sometimes disabling the cache helps but doesn't always prevent hangs.

Overall reliabllity is worse for me.

GarkGarcia:
I'm experiencing the same issue. But I'm not sure what's causing it...

rocky:
Can I suggest we keep this as a branch and revert to the old django for now? I have a patch to go between the old and the new.

GarkGarcia:
> Can I suggest we keep this as a branch and revert to the old django for now? I have a patch to go between the old and the new.

I'd like to try to fix this before reverting the update. I probably messed something up in https://github.com/mathics/Mathics/commit/c3a01606756ea1b620a8da3640fb4620b469313f#diff-41ddb83d5ff1c0f9a3d1315f5201799eR56 or in https://github.com/mathics/Mathics/commit/c3a01606756ea1b620a8da3640fb4620b469313f#diff-5f49e5dd9dc2ab5cb52fbd4f6407500eR109.

GarkGarcia:
I remember that replacing `models.CASCADE` instead of `models.SET_NULL` in https://github.com/mathics/Mathics/commit/c3a01606756ea1b620a8da3640fb4620b469313f#diff-41ddb83d5ff1c0f9a3d1315f5201799eR56 solved the hanging issues we were facing, but it introduced some other errors. I honestly have no idea which one is the most appropriate.

GarkGarcia:
As you can see in this `.har` file, the server apparently hangs on some very specific image/js/css files: https://pastebin.com/i6cDHzHr.

You can open the `.har` file by downloading it from pastebin, opening up the Firefox Developer Tool (press `F12`), going to the "Network" tab, clicking the ⚙️ icon on the top-right corner and selecting "Import HAR File".

GarkGarcia:
> As you can see in this `.har` file, the server apparently hangs on some very specific image/js/css files: https://pastebin.com/i6cDHzHr.
> 
> You can open the `.har` file by downloading it from pastebin, opening up the Firefox Developer Tool (press `F12`), going to the "Network" tab, clicking the gear icon on the top-right corner and selecting "Import HAR File".

[...]

PR #919, opened 2020-09-21
Downgrade Django

GarkGarcia:
This is a follow up to https://github.com/mathics/Mathics/issues/905#issuecomment-695233521. Per @rocky's and @mmatera's, I've downgraded Django to version 1.11.


rocky:
I hope we can figure out why we are having problems here.

PR #924, opened 2020-09-27
Add SystemTimeZone, adjust TimeZone..

rocky:
TimeZone (what was there before) isn't protected and can be changed
SystemTimeZone can't be changed. It is the thing that TimeZone acted
like previously.

There is also some small lint-like things of other files.

### DIFF
--- a/mathics/builtin/datentime.py
+++ b/mathics/builtin/datentime.py
@@ -538,6 +538,23 @@ class AbsoluteTime(_DateFormat):
         return from_python(total_seconds(tdelta))
 
 
+class SystemTimeZone(Predefined):
+    """
+    <dl>
+    <dt>'$SystemTimeZone'
+      <dd> gives the current time zone for the computer system on which Mathics is being run.
+    </dl>
+
+    >> $SystemTimeZone
+     = ...
+    """
+
+    name = '$SystemTimeZone'
+
+    def evaluate(self, evaluation):
+        return Real(-time.timezone / 3600.)
+
+
 class TimeZone(Predefined):
     """
     <dl>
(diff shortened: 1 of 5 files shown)


mathics/builtin/datentime.py
```diff
@@ -549,10 +566,22 @@ class TimeZone(Predefined):
      = ...
     """
 
-    name = '$TimeZone'
+    name = "$TimeZone"
+    value = Real(-time.timezone / 3600.)
+    attributes = []
```
GarkGarcia, line 571: I don't think this like is necessary.

GarkGarcia: (the `attributes = []` line)

rocky: `value` definitely is. Setting `attributes` might be the default, but I personally I don't mind being explicit about what is being set.  I have read ["explicit is better than impllicit"](https://eng.libretexts.org/Bookshelves/Computer_Science/Book%3A_Making_Games_with_Python_and_Pygame_(Sweigart)/06%3A_Simulate/6.21%3A_Explicit_is_Better_Than_Implicit#:~:text=To%20make%20your%20code%20more,works%20without%20outright%20telling%20them). 

However if you feel strongly about this, I can remove.

GarkGarcia: I'd prefer if we didn't set `attributes` explicitly in this case (for consistency sake). We could create a different pull request to deal with this (add explicit definitions of the `attributes` static field to _all_ `Builtin` subclasses) later.

mathics/builtin/datentime.py
```diff
@@ -549,10 +566,22 @@ class TimeZone(Predefined):
      = ...
     """
 
-    name = '$TimeZone'
+    name = "$TimeZone"
+    value = Real(-time.timezone / 3600.)
```
GarkGarcia, line 570: We should probably avoid duplicating code in here. We could make `value` a static field of `SystemTimeZone` and simply clone `SystemTimeZone.value` in here.

rocky: Ok. Would you just then make the changes and commit over this. Thanks!

GarkGarcia: Sure. I'll work on it.

PR #925, opened 2020-09-27
Remove some Python dependencies from the Mathics package

rocky:
* pyplot is used by graphs.py but that won't come back here
* colorama is used by the CLI main, but here I think we only need  minimal stuff.  For color terminal, and syntax coloring, command history across sessions and probably eventually handling graphics better, use  tmathics, and all but the first, isn't handled currently in this package.

Some linting and removing non-essentail trailing blanks was also done (largely by my environtment).

### DIFF
--- a/mathics/builtin/quantities.py
+++ b/mathics/builtin/quantities.py
@@ -180,7 +180,6 @@ class Quantity(Builtin):
 
     def apply_n(self, mag, unit, evaluation):
         'Quantity[mag_, unit_?StringQ]'
-        Expression('Quantity', mag, unit)
 
         if(self.validate(unit, evaluation)):
             if(mag.has_form("List", None)):
--- a/mathics/web/media/js/inout.js
+++ b/mathics/web/media/js/inout.js
@@ -193,6 +193,7 @@ function showGallery() {
 	    'N[E, 30] (* 30-digit Numeric approximation of E *)',
 	    '30!',
 	    '30! // N',
+	    'Sum[2 i + 1, {i, 0, 10}] (* Sum of 1st n odd numbers (n+1)**2 *)',
 	    'n = 8; 2 ^ # & /@ Range[0, n] (* Powers of 2 *)',
 	    'Total[%] (* Sum is 2 ^ n - 1 *)',
 	    '(* Symbolic Manipulation *)',


rocky:
As with the other PR, if there is stuff you want changed. Just go ahead and do - thanks!

mathics/main.py
```diff
@@ -44,38 +44,7 @@ def __init__(self, definitions, colors, want_readline, want_completion):
...
-            ),
-        }
-
-        # Handle any case by using .upper()
-        term_colors = color_schemes.get(colors.upper())
-        if term_colors is None:
-            out_msg = "The 'colors' argument must be {0} or None"
-            print(out_msg.format(repr(list(color_schemes.keys()))))
-            quit()
+        term_colors = (["", "", "", ""], ["", "", "", ""])
```
GarkGarcia, line 47: Won't this make the output non-coloured?

mathics/builtin/quantities.py
```diff
@@ -190,57 +180,57 @@ def apply_makeboxes(self, mag, unit, f, evaluation):
 
     def apply_n(self, mag, unit, evaluation):
         'Quantity[mag_, unit_?StringQ]'
-        expr = Expression('Quantity', mag, unit)
-        
+        Expression('Quantity', mag, unit)
```
GarkGarcia, line 183: This line isn't necessary.

GarkGarcia:
> As with the other PR, if there is stuff you want changed. Just go ahead and do - thanks!

Sure, I'll work on it.

mathics/main.py
GarkGarcia: Indeed, this make the CLI output non-coloured. I'm not sure this is a good idea...

GarkGarcia: I mean, the only thing that was coloured was the `In[n]:=`/`Out[n]=` text, so it doesn't make much of difference in that regard. I'd just like to get some input from other people in @mathics/maintainers before approving this.

GarkGarcia: If we were to remove the colours from the output we'd have to examine the code more carefully to remove the references to `term_colors`. There's no point in storing empty strings.

rocky: Ths is safe in that in Mathics 1.0 colorama was optional. 

At some point I turned the _optional_ require into a madatory one.  

If this is less controversional will will cause less concern, I'd say just go back to the 1.0 code: https://github.com/mathics/Mathics/blob/v1.0/mathics/main.py and remove the required dependency in setup.py that I added subsequently.

GarkGarcia: > It is safe. In Mathics 1.0 colorama was optional.
> 
> At some point I turned the _optional_ require into a madatory one, adding it to `setup.py`.
> 
> If this is less controversional will will cause less concern, I'd say just go back to the 1.0 code: https://github.com/mathics/Mathics/blob/v1.0/mathics/main.py and remove the required dependency in setup.py that I added subsequently.

Ohh I see. So your idea is to make it optional again? In that case, we're done, we can merge the PR. I thought we were ditching colours altogether in the CLI.

rocky: At first I was going to do that. But since there is resistance, I'm going with what is in 1.0 which makes this optional. 

Here, you'd just pull in https://github.com/mathics/Mathics/blob/v1.0/mathics/main.py

GarkGarcia: > At first I was going to ditch the colors altogether since it is only a half-hearted effort. A better job is done in mathicsscript

I'm actually quite in favour of ditching colorama altogether. My points are:

* I'd like to get some input from other maintainers before committing to this.
* If we are going to ditch colorama, there's no reason to keep the `term_colors` variable around (it's just a constant array of empty strings that is never used for anything).

GarkGarcia: @rocky I can work on the second point if we decide to ditch colorama.

GarkGarcia: > At first I was going to ditch the colors altogether since it is only a half-hearted effort.

Indeed, it doesn't get used much. That's why I'm leaning towards ditching it.

rocky: Getting input from others is great! 

At this point, I don't have much of a vested interest one way or another. So please take this and finish it however you feel best. 

But some points to note, I want to minimize the effort on the CLI here, although have something basic as a fallback is nice. 

Over time other front-ends will be coming into place. Following the pattern of IMathics and iwolfram, they should be separate installable packages that manage their own dependencies, and have their own README's  and documentation. 

After 1.1.0, in mathicsscript, as it is now called I would to see/add support for RsT for documentation for the help system, support for mathplot/sympy graphics both window based and ASCII plots as a fallback (which exists in sympy) and so on.

rocky:
Based on recent conversations, this will be reworked in a more recent PR.

PR #929, opened 2020-09-27
Update the links in the documentation to point to mathics.org

GarkGarcia:
Update the links in the documentation to point to mathics.org instead of mathics.github.io


GarkGarcia:
@mathics/maintainers Any objections?

poeschko:
Seems good! My only comment is that there are a few places where it's still using http that could be changed to https.

GarkGarcia:
> Seems good! My only comment is that there are a few places where it's still using http that could be changed to https.

Good idea!

rocky:
I need commit de2e4bdc in order to make a new docker image - a bootstrap problem.

GarkGarcia:
@rocky I think I'm done. Unless there's something you'd like to change, this is no longer WIP.

rocky:
Sorry, my bad. I had the wrong branch. I'm updating docker info.

PR #933, opened 2020-09-28
Rebase of Faster leading blank patterns #625 

mmatera:


rocky:
Thanks again for doing this! 

Seems like a big potential speedup.  There was that comment about Flatten in the other #625 (which I will close with reference to this) after merge. 

I got the impression that this meant that more could be done, not that this might cause a bug.

rocky:
@mmatera one other thing we should do is update `CHANGES.rst` to reflect this fix. If you could handle, thanks. 

(For little stuff like this I don't see a reason not to just commit the change to that file to master)

PR #936, opened 2020-09-30
Qtime

mmatera:
Let's go again...


rocky:
Thanks! There are lots of changes. I'm curious if we can see the promised speed up.

mmatera:
I guess the advantage is apparent from the time that the tests take. On the other hand, there is also an advantage in memory, because this evaluation method does not require to keep the frames for each call in the arguments at the same time...

mmatera:
By the way, we need to fix the travis check to detect the fails. As it is now, if some symbols fails, the tests still pass in green...

mmatera:
Well, it seems that now again it passes all the tests. I leave this here until the weekend, and then we can decide if we keep it or not.

rocky:
Again, thanks.

rocky:
I've looked this over and don't see anything glaring.  I was also able to confirm the after resuls @poke1024 gave in #619. 

The before results (quadratic behavior) I had a hard time verifying. 

So let's go with this, and if there are problems down the line we'll deal with it then.

Issue #748, opened 2018-08-31
Pattern-matching defining a Symbol: different from Mathematica


rebcabin:
```
ClearAll[parse]

(* Consider the following, which shows that {"p", "q"} matches the pattern
   {tok : Alternatives @@ {"p", "q", "r", "s"}, toks___}
 *)

Print[MatchQ[{"p", "q"},
      {tok : Alternatives @@ {"p", "q", "r", "s"}, toks___}]]

(* However, the pattern-matching seems not to work with the same pattern in the
   when definining a DownValue for a symbol *)

parse[{tok : Alternatives @@ {"p", "q", "r", "s"}, toks___}] :=
    Print[{"2", tok, {toks}}]

(* Here is the proof that it doesn't work: the following does not reduce.
   However, it DOES reduce, as expected, in Mathematica. *)

parse[{"p", "q"}]

(* The following works in both mathics and in Mathematica *)

parse[{tok : "p" | "q" | "r" | "s", toks___}] :=
    Print[{"2", tok, {toks}}]

parse[{"p", "q"}]

```

rebcabin:
Here is the output in Mathematica --- Printing down-values for both definitions of `parse` and the values of the parses themselves:

```
True

{HoldPattern[parse[{tok:p|q|r|s,toks___}]]:>Print[{2,tok,{toks}}]}

{2,p,{q}}

Null

{HoldPattern[parse[{tok:p|q|r|s,toks___}]]:>Print[{2,tok,{toks}}]}

{2,p,{q}}

Null
```

rebcabin:
Here is the output from mathics --- exact same input:

```
True
{HoldPattern[parse[{tok : p | q | r | s, toks___}]] :> Print[{2, tok, {toks}}]}
parse[{p, q}]
{HoldPattern[parse[{tok : p | q | r | s, toks___}]] :> Print[{2, tok, {toks}}], HoldPattern[parse[{tok : p | q | r | s, toks___}]] :> Print[{2, tok, {toks}}]}
{2, p, {q}}
Null

```

rebcabin:
Just a word on why this important. If you're writing a parser, than an explicit pattern like `p | q | r | s` is fine. But if you're writing a parser generator (like yacc), then you don't know the alternatives until run-time and you need to define symbols that apply `Alternatives` to arbitrary expressions, as in `Alternatives @@ someListOfAlternatives`. I'm writing a parser generator, and I'm blocked until this issue is resolved.

PR #937, opened 2020-10-02
Fix runtime bugs seen from GS3.m in per_expr()

rocky:
per_expr() needs to return a in interator return, not the default implied
`return None` as can happen.

In Python 3, // is used for truncated integer division. In Python 2 it
was /. That now returns float in Python 3.

Note the brokenness of self.get_wrappings

To reproduce brokenness before build, in `mathics` run:
```
   <<"https://raw.githubusercontent.com/rebcabin/Mathics/master/mathics/packages/GS3.m"
```

### DIFF
--- a/mathics/core/pattern.py
+++ b/mathics/core/pattern.py
@@ -322,7 +322,7 @@ class ExpressionPattern(Pattern):
 
                         if expr_groups:
                             expr, count = expr_groups.popitem()
-                            max_per_pattern = count / len(patterns)
+                            max_per_pattern = count // len(patterns)
                             for per_pattern in range(max_per_pattern, -1, -1):
                                 for next in per_expr(   # nopep8
                                     expr_groups, sum + per_pattern):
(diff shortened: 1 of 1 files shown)


rocky:
Although this doesn't fully fix GS3.m mentioned in the PR, it doesn't give a traceback most likely closer to correct.

mathics/core/pattern.py
```diff
@@ -322,17 +322,22 @@ def per_expr(yield_expr, expr_groups, sum=0):
...
                         else:
                             if sum >= match_count[0]:
                                 yield_expr([])
+                            # FIXME: Is this right?
+                            # For now we'll return basically no match.
+                            yield None
 
                     # for sequence in per_expr(expr_groups.items()):
                     def yield_expr(sequence):
+                        # FIXME: this call is wrong and needs a
```
GarkGarcia, line 339: Is this difficult to fix?

mathics/core/pattern.py
```diff
@@ -322,17 +322,22 @@ def per_expr(yield_expr, expr_groups, sum=0):
...
-                            max_per_pattern = count / len(patterns)
+                            max_per_pattern = count // len(patterns)
                             for per_pattern in range(max_per_pattern, -1, -1):
                                 for next in per_expr(   # nopep8
                                     expr_groups, sum + per_pattern):
                                     yield_expr([expr] * per_pattern + next)
                         else:
                             if sum >= match_count[0]:
                                 yield_expr([])
+                            # FIXME: Is this right?
```
GarkGarcia, line 333: Yes, I think this is right.

GarkGarcia: You can remove the FIXME.

mathics/core/pattern.py
rocky: Short answer is yes. 

The needs a function is some sort of iterator over pattern groups. I don't understand what is desired here let alone how to write such a generator. It may be related to https://github.com/mathics/Mathics/issues/748

mathics/core/pattern.py
rocky: done in 6baa1b

PR #939, opened 2020-10-03
Allow passing back source-code lines..

rocky:
This allows CLI's with history tracking to be able to do the more
sensible thing and group multiple lines as one item.

It also will make it possible for better error reporting.


GarkGarcia (approved):

GarkGarcia:
@rocky Is there anything else you'd like to change before we merge this?

rocky:
No. Some other time pobably after 1.1.0 I'll add position information when there is an error so that the front ends can use the specific text along with the position inforamtion for error display.

PR #941, opened 2020-10-03
Django 3

suhr:
All right, I'll try to hack around...

### DIFF
--- a/mathics/web/templates/base.html
+++ b/mathics/web/templates/base.html
@@ -1,24 +1,25 @@
 {% extends "base_html.html" %}
+{% load static %}
 
 {% block title %}Mathics{% endblock %}
 
 {% block html_head %}
 <!-- including scriptaculous main does not work in Safari -->
-<script type="text/javascript" src="/media/js/scriptaculous/builder.js"></script>
-<script type="text/javascript" src="/media/js/scriptaculous/effects.js"></script>
-<script type="text/javascript" src="/media/js/scriptaculous/dragdrop.js"></script>
-<script type="text/javascript" src="/media/js/scriptaculous/controls.js"></script>
-<script type="text/javascript" src="/media/js/scriptaculous/slider.js"></script>
-<script type="text/javascript" src="/media/js/scriptaculous/sound.js"></script>
-<!--<script type="text/javascript" src="/media/js/scriptaculous/scriptaculous.js"></script>-->
+<script type="text/javascript" src="{% static 'js/scriptaculous/builder.js' %}"></script>
+<script type="text/javascript" src="{% static 'js/scriptaculous/effects.js' %}"></script>
+<script type="text/javascript" src="{% static 'js/scriptaculous/dragdrop.js' %}"></script>
+<script type="text/javascript" src="{% static 'js/scriptaculous/controls.js' %}"></script>
+<script type="text/javascript" src="{% static 'js/scriptaculous/slider.js' %}"></script>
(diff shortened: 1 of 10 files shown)


suhr:
Using `{% static ... %}` actually seems to fix the issue.

suhr:
Nope, it does not work.

suhr:
Strange... is kinda works, but still doesn't work.

suhr:
Works with Django 3.

setup.py
```diff
@@ -67,7 +67,7 @@
 # General Requirements
 INSTALL_REQUIRES += [
     "sympy>=1.6, < 1.7",
-    "django >= 1.8, < 1.12",
+    "django >= 3.0",
```
rocky, line 70: Please let's put a cap on this like `<4.0`. (If it were me I'd've made the cap 3.2. Otherwise there is likely automatic breakage. when 3.2 or 4.0 come out. 

Thanks again.

mathics/web/templates/base.html
```diff
@@ -1,24 +1,25 @@
...
+<!--<script type="text/javascript" src="{% static 'js/scriptaculous/scriptaculous.js' %}"></script>-->
 {% endblock %}
 
 {% block html_body %}
 
 <header>
 <div id="headerleft">
-  <img id="logo" class="load" src="/media/img/logo-heptatom.svg" height="32" alt="Logo" />
-  <img id="logotext" class="load" src="/media/img/logo-text.svg" height="26" alt="Mathics" />
+  <img id="logo" class="load" src="{% static 'img/logo-heptatom.svg' %}" height="32" alt="Logo" />
```
rocky, line 21: I am curious if this fixes the recent doc problems when you click on links under docs and you see for example:

```[04/Oct/2020 11:39:03] "GET /ajax/doc/ HTTP/1.1" 200 14062
Not Found: /ajax/doc/<function keep_lazy.<locals>.decorator.<locals>.wrapper at 0x7f0dec8105f0>/<function keep_lazy.<locals>.decorator.<locals>.wrapper at 0x7f0dec812050>/
[04/Oct/2020 11:39:11] "GET /ajax/doc/%3Cfunction%20keep_lazy.%3Clocals%3E.decorator.%3Clocals%3E.wrapper%20at%200x7f0dec8105f0%3E/%3Cfunction%20keep_lazy.%3Clocals%3E.decorator.%3Clocals%3E.wrapper%20at%200x7f0dec812050%3E/ HTTP/1.1" 404 4691
```

Or is more stuff like this needed around the ajax links?

rocky (changes requested):

mathics/web/templates/base.html
suhr: Didn't test but probably not. Ajax links are not static.

suhr: This issue should be related to `doc` functions inside `web/views.py`. It seems like `documentation.get_section(part, chapter, section)` fails, so the methods return Http404.

This should be fixed in another PR.

rocky: Ok. Note though that the documentation to work in Django is the last major thing before we can do a release. (other than small stuff like making a pass over docs to make sure they are up to date and look nice) 

If' you'd create the ticket and fix in another PR, I'd be grateful. Thanks.

rocky (approved):
There is one other small change that should be done:  `error_500_view(request)` ->  `error_500_view(request, exception)`. 

I've handled that and the `CHANGES.rst` in cc27eb2

PR #943, opened 2020-10-04
Fix ajax requests

suhr:


rocky (commented):
Thanks! Will check over now and merge when CI finishes.

suhr:
By the way, is there a way to install Mathics with documentation without building PDF docs? I don't want to install LaTeX.

suhr:
Also, we really should use f-strings everywhere.

rocky:
> By the way, is there a way to install Mathics with documentation without building PDF docs? I don't want to install LaTeX.

I am not 100% sure, but try `make mathics/doc/tex/data`. If you have [`remake`](https://remake.readthedocs.io/en/latest/features.html) installed, you might be able to see this by running `remake --tasks`. 

The make command expands to:

```
python mathics/test.py -o
```

After a 1.1 release the way documentation and testing is done really needs to be rethought and redone using current 2020 practice.

rocky:
@suhr  Unless you have more changes, let's get this in.

suhr:
Sure, feel free to merge this.

PR #944, opened 2020-10-04
4 small tweaks:

rocky:
urls.py: blacken
setup.py: remove not necessary dependencies
   - colorama is optional as it was in 1.0
   - pyplot was part of graphs. It will appear in a depencency in the
     corresponding graphs pymathics module
views.py: remove some unused imports

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,8 +40,7 @@ pytest:
 	py.test test $o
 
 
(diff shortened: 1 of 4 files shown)


GarkGarcia (approved):
Sees good to me.

Issue #946, opened 2020-10-04
Some tests are not suitable for a sandbox


suhr:
There are test that fail inside a nix sandbox:

 - OpenRead in Reference of built-in symbols / File Operations
 - SetDirectory in Reference of built-in symbols / File Operations
 - FileFormat in Reference of built-in symbols / Importing and Exporting

Output:

```
4417 ( 3): TEST OpenRead["https://raw.githubusercontent.com/mathics/Mathics/master/README.rst"]
----------------------------------------------------------------------
Test failed: OpenRead in Reference of built-in symbols / File Operations
OpenRead["https://raw.githubusercontent.com/mathics/Mathics/master/README.rst"]
Result: OpenRead[https://raw.githubusercontent.com/mathics/Mathics/master/README.rst]
Wanted: InputStream[...]
Additional output:
General::noopen: Cannot open https://raw.githubusercontent.com/mathics/Mathics/master/README.rst.

4524 ( 1): TEST SetDirectory[]
----------------------------------------------------------------------
Test failed: SetDirectory in Reference of built-in symbols / File Operations
SetDirectory[]
Output:
SetDirectory::cdir: Cannot set current directory to /homeless-shelter.
Wanted:

4653 (10): TEST FileFormat["ExampleData/benzene.xyz"]
----------------------------------------------------------------------
Test failed: FileFormat in Reference of built-in symbols / Importing and Exporting
FileFormat["ExampleData/benzene.xyz"]
Result: Binary
Wanted: XYZ
```

rocky:
"Common knowledge"  is what you learned 10 minutes ago. 

From @GarkGarcia 
(From the Doctest class definition docstring)
"""
DocTests formating rules:

 #> signifies private test that does not appear as part of the documentation
X> outputs the docs as normal, but the test is not run
= compares the result text
: compares an (error) Message
| signifies Print output
```

Since you're in the best position to know what should be omitted, my suggestion is just to mark the tests with `X>` and put in a PR.

PR #949, opened 2020-10-04
Disable some tests

suhr:
Fixes https://github.com/mathics/Mathics/issues/946.

### DIFF
--- a/mathics/doc/doc.py
+++ b/mathics/doc/doc.py
@@ -1113,6 +1113,7 @@ class DocTest(object):
 
     #> signifies private test that does not appear as part of the documentation
     X> outputs the docs as normal, but the test is not run
+    S> the test is ignored in a sandboxed environment
     = compares the result text
     : compares an (error) Message
     | signifies Print outpt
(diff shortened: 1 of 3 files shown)


rocky (commented):
It occurred to me later that it might be slicker to add a new indicator like `S>` to indicate that the test should be skipped _only_ in a sandboxed environment. 

`doc/doc.py`'s DocTest would be the place to change I think. Are you up for it?

mmatera:
Yep: loot at mathics/doc/doc.py  line 1110

suhr:
I added `S>` as a synonym to `X>`. I'm not going to implement it properly in this PR.

suhr:
Feel free to implement it yourself though!

mmatera:
I think it should be easy to replace all X> by S> in the code, and have just a unique notation...

suhr:
Yeah, these tests use `S>` now.

rocky (approved):
LGTM - thanks for doing this!

mathics/doc/doc.py
```diff
@@ -1124,7 +1125,7 @@ def __init__(self, index, testcase):
         # Private test cases are executed, but NOT shown as part of the docs
         self.private = testcase[0] == '#'
         # Ignored test cases are NOT executed, but shown as part of the docs
-        if testcase[0] == 'X':
+        if testcase[0] == 'X' or testcase[0] == 'S':
```
rocky, line 1128: I hit merge too soon. Shouldn't this be 

```
or (testcase[0] == 'S' and os.environ.get("SANDBOXED", False))
```

rocky (commented):
Ooops noiticed that there's no way to have the test fire here. Se comment below.

PR #951, opened 2020-10-07
Review docs system; no trailing blanks required

rocky:
Doc system updates & pass over docs
    
* S> and X> directives were broken. They are less broken now
* struction.py: Combinatorica -> CombinatoricaOld since this is really obsolete
* doc/doc.py: fix S> and X>

Go over examples in system.py

Finally! No required trailing blanks!
    
Add `#<--#` to the end of a test line to signify that the test ends in a single blank

### DIFF
--- a/mathics/builtin/plot.py
+++ b/mathics/builtin/plot.py
@@ -1056,8 +1056,8 @@ class PieChart(_Chart):
 class BarChart(_Chart):
     """
     <dl>
-    <dt>'BarChart[{$p1$, $p2$ ...}]'
-        <dd>draws a bar chart.
+        <dt>'BarChart[{$p1$, $p2$ ...}]'
+        <dd>makes a bar chart with lengths $b1$, $b2$, ....
     </dl>
 
     >> BarChart[{1, 4, 2}]
@@ -1175,8 +1175,8 @@ class BarChart(_Chart):
 class Histogram(Builtin):
     """
     <dl>
-    <dt>'Histogram[{$x1$, $x2$ ...}]'
-        <dd>gives a histogram using the values $x1$, $x2$, ....
+        <dt>'Histogram[{$x1$, $x2$ ...}]'
+        <dd>plots a histogram using the values $x1$, $x2$, ....
     </dl>
 
     >> Histogram[{3, 8, 10, 100, 1000, 500, 300, 200, 10, 20, 200, 100, 200, 300, 500}]
--- a/mathics/test.py
+++ b/mathics/test.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
+
 import sys
 import re
 import pickle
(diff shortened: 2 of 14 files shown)


rocky:
Even though there is more to be done here, this is getting large. So my suggestion here is close this out and keep going in a future PR. 

While there may be a few tests that seem to be failing which I haven't investigated, the longer term plan after 1.1.0 is to do an overhaul on testing so we can do this much better and faster, and in a more modular way.

It is possbile that some of the testing code I added, breaks these but it is also possible and seems likely that some these were always broken but somehow the code didn't register that.

mmatera:
> Even though there is more to be done here, this is getting large. So my suggestion here is close this out and keep going in a future PR.
> 
> While there may be a few tests that seem to be failing which I haven't investigated, the longer term plan after 1.1.0 is to do an overhaul on testing so we can do this much better and faster, and in a more modular way.
> 
> It is possbile that some of the testing code I added, breaks these but it is also possible and seems likely that some these were always broken but somehow the code didn't register that.

Before merge this, I would like which tests are not working now, and to which changes the failure is associated. Probably is just about how the output is sent now with your improvements, and then is just a matter of adding or removing a white space in some places. I could work on this during the weekend.

rocky:
>  I would like which tests are not working now, and to which changes the failure is associated. Probably is just about how the output is sent now with your improvements, and then is just a matter of adding or removing a white space in some places. I could work on this during the weekend.

Go for it!

rocky:
Also, if the testing system still requires a trailing blank at the end of the line, I'd be grateful if that were removed. 

This has caused undue hassle.

mathics/test.py
```diff
@@ -280,7 +273,15 @@ def test_all(
         print("Save TeX")
         with open_ensure_dir(settings.DOC_TEX_DATA, "wb") as output_file:
             pickle.dump(output_tex, output_file, 0)
-    return failed == 0
+        return True
+        return False
```
GarkGarcia, line 277: This line is unreachable...

mathics/builtin/plot.py
```diff
@@ -1056,8 +1056,8 @@ def labels(names):
 class BarChart(_Chart):
     """
     <dl>
-    <dt>'BarChart[{$p1$, $p2$ ...}]'
-        <dd>draws a bar chart.
+        <dt>'BarChart[{$p1$, $p2$ ...}]'
+        <dd>makes a bar chart with lengths $b1$, $b2$, ....
```
GarkGarcia, line 1060: Shouldn't this be `$p1$, $p2$, ...`

rocky:
@mmatera this is taking a while to review and this is becoming something that is useful in going over the docs. 

So I'm merging now, but will not delete the branch so you or others can add more commits and changes on top of this. 

See also #957

mathics/builtin/plot.py
rocky: Oops - yes you are correct. Fixed in dcdf81a7

mathics/test.py
rocky: Thanks! Fixed in dcdf81a7

PR #959, opened 2020-10-11
Fix #958

### LINKED ISSUE #958: The Django interface fails to run the examples in the documentation
This is a follow up to https://github.com/mathics/Mathics/issues/945#issuecomment-706735449. Here's the bug:

Screenshot from 2020-10-11 14-17-36

The issue is that when the user presses the `=` button in the documentation tab to evaluate an example somewhow the example get's HTML-escaped (the spaces are replaced with `&nbsp;`). This is a bug.

I'd like to get this fixed before the 1.1 if it's possible.

GarkGarcia:
See https://github.com/mathics/Mathics/issues/958 for details.

### DIFF
--- a/mathics/web/media/js/prototype/prototype.js
+++ b/mathics/web/media/js/prototype/prototype.js
@@ -645,7 +645,11 @@ Object.extend(String.prototype, (function() {
   }
 
   function unescapeHTML() {
-    return this.stripTags().replace(/&lt;/g,'<').replace(/&gt;/g,'>').replace(/&amp;/g,'&');
+    return this.stripTags()
+      .replace(/&lt;/g,  '<')
+      .replace(/&gt;/g,  '>')
+      .replace(/&amp;/g, '&')
+      .replace(/&nbsp;/g,' ');
   }
 
 


GarkGarcia:
@suhr The error is still there, even with `.replace(/&nbsp;/g,' ')`

GarkGarcia:
> @suhr The error is still there, even with `.replace(/&nbsp;/g,' ')`

Never mind, it's a caching issue. The patch works as expected.

rocky:
I broke thing trying to extend CI testing to 3.9. I've reverted that change in 884a05bb. 

@suhr - maybe you can figure out how to get CI testing for 3.9 going  - see https://github.com/mathics/Mathics/runs/1238691570

Looks like a newer base LLVM is need for llvmlite.

rocky (approved):
This stuff is temporary, so sure.

rocky:
On other thing that I've mentioned and would be good to do in the future: pleaes create branches off of mathics, not your own repo.- it makes it easier to commit on top of. Or here rebase off of master.

PR #961, opened 2020-10-12
145 new tests in symbolic logic; Quit at the ends; not integrated wit…

rebcabin:
…h test framework

I added most of my old tests in symbolic logic. I withheld a few at the bleeding end of GS3.m, which are not visible in this PR.

I tried to figure out how to integrate them with your existing test frameworks. It looks like in pytest I must write all the expressions in AST form, but there is far too much here to do that by orders of magnitude. It didn't seem appropriate to put them in the automated tests in the manual because I am not proposing any new built-ins: these are just applications. But they work the bejeebers out of the evaluator, so I hope you're interested and can use them.


rocky (changes requested):
First and foremost thanks for doing this. We'll make use of it one way or another. 

We already have non-pytest kinds of tests - that is not a problem. For example we have one for testing [Django](https://github.com/mathics/Mathics/blob/master/mathics/test.py).

In looking at this I see that our `Exit[]` (and `Quit[]`) functions do not exit Mathics, let alone exit with a return code.  I have a PR #962 in to address this.

Once that clears, all I would ask is that instead of calling `Quit[]`, use `Exit[Min[totalWrong, 255]]`.  I recall from David Korn that setting a max value of 255 works on most OS's, but I could be wrong here. 

Having set a return value in `Exit[]`, if I have this correct (and I often don't: that's one reason I work on debuggers), then this along with a simple shell script looping over GS1..3 is enough to tell CI whether things are okay. We can inspect the logs for further info, when things go wrong.

Well, there may also be the weirdness of how to get into Mathics in such a way so that `GS0.m` is included. That may require another Mathics fix to be able do what in POSIX is `cd`.

I hope you are able to make these small changes. Don't give up on us just yet...

rebcabin:
Will be very happy to make the requested changes. There's no giving up on you, I'm already committed to using mathics for an algebraic optimizer for a code generator for a hardware projects.

rebcabin:
Ok, just pulled mathics upstream master into my fork (rebseal/master); did make & make check; made my little changes as requested above; and pushed to rebseal/master. Exit isn't quite working yet, so I followed it with a "just-in-case" Quit.

rocky (commented):
Great - Thanks! 

Let's wait on Exit landing so I can try this. Also a little shell script that's something along the lines of 

```
for gs_test in GS[1-3].m; do 
   if ! mathics -e "<<$gs_test";  then
       exit $?
   fi
exit 0
```

Should be put somewhere where it will be run with CI.

rocky:
Although incomplete with respect to getting this hooked into CI. Let's merge and continue...

rebcabin:
pulled and run on Python 3.7 (3.6 sticks on test_console). Thanks!

rocky:
@rebcabin I just spent $25 of my hard-eanred money buying the book. I have Gries' other books. (Well, I threw out the compiler constructuion book a while ago.)

PR #962, opened 2020-10-12
Exit exits Mathics; it can set return code too.

rocky:
I don't understand why we had an Exit doesn't exit and is an alias for Quit. 

It is very useful in testing to have a real Exit from Mathics with a return code.

### DIFF
--- a/mathics/builtin/evaluation.py
+++ b/mathics/builtin/evaluation.py
@@ -54,26 +54,27 @@ class RecursionLimit(Predefined):
     #> ClearAll[f];
     """
 
-    name = '$RecursionLimit'
+    name = "$RecursionLimit"
     value = 200
 
     set_python_recursion_limit(value)
 
     rules = {
-        '$RecursionLimit': str(value),
+        "$RecursionLimit": str(value),
     }
 
     messages = {
-        'reclim': "Recursion depth of `1` exceeded.",
-        'limset': (
+        "reclim": "Recursion depth of `1` exceeded.",
+        "limset": (
             "Cannot set $RecursionLimit to `1`; "
             "value must be an integer between 20 and %d; "
-            "use the MATHICS_MAX_RECURSION_DEPTH environment variable to allow higher limits.") % (
-                MAX_RECURSION_DEPTH),
+            "use the MATHICS_MAX_RECURSION_DEPTH environment variable to allow higher limits."
+        )
+        % (MAX_RECURSION_DEPTH),
     }
 
     rules = {
-        '$RecursionLimit': str(value),
+        "$RecursionLimit": str(value),
     }
 
     def evaluate(self, evaluation) -> Integer:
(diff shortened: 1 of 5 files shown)


mmatera:
Actually, Quit[n] should return to the operative system an exit code n. https://reference.wolfram.com/language/ref/Quit.html
I do not know why this function was included, but I think it was related to an attempt to fix the behaviour of Quit /Quit[]. In any case, I would copy this behaviour in Quit too.

rocky:
> Actually, Quit[n] should return to the operative system an exit code n. https://reference.wolfram.com/language/ref/Quit.html
> I do not know why this function was included, but I think it was related to an attempt to fix the behaviour of Quit /Quit[]. In any case, I would copy this behaviour in Quit too.

I am for doing that as well. It would mean changing that doc test since implementing the correct behavior will mean leaving Mathics. But doing things this way is like the tail wagging the dog: The tests are supposed to serve checking the correct functionality. 

If you are up for taking it from here I'd be grateful.

mmatera:
I can try to do that later. One thing is that, along the tests requiring to clear the definitions, we should use Clear[All] instead Quit...

mmatera:
(and fix Clear[All] to the right behaviour: not to clear the symbol "All", but do what is Quit doing now...)

rocky:
One thing I've been meaning to mention, I think (but am not totally sure) that one way to make one builtin function be an alias of another is to just by using the OO subclassing mechanism. 

For example, to have `Quit[]` be an alias for `Exit[]`  which is implemented

```python
class Quit(Exit):
   """Doc string for Quit .. so that the docstring for exit doesn't kick in...
   """
   # Any other changes to Exit that Quit should have
```

mathics/builtin/evaluation.py
```diff
@@ -450,22 +450,41 @@ class Out(Builtin):
...
     """
 
-    def apply(self, var, evaluation):
-        "Exit[var___]"
-        evaluation.definitions.set_user_definitions({})
-        evaluation.definitions.clear_pymathics_modules()
-        if isinstance(var, Integer):
-            exit(var.get_int_value())
-        exit(0)
+    def apply(self, evaluation, n):
```
rocky, line 483: I think (but am not totally sure, that if you change `Exit(builtin)` to `Exit(Quit)` you don't need to duplicate `def apply()`

mathics/builtin/evaluation.py
```diff
@@ -462,14 +462,14 @@ class Quit(Builtin):
     """
 
     def apply(self, evaluation, n):
-        'Quit[n___]'
+        '%(name)s[n___]'
```
rocky, line 465: Very slick!

mmatera: Seems to work...

rocky:
Looks good. If there is more - we can continue in more PRs.

GarkGarcia:
> One thing I've been meaning to mention, I think (but am not totally sure) that one way to make one builtin function be an alias of another is to just by using the OO subclassing mechanism.
> 
> For example, to have `Quit[]` be an alias for `Exit[]` which is implemented
> 
> ```python
> class Quit(Exit):
>    """Doc string for Quit .. so that the docstring for exit doesn't kick in...
>    """
>    # Any other changes to Exit that Quit should have
> ```

That should work, yes. But it's more appropriate to include a rule in `rules` static field of the class (such as in https://github.com/mathics/Mathics/blob/master/mathics/builtin/calculus.py#L436)

GarkGarcia:
> I don't understand why we had an Exit doesn't exit and is an alias for Quit.

According to the Wolfram docs, `Exit` is supposed to be an alias for `Quit`. Some time in the past I made `Quit[]` work as you described (actually quitting the front-end), but @mmatera asked me to revert the commit because it interfered with a test case of something he was working on.

mmatera:
Indeed, I did. Now, I removed the doctests for Quit[] (and Exit[]) and the former functionality is held by using ClearAll[All].
The current version then kills the text interface, but not the web interface, where it behaves as an alias for ClearAll[All].

Also, I agree with you into that the better way to implement the alias is by defining a rule. I will try to correct this.

GarkGarcia:
> The current version then kills the text interface, but not the web interface, where it behaves as an alias for ClearAll[All].

That's awesome!

Issue #960, opened 2020-10-12
Bug in Table when substituting a list for an iterator


rebcabin:
(this one is blocking my upcoming PR with many tests in symbolic logic)

Kindly consider the following small, self-contained repro case, which works in Mathematica:

```
ClearAll[tc, or, fn, a, b, t, f]

binaryFunctionList = {tc, or}

binaryTruthTableBugged =
    Table[fn[a, b],
        {fn, binaryFunctionList},
        {a, {t, f}}, {b, {t, f}}]

binaryTruthTableNoBug =
    Table[fn[a, b],
        {fn, {tc, or}},
        {a, {t, f}}, {b, {t, f}}]
```

In mathics, after a fresh start, it produces:

```
Quit[]
In[1]:= <<"GS2bug2.m"
Table::iterb: Iterator does not have appropriate bounds.
Out[1]= {{{tc[t, t], tc[t, f]}, {tc[f, t], tc[f, f]}}, {{or[t, t], or[t, f]}, {or[f, t], or[f, f]}}}

In[2]:= 
```

rocky:
I'll take a look at this when I have a chance. 

Something I've been meaning to ask of those who know Mathematical better than I: if something like this happened in Mathematica, what would you typically use to track down what's going wrong?

mmatera:
Table[F[x],{x,{a,b,c,d}]] 
returns in wma {F[a],F[b],F[c],F[d]}

I will try to implement this.

mmatera:
Already solved (I think). Please have a look at PR #963

rocky:
@rebcabin ... and if this addresses the issue please close.

PR #963, opened 2020-10-12
evaluating the iterator argument before evaluating the iteration func…

mmatera:
This should fix the issue #960 with Table and other iterator functions (like Sum). The issue was due to the lack of evaluation of the iterator argument before the method was chosen.

### DIFF
--- a/mathics/builtin/lists.py
+++ b/mathics/builtin/lists.py
@@ -1971,10 +1971,15 @@ class _IterationFunction(Builtin):
 
     def apply_range(self, expr, i, imax, evaluation):
         '%(name)s[expr_, {i_Symbol, imax_}]'
-
+        imax = imax.evaluate(evaluation)
         if imax.has_form('Range', None):
+            # Fixme: this should work as an iterator in python3, not
+            # building the sequence explicitly...
             seq = Expression('Sequence', *(imax.evaluate(evaluation).leaves))
             return self.apply_list(expr, i, seq, evaluation)
+        elif imax.has_form('List', None):
+            seq = Expression('Sequence',*(imax.leaves))
+            return self.apply_list(expr, i,  seq, evaluation)
         else:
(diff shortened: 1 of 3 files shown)


rocky (approved):
Wow - that was fast! Thanks! 

I wonder if a a py.test in the same fashion as that [`test_evaluation.py`](https://github.com/sealemar/Mathics/blob/master/test/test_evaluation.py) mentioned with respect to the orderless "or" can be written?

mmatera:
> Wow - that was fast! Thanks!
> 
> I wonder if a a py.test in the same fashion as that [`test_evaluation.py`](https://github.com/sealemar/Mathics/blob/master/test/test_evaluation.py) mentioned with respect to the orderless "or" can be written?

Do you mean for Iterators?

rocky:
> > Wow - that was fast! Thanks!
> > I wonder if a a py.test in the same fashion as that [`test_evaluation.py`](https://github.com/sealemar/Mathics/blob/master/test/test_evaluation.py) mentioned with respect to the orderless "or" can be written?
> 
> Do you mean for Iterators?

Yes for the iterators. Is this possible?

mmatera:
> > > Wow - that was fast! Thanks!
> > > I wonder if a a py.test in the same fashion as that [`test_evaluation.py`](https://github.com/sealemar/Mathics/blob/master/test/test_evaluation.py) mentioned with respect to the orderless "or" can be written?
> > 
> > 
> > Do you mean for Iterators?
> 
> Yes for the iterators. Is this possible?

I guess so... The @redcabin  tests would work? Something
s={1,2,3};Table[F[x],{x,s}]

and 
s={x,1,10};Table[F[x],s]
?
(BTW, I am fixing this second too now)

rocky (approved):
Wow this is really fantastic! Thanks!

PR #964, opened 2020-10-13
Hook in Gries & Schneider test into test suite

rocky:


GarkGarcia (approved):

mmatera (approved):

mmatera:
I just notice, that for some reason the new tests break the documentation test, and force to run twice. @rocky , could you check this before merge?

rocky:
The broken doctests are from this old PR #951 https://travis-ci.org/github/mathics/Mathics/builds/734781764 that I had hoped you'd look at.  Part of the root cause I think is changing `#>` from 

As I have come to realize in going over docs what we have now in terms in the doctests is a bit incomplete, and there is this how doc.private vs .ignore and what output to show in the docs issue. It is a mess. 

For examples we may 

* show an example but not run/test it. $ProcessID is like this
* show an example and test it
* don't show an example and not test it either. This is for example setup for thigs like `Get`. File cleanup might be like this too
* don't show an example but test it. 

We currently don't have enough indicators for these. 

Right now I'm happy with whatever surgery is needed to get this done. When I look at the examples for Get/Put etc a number of them are complicated and not of interest. 

We have other means of testing sucn as the recently added test_evaluation.py. That's where some of this should go.

rocky:
> The problem with Get/Put is that we have to write examples that should work on any platform. I

That is easier done outside of the docsystem, as a pure test.

For the docsystem what I think would be nice for things like this is a form of `#>` that says add this to the list of examples but take the output as given. So for example:

```
!>  
<<"MyFile"
```

In this kind of example, don't try to run `<<MyFile``; and if output were specfied, use that blindly.

If you want to test Get and Put write that in Python pytest.

rocky:
Merging. The CI testing will be addressed separately. It is a separate (and messy) issue.

rocky:
@mmatera #968 is a start at simplifying the TravisCI testing and removing duplicte testing. 

 It's failing right now because of the Get/Put issues.

PR #967, opened 2020-10-13
fix Get/Put test

mmatera:
This PR fixes the tests for Get and Put, by using $TemporaryDirectory as the root for temporary files.

PR #968, opened 2020-10-13
Simpler, reduced TravisCI

rocky:
This greatly simplifies testing under TravisCI. It takes less time and reduces the load on TravisCI which means more builds will get done faster. 

However I believe all of the essential testing is done - basically all we want to do here is test the document building with all dependencies enabled and to do Pypy 3.6 testing. 

pymathics is gone - we don't have pymathics modules right now.

Note:  we do other testing using Github workflows. This still takes too long - over 30 minutes, so more work is needed.


mmatera:
OK. Then, if you agree, I will close the fix-filetests and work on this branch, implementing the following changes:

* Remove tests from docstrng involving <<  or >>  which do not allow to work on something that is not a String
* Try Get/Put over files relative to $TemporaryDirectory[].
* Put tests for << and >> in something like  test/test_files.py

mmatera:
Regarding pymathics, the idea would be to put natlang and the other modules back in the builtin folder?
About this, I guess that the best would be to keep apart, and implement tests for each module which do not involve to run the full tests over the builtins.

rocky:
> Regarding pymathics, the idea would be to put natlang and the other modules back in the builtin folder?
> About this, I guess that the best would be to keep apart, and implement tests for each module which do not involve to run the full tests over the builtins.

The pymathics while a great idea right now is incomplete. There was/is some sort of pre-existing knowledge by the core that `natlang` existed. It was so close, but just not ready yet. I was delighted when it was decided to defer this after 1.1.0. 

Personally, what I'd like to see is a `natlang` github project same as `iwolfram` or `mathicsscript`.  Each one is installable on PyPI and each repo has a `setup.py`  which its own dependencies; `nltk` for `natlang`.  Each pymodule has its own test suite and doctest test suite.

The current organization of Mathics seems to suggest that several installable Python modules were to be in one git repository, but personally I find this harder to manage. Yes, there is a downside that some things in say pymodules have to be coordinated with the core.

rocky:
> * Remove tests from docstrng involving <<  or >>  which do not allow to work on something that is not a String
> * Try Get/Put over files relative to $TemporaryDirectory[].
> * Put tests for << and >> in something like  test/test_files.py

Yes this is good, but couldn't the tests be done more along the line of`test_execution` where you call `evalutate()` of a string and check the result if you need to do it then or feed it into some other `evaluate()` and so on, calling `assert` when needed?

mmatera:
#967 fix one part of the problem, and I guess we can just merge it. The other part could be implemented afterwards.

mmatera:
Regarding LoadModule and pymathics, LoadModule works properly, while the pymathics are installed together with mathics, so what is missing now is to improve tests and documentation. In the meantime, we can load them by means of a init.m file loading all the available modules.

rocky:
> In the meantime, we can load them by means of a init.m file loading all the available modules.

It has been a while since I looked at this so my memory may be faulty.
 
But if I recall correctly what was/is missing from pymathics was a way to get the symbols included into the Mathics namespace.  There was some list and those were used to do a Python `import` of some means.

`init.m` doesn't suffer this problem because `evaluation()` handles getting names installed which happens all of the time.

Now that `init.m` works that would be fine for pymathics modules. 

However there is still the issue of organizing the individual packages as Python-installable modules. I started doing a little of that but didn't finish. And now I think this should be done outside of the Mathics repo. 

And on top of that is as you mention the testing and hooking into doctest. Doctest is a little bit better off now, but still it a ways from being usable as a Python module from the outside. 

As I said before all of this felt like a 1000 little papercuts (okay maybe just 4 or 5), and we weren't handling any of this all that well. On top of that the natlang package is probably one of the least interesting and not very complete which was also the case with graphs when you considered its the plotting capability compared to what networkx can _currently_ give.

GarkGarcia (approved):

GarkGarcia:
Seems alright to me.

rocky:
That CI has been backing up today, has been a bit of a trouble here. 

Also, our CI settings had been running tests for both on branch builds and PRs.  This meant that PRs were getting run twice. Since each PR is also a branch build.  It is set now just on a branch build.

PR #971, opened 2020-10-14
Exit as a rule inside Quit

mmatera:


rocky:
It is okay to do this this way, but let's understand what's going on between this and subclassing. 

When things are done as a rewrite,the code size is a little less: there is only a docstring created instead of two classes, but the run time is a little more because we need to go through the rewrite mechanism every time as oppsed to Python's method lookup mechanism. 

Unfortunately right now we don't have good diagnostic information for rewriting. We need that and the future should add this. So in terms of debugging what's going on rewrite is a little more difficult to follow that Python's mechanism. This ultimately will change though. 

For this particluar function which is generally not called or called once only before exiting, worrying about effiicency really doesn't matter.  

My purpose here is just to point out and have us _understand_ that there are tradeoffs and what they are.

mmatera:
Due to the way the interpreter implements the call to the builtin symbols, I think that this way (defining a rule) is better than defining a subclass: a subclass implies to store in the "definitions" object two symbols instead just one.

mmatera:
I merge this because it is just what @GarkGarcia asked, @rocky seems to aprove it,  and is quite easy to undo.

PR #972, opened 2020-10-14
adding tests for  Exit,Quit and file operations

mmatera:
Just basic tests for these functionalities

### DIFF
--- a/test/test_evaluation.py
+++ b/test/test_evaluation.py
@@ -25,3 +25,19 @@ def test_evaluation(str_expr, str_expected):
     expected = _evaluate(str_expected)
 
     assert result == expected
+
+
+
+def test_exit():
+    try:
+        _evaluate("Exit[-37]")
+    except SystemExit as e:
+        assert e.code == -37
+
+
+def test_exit():
+    try:
+        _evaluate("Quit[-37]")
+    except SystemExit as e:
+        assert e.code == -37
+
--- a/test/test_evaluation.py
+++ b/test/test_evaluation.py
@@ -35,7 +35,7 @@ def test_exit():
         assert e.code == -37
 
 
-def test_exit():
+def test_quit():
     try:
         _evaluate("Quit[-37]")
     except SystemExit as e:
--- a/test/test_file.py
+++ b/test/test_file.py
@@ -1,0 +1,36 @@
+from mathics.core.parser import parse, SingleLineFeeder
+from mathics.core.definitions import Definitions
+from mathics.core.evaluation import Evaluation
+import pytest
+
+
+definitions = Definitions(add_builtin=True)
+evaluation = Evaluation(definitions=definitions, catch_interrupt=False)
+
+
+def _evaluate(str_expression):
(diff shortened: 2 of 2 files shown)


test/test_evaluation.py
```diff
@@ -25,3 +25,19 @@ def test_evaluation(str_expr, str_expected):
...
+
+
+def test_exit():
+    try:
+        _evaluate("Exit[-37]")
+    except SystemExit as e:
+        assert e.code == -37
+
+
+def test_exit():
```
rocky, line 38: Note that since `test_exit()` is defined twice, the second one wins and the first does nothiing. 

Did you mean one to be `test_quit()`?

rocky (approved):
Thanks for adding the tests for `<<` and `>>` this way rather than as doc tests.

mmatera:
Indeed, it should be test_quit

test/test_file.py
```diff
@@ -0,0 +1,36 @@
...
+    temp_directory = temp_directory[1:-1]
+    temp_filename = temp_directory + "/testfile"
+    print(temp_filename)
+    result = _evaluate("40! >> " + temp_filename).to_python()
+    assert result is None
+
+    result = _evaluate("<< " + temp_filename)
+    assert result == _evaluate("40!")
+
+    result = _evaluate("DeleteFile[\"" + temp_filename + "\"]").to_python()
```
GarkGarcia, line 29: This shoud probably be an f-string (`f"DeleteFile[\"{temp_filename}\"]"`)

GarkGarcia: Not a big deal thought, we can merge the PR. Down the line I'd like to covert this (and every other instance of string interpolation done with C-style format strings or the `+` operator) to an f-string.

mmatera: Why? I don't get used to use fstrings. What is the advantages of use them?

GarkGarcia: > Why? I don't get used to use fstrings. What is the advantages of use them?

Well, f-strings are the "official" way to do string interpolation in Python nowadays. They are more legible than concatenating strings, and they are more flexible than the printf-like syntax used in the `%` operator (which used to be the "official" way to interpolate strings in Python 2.x).

The example in question doesn't really illustrate this points well enghout, but image you're trying to interpolate 4 values into a string. You can do it the "concatenation route" as you discribed:

```py
"blabla" + x + "bleble" + y + "blibli" + z + "bloblo" + w + "blublu"
```
Or you can do it with f-strings:
```py
f"blabla{x}bleble{y}blibli{z}bloblo{w}blublu"
```

In my opinion, the f-string is a lot more readable than the concatenation thing (I can get a much clearer picture of what the interpolated string is going to look like). F-strings are also safer, for example, `"a" + x + "b"` raises a `TypeError` when `x` is not a string, while `f"a{x}b"` doesn't (`f"a{x}b"` is equivalent to `"a" + str(x) + "b"`). You can find more information on f-strings in [here](https://realpython.com/python-f-strings/).

PR #973, opened 2020-10-15
Set to use SansSerif by default

rocky:
Having used both for a while, as others in the past have observed SansSerif looks nicer than without.

Now that we have online docs this is all the more apparent.

(Not sure why we center, things either way, but I guess that's different)

In discussing this with Gark, the feeling is that it would be nice if we had a user profile for preferences like that could set.

However we don't have that right now. Given this my take is SansSerif is a better default.

### DIFF
--- a/mathics/builtin/importexport.py
+++ b/mathics/builtin/importexport.py
@@ -1084,7 +1084,7 @@ class ExportString(Builtin):
      . 3,
      . 4,
     >> ExportString[Integrate[f[x],{x,0,2}], "SVG"]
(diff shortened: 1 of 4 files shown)


mmatera (approved):

mmatera:
For me this is OK. As a side comment, we should implement some tests in the pytest engine checking that mathml outout has the proper format.

rocky:
> As a side comment, we should implement some tests in the pytest engine checking that mathml outout has the proper format.

@mmatera  Agreed. There should be orders of magnitude more checking in pytest. And pytest can do this faster than the doctest mechanism. 

There is also better segregation of tests spreading out CI that could be done too. 

Part of the goal of #968 was to speed up TravisCI by just focusing on full-module doctests. More could/should be added in gitflows and we could hook up both CircleCI and Appveyor.

PR #982, opened 2020-10-17
Support for comparisons between SympyConstants

mmatera:
Add support to compare SympyConstants


mathics/builtin/__init__.py
```diff
@@ -6,7 +6,8 @@
     algebra, arithmetic, assignment, attributes, calculus, combinatorial, compilation,
     comparison, control, datentime, diffeqns, evaluation, exptrig, functional,
     graphics, graphics3d,
-    image, inout, integer, iohooks, linalg, lists, logic,
+    # image,
+    inout, integer, iohooks, linalg, lists, logic,
```
weakit, line 10: This right here is what I'm talking about.

weakit (commented):
I may be wrong, but there seem to be a few unrelated changes in a few files. And a removed picture in the documentation.
Could you go over them?

It'd also be helpful if you could fix the tests and add a few new ones. I'll take a look at this tommo and maybe try to help.

mathics/builtin/__init__.py
mmatera: Yes, my mistake. I always comment out this module for local testings...

weakit: No probs, nobody's perfect!

weakit:
This wouldn't really work for something like `Max[Pi, Pi + 1]`, would it?

mmatera:
> This wouldn't really work for something like `Max[Pi, Pi + 1]`, would it?

No as it is. This PR requires more work, but I didn't find the time to do it. The idea was first to identify where in the code the comparison is done, and then, look for the smartest way to extend it using sympy.

weakit:
> This PR requires more work, but I didn't find the time to do it.
I think I'll try working on it later today.

rocky:
@mmatera is this still relevant?

weakit:
I don't think so. PR #1000 closed #965.

PR #983, opened 2020-10-17
Go over section titles; add $UserName

rocky:
Start to regularize capitalization in title names and align better
with Mathematica 5 names. (MMA 5 is the book I have laying around).

One 3D gallery example replace with perhaps a better one. 

blacken another file. Make some files not sensitive to trailing blanks.

### DIFF
--- a/mathics/builtin/system.py
+++ b/mathics/builtin/system.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 
 """
-System functions
+Global System Information
 """
 
 
@@ -115,14 +115,18 @@ class GetEnvironment(Builtin):
             env_var = var.get_string_value()
             tup = (
                 env_var,
-                "System`None" if env_var not in os.environ else String(os.environ[env_var]),
+                "System`None"
+                if env_var not in os.environ
+                else String(os.environ[env_var]),
             )
 
             return Expression("Rule", *tup)
 
         env_vars = var.get_sequence()
         if len(env_vars) == 0:
-            rules = [Expression("Rule", name, value) for name, value in os.environ.items()]
+            rules = [
+                Expression("Rule", name, value) for name, value in os.environ.items()
+            ]
             return Expression("List", *rules)
 
 
@@ -223,10 +227,7 @@ class Packages(Predefined):
     def evaluate(self, evaluation):
         return Expression(
             "List",
-            *(
-                String(name)
-                for name in evaluation.definitions.get_package_names()
-            ),
+            *(String(name) for name in evaluation.definitions.get_package_names()),
         )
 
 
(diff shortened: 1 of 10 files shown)


GarkGarcia (approved):

mathics/builtin/system.py
```diff
@@ -347,6 +350,24 @@ def evaluate(self, evaluation):
...
 
+class UserName(Predefined):
+    """
+    <dl>
+      <dt>$UserName
+      <dd>returns a string describing the type of computer system on which
+      \Mathics is being run.
+    </dl>
+
+    X> $UserName
```
mmatera, line 361: This is not suitable for CI. The test should be implemented in pytest

mathics/web/media/js/inout.js
```diff
@@ -225,8 +225,8 @@ function showGallery() {
 	    'Graphics[Table[{EdgeForm[{GrayLevel[0, 0.5]}], Hue[(-11+q+10r)/72, 1, 1, 0.6], Disk[(8-r){Cos[2Pi q/12], Sin [2Pi q/12]}, (8-r)/3]}, {r, 6}, {q, 12}]]',
 	    '(* 3D graphics *)',
 	    'Graphics3D[Sphere[{0, 0, 0}, 1]]',
-	    'Plot3D[Exp[x] Cos[y], {x, -2, 1}, {y, -Pi, 2 Pi}, PlotPoints->21]',
             'Plot3D[Sin[x y], {x, -2, 2}, {y, -2, 2}, Mesh->Full, PlotPoints->21]',
+	    'Plot3D[ Abs[Zeta[x + I y] ], {x, -2, 2}, {y, 2, 20}, PlotPoints->30]',
```
mmatera, line 229: I like this one!

mmatera (commented):
Apart from the issue on the test of $User, the rest I think is fine.

mathics/builtin/system.py
rocky: I am really confused about the meaning of `X>` and how this all works....

Here was the intent...

I want something that *isn't* run in doctest, but appears as an example.  How is this coded.

```
X> $UserName
= rocky
```

or 

```
X> $UserName
= ...
```

or 
``` 
X> $UserName
```

?

In our docs where we explain this stuff, `>>`, `X>`, `#>`, `=`, `...`  it woudl be great if we described this, and gave examples where one is appropriate and why.

GarkGarcia: I'm confused about that too. I didn't realize those where not mentioned in the documentation. I'll try to fix this when I'm refactoring the wiki.

rocky: Ok. Thanks for fixing when you get a chance. 

What should be used when youwant something that isn't run in doctest, but does appear as an example -  and with some output that is typical of output that might be produced?

Issue #974, opened 2020-10-15
`RepeatedNull`  (`...`) has bug; causes  `$RecursionLimit` execeded


rocky:
Split off from #970 so this can be found and tracked separately

This is how WMA handles `...` (`RepeatedNull`):

```
In: RepeatedNull[a]//FullForm
Out: FullForm[a...]

In: a...
Out: a...
```

and this is what happens in Mathics:

```
In: FullForm["a"...]
Out: RepeatedNull["a"]

In: "a"...
Msg: | $RecursionLimit::reclim: Recursion depth of 200 exceeded.
Out: $Aborted
```

rocky:
Here is what's going on as best as I can tell. 

To format the result Mathics "cleverly" decides to use 
```
Postfix`List[HoldForm[RepeatedNull["a"]]], "...", 170]
```

`...`, I guess, is an operator. 

The infinite recursion seems to be on `List[HoldForm[RepeatedNull["a"]]`.  `List` evaluates its arguments, and somehow it is not seeing the `HoldForm`.

mmatera:
Partially addressed by PR #993

PR #988, opened 2020-10-19
WIP: Unprotect["`*"] and Protect["`*"]

rocky:
"`*" is the same as "*`*" - is this right?

### DIFF
--- a/mathics/builtin/attributes.py
+++ b/mathics/builtin/attributes.py
@@ -10,7 +10,7 @@ However, you can set any symbol as an attribute, in contrast to \Mathematica.
 
 
 from mathics.builtin.base import Predefined, Builtin
-from mathics.core.expression import Symbol, Expression
+from mathics.core.expression import Expression, Symbol
 from mathics.builtin.assignment import get_symbol_list
 
 
@@ -109,7 +109,7 @@ class ClearAttributes(Builtin):
     >> ClearAttributes[f, Flat]
     >> Attributes[f]
      = {}
-    Attributes that are not even set are simply ignored:
+    Attributes that are not set are ignored:
     >> ClearAttributes[{f}, {Flat}]
     >> Attributes[f]
      = {}
@@ -119,7 +119,6 @@ class ClearAttributes(Builtin):
 
     def apply(self, symbols, attributes, evaluation):
         'ClearAttributes[symbols_, attributes_]'
-
         symbols = get_symbol_list(symbols, lambda item: evaluation.message(
             'ClearAttributes', 'sym', item, 1))
         if symbols is None:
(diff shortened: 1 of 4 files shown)


rocky:
@GarkGarcia and @mmatera I've been trying to get 

```
Import@"https://raw.githubusercontent.com/jkuczm/MathematicaSyntaxAnnotations/master/NoInstall.m"
```

or any of the Noinstall.m's working that are over at repos in https://github.com/jkuczm/

If you have a chance could you look at this PR and the failure in:


```
In[1]:=Import@"https://raw.githubusercontent.com/jkuczm/MathematicaSyntaxAnnotations/master/NoInstall.m"

Clear::wrsym: Symbol $OperatingSystem is Protected.  Clear::wrsym:
Symbol List is Protected.  Clear::wrsym: Symbol Symbol is Protected.
Clear::wrsym: Symbol {SyntaxAnnotations`} is Protected.  Clear::wrsym:
Symbol Unix is Protected.  Clear::wrsym: Symbol List is Protected.
Clear::wrsym: Symbol Symbol is Protected.  $RecursionLimit::reclim:
Recursion depth of 200 exceeded.  Out[0]= $Aborted
```

The sense I am gertting is that the Recursion depth of 200 may mean that one command has an error itereating over all symbols and there are over 200 such errors. This probalby is a separate problem too. 

I won't have time to work on this for a while, so just continue if you can and are able.

mathics/builtin/attributes.py
```diff
@@ -173,6 +191,19 @@ class Unprotect(Builtin):
         'Unprotect[symbols__]': 'ClearAttributes[{symbols}, Protected]',
     }
 
+    def apply(self, pattern, evaluation):
+        "Unprotect[pattern_]"
```
mmatera, line 195: The defined behavior is different from the established in the docstring: this does not accept a list of patterns, but just one.  
,

mmatera (commented):
Let go by parts, like Jack...
"`*" is equivalent to "Everything in the current context", and should be equivalent to "*"
while "*`*" seems to indicate "Everything in every Context".

mmatera:
Regarding the a..  and a... issue, I found that the \$RecursionLimig  happens when the parser tryies to format Repeated or RepeatedNull. I am working on that, but slowly...

mathics/builtin/attributes.py
rocky: Thanks for pointing out. 

Yes, that should probably be `pattern__` and the code will have to get more complicated too. 

I got the docstring from Mathematica. And I doubt the code handles _form_ like it should as well.

However, the particular problem I wanted to fix was `Unprotect["\[GraveAccent]*]` though which explains the discrepancy. 

I'll address this when I have a chance to come back to it.

rocky:
> Regarding the a.. and a... issue, I found that the `$RecursionLimit` happens when the parser tries to format `Repeated` or `RepeatedNull`. I am working on that, but slowly...

It has nothing to do with the parser, and everything to do with the formatter. I have updated issue #974

mathics/core/definitions.py
```diff
@@ -296,10 +296,13 @@ def get_matching_names(self, pattern) -> typing.List[str]:
...
-            ctx_pattern = ((ctx_pattern + '`')
-                           .replace('@', '[^A-Z`]+')
-                           .replace('*', '.*')
-                           .replace('$', r'\$'))
+            if ctx_pattern == "":
+                ctx_pattern="System`"
+            else:
+                ctx_pattern = ((ctx_pattern + '`')
+                               .replace('@', '[^A-Z`]+')
+                            .replace('*', '.*')
```
GarkGarcia, line 304: We should probably align this properly before merging.

rocky:
All of these changes are subsumed in #1003

PR #991, opened 2020-10-26
Fix #989

### LINKED ISSUE #989: Parser can't differentiate between ReplaceAll and Division

mmatera:
This is a more elegant way to fix #989 than #990

### DIFF
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -658,7 +658,11 @@ class Divide(BinaryOperator):
     >> Pi / 4
      = Pi / 4
 
+    
+
     Use 'N' or a decimal point to force numeric evaluation:
+    >> a /.5
+     = 2. a
     >> Pi / 4.0
      = 0.785398
     >> 1 / 8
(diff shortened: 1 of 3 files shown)


rocky:
I think you have to use  mark a lookahead, i.e. `(?<= ...)` somehow.

mmatera:
You are right! I hope now it works...

mmatera:
So, the unique thing that deserves revision is the test in the docstring. @rocky, if you agree, I merge this.

rocky:
If you ask my opinion, for an example it would need more explanation in the example. 
That casually this looks like it might be `ReplaceAll[]` for example if this were `/.x` instead of `/.2`. 

Otherwise I think it would be better done as a test in `test/test_evaluation.py` because  `a /.5 == 2 a` isn't all that interesting in of itself. 

Your choice though. 

And feel free to merge whenever you like.

PR #993, opened 2020-10-27
Fix ldots

mmatera:
This PR fixes the behavior of Repeated and RepeatedNull and introduce catch exceptions for wrong formatted patterns.
`Import@"https://raw.githubusercontent.com/jkuczm/MathematicaCellsToTeX/master/NoInstall.m"`
 now still does not loads, but at least do not kill the interpreter... 

@rocky please have a look at this.

### DIFF
--- a/mathics/builtin/patterns.py
+++ b/mathics/builtin/patterns.py
@@ -36,7 +36,7 @@ The attributes 'Flat', 'Orderless', and 'OneIdentity' affect pattern matching.
 
 
 from mathics.builtin.base import Builtin, BinaryOperator, PostfixOperator
-from mathics.builtin.base import PatternObject
+from mathics.builtin.base import PatternObject, PatternError
 from mathics.builtin.lists import python_levelspec, InvalidLevelspecError
 
 from mathics.core.expression import (
@@ -206,6 +206,9 @@ class Replace(Builtin):
             return result
         except InvalidLevelspecError:
             evaluation.message('General', 'level', ls)
+        except PatternError as e:
+            evaluation.message('Replace','reps', rules)
+            
 
 
 class ReplaceAll(BinaryOperator):
@@ -267,13 +270,19 @@ class ReplaceAll(BinaryOperator):
 
     def apply(self, expr, rules, evaluation):
         'ReplaceAll[expr_, rules_]'
+        try:
+            rules, ret = create_rules(rules, expr, 'ReplaceAll', evaluation)
 
-        rules, ret = create_rules(rules, expr, 'ReplaceAll', evaluation)
-        if ret:
-            return rules
+            if ret:
+                return rules
 
-        result, applied = expr.apply_rules(rules, evaluation)
-        return result
+            result, applied = expr.apply_rules(rules, evaluation)
+            return result
+        except PatternError as e:
+            evaluation.message('Replace','reps', rules)
+        
+
+
 
 
 class ReplaceRepeated(BinaryOperator):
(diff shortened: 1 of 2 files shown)


rocky:
@mmatera Thanks for doing this. I'll look at Wednesday night.

mathics/core/expression.py
```diff
@@ -367,6 +368,29 @@ def format_expr(expr):
...
+                    if len(leaves)==1:
+                        return Expression("System`HoldForm",
+                                      Expression("System`Postfix",
+                                      Expression(
+                                          "System`List",
+                                          leaves[0]
+                                      ),"...",170))
+                    else:
+                        return Expression("System`HoldForm",expr)
+
```
rocky, line 393: An interesting way to handle this. 

I suspect there is a better way, but I don't know what it is. 

And as you have said it doesn't fully address the problem of 

```
In[1] = a...
Out[1] = a...
```

But this is an improvement.

rocky (approved):
Let's go with this for now. It is an improvement even if it is not ideal or elegant. 

Again, thanks for doing. 

I am assuming this gets us closer to handling more imports which would be super..

mathics/builtin/patterns.py
```diff
@@ -206,6 +206,8 @@ def apply_levelspec(self, expr, rules, ls, evaluation, options):
             return result
         except InvalidLevelspecError:
             evaluation.message('General', 'level', ls)
+        except PatternError as e:
```
GarkGarcia, line 209: There's no need to extract `e` in here.

mathics/builtin/patterns.py
```diff
@@ -267,13 +269,16 @@ class ReplaceAll(BinaryOperator):
...
-        if ret:
-            return rules
+            if ret:
+                return rules
 
-        result, applied = expr.apply_rules(rules, evaluation)
-        return result
+            result, applied = expr.apply_rules(rules, evaluation)
+            return result
+        except PatternError as e:
```
GarkGarcia, line 280: Same as before.

mathics/builtin/patterns.py
```diff
@@ -309,8 +314,12 @@ class ReplaceRepeated(BinaryOperator):
 
     def apply_list(self, expr, rules, evaluation):
         'ReplaceRepeated[expr_, rules_]'
+        try:
+            rules, ret = create_rules(rules, expr, 'ReplaceRepeated', evaluation)
+        except PatternError as e:
```
GarkGarcia, line 319: Same as before.

mathics/builtin/patterns.py
```diff
@@ -374,9 +383,13 @@ def apply(self, expr, rules, max, evaluation):
             if max_count is None or max_count < 0:
                 evaluation.message('ReplaceList', 'innf', 3)
                 return
+        try:
+            rules, ret = create_rules(
+                rules, expr, 'ReplaceList', evaluation, extra_args=[max])
+        except PatternError as e:
```
GarkGarcia, line 389: Same as before.

mathics/builtin/patterns.py
```diff
@@ -1114,7 +1127,8 @@ def init(self, expr, min=1):
         self.min = min
         if len(expr.leaves) == 2:
             leaf_1 = expr.leaves[1]
-            if (leaf_1.has_form('List', 1, 2) and all(leaf.get_int_value() for leaf in leaf_1.leaves)):
+            allnumbers = all(not (leaf.get_int_value() is None) for leaf in leaf_1.get_leaves())
+            if (leaf_1.has_form('List', 1, 2) and allnumbers ):
```
GarkGarcia, line 1131: There's not need for the parenthesis in this condition.

mathics/builtin/patterns.py
```diff
@@ -1114,7 +1127,8 @@ def init(self, expr, min=1):
         self.min = min
         if len(expr.leaves) == 2:
             leaf_1 = expr.leaves[1]
-            if (leaf_1.has_form('List', 1, 2) and all(leaf.get_int_value() for leaf in leaf_1.leaves)):
+            allnumbers = all(not (leaf.get_int_value() is None) for leaf in leaf_1.get_leaves())
```
GarkGarcia, line 1130: `leaf.get_int_value() is not None` is more idiomatic than `not (leaf.get_int_value() is None)`

rocky:
Thanks!

PR #998, opened 2020-11-05
Symbolic comparisions

### LINKED ISSUE #956: Symbolic Min/Max
`Min[E, Pi]` currently does not give a result.

This is what Wolfram gives:

image

`sympy` has inbuilt Min/Max functions that have the same functionality:

```python
>>> Min(E, pi)
E
```

One thing to note tho is that `sympy`'s min/max is pretty slow with large (>1000 element) lists. See sympy/sympy#16249.

weakit:
Closes #956

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,5 @@ mathics/web/media/pdf/
 mathics/web/media/doc/classes.pdf
 mathics/web/media/doc/classes.png
 tmp
+
+.idea/
--- a/mathics/builtin/base.py
+++ b/mathics/builtin/base.py
@@ -385,8 +385,8 @@ class AtomBuiltin(Builtin):
     # which are by default not in the definitions' contribution pipeline.
     # see Image[] for an example of this.
 
-    def get_name(self) -> str:
-        name = super(AtomBuiltin, self).get_name()
+    def get_name(self, short=False) -> str:
+        name = super(AtomBuiltin, self).get_name(short=short)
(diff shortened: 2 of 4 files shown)


mmatera:
It looks nicer, but we need to be careful with how numbers and certain mathics structures handled different in sympy are compared.

weakit:
> It looks nicer, but we need to be careful with how numbers and certain mathics structures handled different in sympy are compared.
Thanks. And yes, I'm working on that.

weakit:
It looks like all the tests have passed. Maybe we can add a few more tests? 
And also could someone help me implement the `General::nord` message? (for when complex numbers are compared)

mmatera:
> It looks like all the tests have passed. Maybe we can add a few more tests?
Great! It would be nice to have more tests if you can think a couple.

> And also could someone help me implement the `General::nord` message? (for when complex numbers are compared)
Regarding the message, you can implement it as an exception, raise it from do_cmp, capture in the different builtins, and raise the message from there.

weakit:
>Regarding the message, you can implement it as an exception, raise it from do_cmp, capture in the different builtins, and raise the message from there.

Thanks! I'll look into that.

GarkGarcia:
@weakit Great work!

weakit:
I couldn't really come up with any more tests. I already added some in my previous commit.
And as for the message, I think I'll tackle that later.

I think this PR can be merged.

PR #1002, opened 2020-11-09
Remove "Protect" Attribute hackiness on Builtin

rocky:
There are a number of built-ins that don't want to be Protected by default, and we currently have no way to indicate that.

Currently "MakeBoxes" is hacked in to change this and "assignment" has another hack to change this as well. And they do this using the `List#clear()` function which removes *all* attributes, not just Protect.

In this change, setting the `attribute = ("Unprotect", ...)`  indicates to Builtin we don't want to add the "Protect" attribute.

Note however this isn't a true attribute like "Protect" is - it only means something in the declaration of a Builtin, and doesn't exist at run time.

The need for this change becomes all the more important as we move things into PyMathics because a growing number of these for variable-like definition will be desired.

Another Note: in the future we may want to add a class specificaly for variable-like definition.

### DIFF
--- a/mathics/builtin/__init__.py
+++ b/mathics/builtin/__init__.py
@@ -114,14 +114,6 @@ def contribute(definitions):
         if name != 'System`MakeBoxes':
             item.contribute(definitions)
 
-    # Is there another way to Unprotect these symbols at initialization?
-    definitions.get_attributes('System`$PreRead').clear()
-    definitions.get_attributes('System`$Pre').clear()
-    definitions.get_attributes('System`$Post').clear()
-    definitions.get_attributes('System`$PrePrint').clear()
-    definitions.get_attributes('System`$SyntaxHandler').clear()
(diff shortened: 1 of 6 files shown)


rocky:
@mmatera @GarkGarcia Will be making a release over the weekend. Would like this to get in. 

Note if there are other things to get in for 1.1, now is the time to start doing this

mmatera:
There is a few things in the celltotex branch that would be nice to include in the release. In particular, the improvement handling assignment to named patterns. I could go over this on Friday evening, I hope...

rocky:
> There is a few things in the celltotex branch that would be nice to include in the release. In particular, the improvement handling assignment to named patterns. I could go over this on Friday evening, I hope...

@mmatera that would be awesome. If covers #988 we can close that. Also if you'd make a pass over CHANGES.rst that'd be awesome too.

PR #995, opened 2020-10-31
WIP Support for Celltotex

mmatera:
In this branch, I am trying to include all the fixes needed to make that CellsToTeX package

`Import@"https://raw.githubusercontent.com/jkuczm/MathematicaCellsToTeX/master/NoInstall.m"` 
This involves improvements in WL native Patterns support, Protect/Unprotect mechanisms, Names[], etc. The idea would be to merge this branch when the Import stops showing errors, including adding some tests.

PR #1003, opened 2020-11-14
Assignnamepattern

mmatera:
Here are the improvements in  #995  that are currently functional.

### DIFF
--- a/mathics/builtin/assignment.py
+++ b/mathics/builtin/assignment.py
@@ -17,6 +17,28 @@ from mathics import settings
 from mathics.core.definitions import PyMathicsLoadException
 
 
+def repl_pattern_by_symbol(expr):
+    leaves = expr.get_leaves()
+    if len(leaves) == 0:
+        return expr
+
+    headname = expr.get_head_name()
+    if headname == "System`Pattern":
+        return leaves[0]
+
+    changed = False
+    newleaves = []
+    for leave in leaves:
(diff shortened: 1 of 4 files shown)


rocky (approved):
Looked at and as far a the code looks, I just see good stuff and improvements. 

No tests as yet. I suppose that can come later. 

Will wait for CI to finish (or maybe it has by now). 

For this PR and the one for sessions it would be good to update CHANGES.rst. 

You are probably in a better position to update this than I am, but if you don't have time to do I'll do this as [part of the release process](https://github.com/mathics/Mathics/wiki/Releasing#update-changes). 

And thanks for moving the ball forward in being able to import external WL project code.

rocky:
After this goes in https://github.com/mathics/Mathics/wiki/Calling-Mathics-library-from-within-Python-code should be updated, right?

mmatera:
I added the changes to CHANGE.rst. I guess that MathicsSession is something that we can include, even if the implementation is very primitive. 
Later I can modify the wiki.

rocky:
@mmatera  thanks. It is all merged now. 

@GarkGarcia I am starting the release process. Please no changes to master until this is done. 

I will be committing to master which is normally not advised and maybe in the future we can change this. 

Here it will simplify things a little.  As is happening right now there are some merges (of comment-like code) that are occuring before the CI process finishes. And there may be more of this. However before the final tagging I'll wait for the 30-minute or so CI process to finish.

PR #1006, opened 2020-11-15
Update docker installation

rocky:
Install pymathics.natlang
Docker files are now in app not docker.

Down the line I'd like to see adding a simple script to run either the web and cli. 

```
$ docker run --rm -it --name mathics-web -p 8000:8000 -v /tmp:/usr/src/app/data mathicsorg/mathics --mode ui
$ docker run --rm -it --name mathics-cli -v /tmp:/usr/src/app/data mathicsorg/mathics --mode cli
```

is a bit too tedious. Not sure how to get that done though.


mmatera:
Maybe just add a docker-mathicsscript and docker-mathicsweb scripts to bin/?

rocky:
> Maybe just add a docker-mathicsscript and docker-mathicsweb scripts to bin/?

Yes, for the unsigned tests that makes sense for those things that do IO that are boring that those would work too. Right now we don't seem to be capturing error messages properly.

PR #1007, opened 2020-11-15
Mac OSX (and probably Windows)  fixes

### LINKED ISSUE #1005: Tests fail on MacOSX (and probably MS Windows)
OS/X doesn't have `/dev/full` so some of the I/O tests fail. 

```
4405 (83): TEST strm = OpenWrite["/dev/full", BinaryFormat -> True]
Test failed: BinaryWrite in Reference of Built-in Symbols / File Operations
strm = OpenWrite["/dev/full", BinaryFormat -> True]
Result: OpenWrite[/dev/full, BinaryFormat -> True]
Wanted: OutputStream[...]
Additional output:
General::noopen: Cannot open /dev/full.

4406 (84): TEST BinaryWrite[strm, {39, 4, 122}]
Test failed: BinaryWrite in Reference of Built-in Symbols / File Operations
BinaryWrite[strm, {39, 4, 122}]
Result: BinaryWrite[OpenWrite[/dev/full, BinaryFormat -> True], {39, 4, 122}]
Wanted: OutputStream[...]

4407 (85): TEST Close[strm]
Test failed: BinaryWrite in Reference of Built-in Symbols / File Operations
Close[strm]
Output:
General::openx: OpenWrite[/dev/full, BinaryFormat -> True] is not open.
Wanted:
::: No space left on device.
```

Although I haven't checked, the same is probably true for MS Windows.

We could/should also set up CI testing on MacOSX.

rocky:
OSX (and probably Windows fixes)
    
    Sandbox some tests if running OSX and Windows
    
    Install scikit-image on OSX.

Fixes #1005


mmatera (commented):
Another way around would be to mark these tests as examples without testing) marking them with # instead S, and move the testing stuff to the pytest engine.

rocky:
> Another way around would be to mark these tests as examples without testing) marking them with # instead S, and move the testing stuff to the pytest engine.

Yes, for the unsigned tests that makes sense for those things that do IO that are boring that those would work too. Right now we don't seem to be capturing error messages properly.

mmatera (approved):

PR #1016, opened 2020-11-21
WIP: combinatorica examples

rocky:
Basically try the examples in the first page of the book.
The test harness has been move from GS0.m to test_driver.m
so we can reuse it. (It really doesn't have anything to do with
Gries-Schneider code.)

<strike>
Note: GS tests were broken and to some extent still are.

Exit[1] no exits, but Quit[] does.

Also mathics -e doesn't catch Aborted and things like that and it
should.

If we caught that, in we'd see that reality GS3.m can sometimes fail in a timeout. 
</strike>
The above was put in another PR and merged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@
 .pydevproject
 .settings
 .vscode
+/.cache
 /.python-version
 /Mathics3.egg-info
 ChangeLog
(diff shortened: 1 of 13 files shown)


rocky:
@mmatera note that we currently have some brokenness in the GS tests.

rocky:
Superceded by #1018

PR #1018, opened 2020-11-22
Reinstate CombinatoricaLite

rocky:
A way to add back in Combinatorica (as CombinatoricaLite) incrementally as a package.

@mmatera you probably suggested autoload packages sometime in the past. I am sorry I didn't understand what you meant then. 

One thing about the autoload mechansim here that could be improved is not forcing the autoload initially. 

It might be nice to have add a `AutoLoad[]` function works like `LoadModule` to do the actually pulling in. Even slicker (which is how GNU Emacs's autoload works) is to list a number of functions which when called will automatically issue the `AutoLoad[]` function. That way loading is fact, but you get what you want on demand. 

This kind of method is what I was getting at in https://github.com/mathics/Mathics/pull/995#discussion_r528219485 . However `BinarySearch[]` in Mathics doesn't currently work.


mmatera:
> A way to add back in Combinatorica (as CombinatoricaSmall) incrementally as a autoload package.
> 
> @mmatera you probably suggested autoload packages sometime in the past. I am sorry I didn't understand what you meant then.
> 
> One thing about the autoload mechansim here that could be improved is not forcing the autoload initially.
> 
> It might be nice to have add a `AutoLoad[]` function works like `LoadModule` to do the actually pulling in. Even slicker (which is how GNU Emacs's autoload works) is to list a number of functions which when called will automatically issue the `AutoLoad[]` function. That way loading is fact, but you get what you want on demand.
> 
> This kind of method is what I was getting at in [#995 (comment)](https://github.com/mathics/Mathics/pull/995#discussion_r528219485) . However `BinarySearch[]` in Mathics doesn't currently work.

Actually, if you do not want to load the package, you can just load from a session by calling 
```Needs["Combinatorica`"]```
This is the standard way to load native packages in WL.  We can put in the same folder where Calculus.m.

File: mathics/builtin/scoping.py [hunk 1]
```diff
@@ -413,8 +411,7 @@ class Contexts(Builtin):
 
     ## this assignment makes sure that a definition in Global` exists
     >> x = 5;
-    >> Contexts[] // InputForm
-     = {"CombinatoricaOld`", "Global`", "ImportExport`", "Internal`", "System`", "System`Convert`B64Dump`", "System`Convert`Image`", "System`Convert`JSONDump`", "System`Convert`TableDump`", "System`Convert`TextDump`", "System`Private`", "XML`", "XML`Parser`"}
+    X> Contexts[] // InputForm
```
mmatera, line 414: And what would be the output of the test?

mmatera (approved):
I am OK in general with the changes. There are a few doctests that maybe we would like to keep, but we can put back later.  Apart from that, it would be great if we can put Combinatorica back. If this pass the tests, let's do the merge.

mathics/builtin/scoping.py
rocky: My understanding [from the wiki](https://github.com/mathics/Mathics/wiki/Documentation-markup) is that `X>` just puts in something as an example. 

I don't want to specify output since that is fragile and changing over time. 

If my understanding or the wiki is wrong please suggest a change or better just fix. Thanks.

mmatera: Sorry, you are right. I just forgot it. Thanks!

rocky:
> Actually, if you do not want to load the package, you can just load from a session by calling
> `` Needs["Combinatorica`"] ``
> This is the standard way to load native packages in WL. We can put in the same folder where VectorAnalysis.m is.

I didn't know this, but this is a great Idea! Is done now in c52ce9b8 

That said, this package is very weak as it stands now is pretty weak. 

Fixing up BinarySearch would get us back to what was around in 1.0.

mmatera (approved):

rocky:
> I am OK in general with the changes. There are a few doctests that maybe we would like to keep, but we can put back later. 

The entire doc system needs a major overhaul in terms of:

* how it is tagged (note that packages don't have that nice documentation that comes with the builtins) which leads to 
* how it is implemented in core and in the front-ends, and finally
* the content itself which will also tie into tagging 

In 1.1 the documentation was at a local maximum. A lot of stuff needs to be broken down, rethought and redone.

But think if this sort of like the post 1.0 and pre 1.1 days where there were graph package (done lamely), the natlang which came  built-in but was a nightmare to build in CI, and where we had little bit of combinatorica rolled in. And maybe at some point there was Jupyter integration elsewhere. 

In terms of *features*,  we slid back when we went into 1.1. But post 1.1, having faith that we'd keep going we are now just getting to get all of this back in. But in a much better and sustainable way.

rocky:
@mathics/maintainers  I am thinging that maybe CombinatoricaLight (or CombinatoricaLite) is a better name than CombinatoricaSmall?

mmatera:
I go for CombinatoricaLite

Issue #1014, opened 2020-11-17
Are we handling `from_python(bool)` and `is_true(from_python(bool))` correctly?


rocky:
I think right now converting a boolean creates a 0 or 1 integer. Is this want we want to do? 

Ideally  for a boolean _x_ `from_bool(x).is_true() == x` 

WL does have True and False

rocky:
There is a broader point that I've been meaning to mention ever since on one of the google groups we glibly suggested that one would be able magically run `to_python()`, `from_python()`, or `to_sympy()` and magically that would work. 

When I look at the existing types of conversions, things are a bit lacking and probalby there is an easy set of things to do to make this work better. Working though specific examples (such as trying to do this in mathicsscript for example) has and probalby will get us closer to supporting the easy stuff, and will probably give us a better understanding of how to guide others to use Mathics as a Python library. (This was one of our bullet points in the Sage-Day demo).

However I know that to do a good job in any of these functions would take a signifcant amount of work, and a significant amount of explaining the subtleties.

The good news is that we've started on this journey, and I think the journey is worth it.

 The bad news is that it is still a long journey.

mmatera:
Let's start fixing it with PR #1020

rocky:
Fixed by #1020. 

The broader issue of handling to_python() remains a research project and probably not something to track in a issue without breaking this up into small pieces like this one (handle boolean) .

Of course adding more tests and round-trip tests is a worthwhile endeaver too, but that's also probably a separate issue.

PR #1020, opened 2020-11-22
adding support for bool in from_python

mmatera:
This would be a fix for issue #1014

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -89,6 +89,11 @@ def from_python(arg):
     number_type = get_type(arg)
     if arg is None:
         return Symbol('Null')
+    if isinstance(arg, bool):
+        if arg:
+            return Symbol("System`True")
+        else:
+            return Symbol("System`False")
     if isinstance(arg, int) or number_type == 'z':
         return Integer(arg)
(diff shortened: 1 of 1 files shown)


mathics/core/expression.py
```diff
@@ -89,6 +89,11 @@ def from_python(arg):
     number_type = get_type(arg)
     if arg is None:
         return Symbol('Null')
+    if isinstance(arg, bool):
```
rocky, line 92: This is exactly what I was thinking too - thanks.

GarkGarcia:
@mmatera @rocky I guess we're ready to merge this?

rocky:
I think so - @mmatera ? I have in the back of my mind to add  a round-trip test for all types. But that can come later.

mmatera:
I agree: it would be nice to add some examples, but we can do it later. Let's merge this as it is if it is OK for you

rocky:
Fixed by #1020. The broader issue of handling `to_python()` remains a research project and probably not something to track in a issue without breaking this up into small pieces like this one (handle boolean) .

Of course adding more tests and round-trip tests is a worthwhile endeaver too, but that's also probably a separate issue.

PR #1025, opened 2020-11-22
Add Catch/Throw and With

mmatera:


rocky:
Thanks for doing this. I would like to look at in detail possibly late the upcoming week.

rocky (approved):
Sorry for the delay. Finally got around at looking at all of it.  I like `With`.

LGTM although as you say `Catch` could be improved.

Issue #1024, opened 2020-11-22
Compiling Expression Modules? 


rocky:
In Combinatorica 2.0 there are a few compiled modules, and none of them can get compiled. 

Here is an example: 

```
RP = Compile[{{n, _Integer}},
             Module[{p = Range[n],i,x,t},
	            Do [x = Random[Integer,{1,i}];
		        t = p[[i]]; p[[i]] = p[[x]]; p[[x]] = t,
		        {i,n,2,-1}
		    ];
	            p
	     ]
        ]
```

The error message we get back is: 

```
Compile::comperr: Expression Module[{p = Range[n], i, x, t}, Do[x = Random[Integer, {1, i}] ; t = p[[i]] ; p[[i]] = p[[x]] ; p[[x]] = t, {i, n, 2, -1}] ; p] could not be compiled.
```

A "could not be compiled" message occurs for all of them, so I suspect this is more about Expression Modules than it is about the construct inside that needs compiling. 

By the way, what would be a workaround for this?

mmatera:
Maybe when the expression can not be compiled, Compile should return a pure `Function`

rocky:
> Maybe when the expression can not be compiled, Compile should return a pure `Function`

Ok - thanks for doing.

GarkGarcia:
I'm not sure how compiling works in Mathics, but why can't a module be compiled?

rocky:
There is definitely something we can do better here. I just don't think anyone has yet, investigated doing this. 

Using `Function` is just workaround for compatibility now so we can pull in more external packages. 

What "Compile" means is a bit weird. Whatever is done in Mathematica probably doesn't work here since this is not Objective C based. We could simply bytecompile the function using Python's `compile()` function. Or some involved thing with LLVM, or somehow hook into Cython. And there are probably other possibilities.

mmatera:
Indeed, the current implementation is based on llvm. We could change it, or define a pymathics that overwrite the default behavior. The problem is by now that there are certain structures that are not implemented. This was just the simpler workaround I could think when the issue comes up

rocky:
Compiled functions have been improved. The example still fails, but for another reason. Will open an issue for that specifically when it has been narrowed.

PR #1026, opened 2020-11-22
fix compile

mmatera:
@rocky  this should fix issue #1024, until a better solution comes up


mmatera:
I merged this, but I am going to keep the branch in case we figure out how to implement a true Compile for Module.

PR #1027, opened 2020-11-25
Small tweaks, more tests

rocky:
"Catch[] needs to make sure it controls the expression evaluation
rather than having its expression get evaluated before it has a
change to wrap it in a `try`.

Internally, a plain `raise` is a bit more tidy Python wise
since the Python traceback will the path from raise to handler without
intervening tests of inappropriate exception handlers. This is a
matter of taste though.

Add more unit test tests.

### DIFF
--- a/mathics/builtin/control.py
+++ b/mathics/builtin/control.py
@@ -751,6 +751,7 @@ class Catch(Builtin):
      = 24
 
     """
+    attributes = ("HoldAll",)
 
     def apply1(self, expr, evaluation):
         'Catch[expr_]'
@@ -759,9 +760,9 @@ class Catch(Builtin):
         except WLThrowInterrupt as e:
             return e.value
         return ret
- 
+
     def apply3(self, expr, form, f, evaluation):
-        'Catch[expr_, form_, f__:Identity]'
+        "Catch[expr_, form_, f__:Identity]"
         try:
             ret = expr.evaluate(evaluation)
         except WLThrowInterrupt as e:
@@ -771,7 +772,9 @@ class Catch(Builtin):
             if match.is_true():
                 return Expression(f, e.value)
             else:
-                raise e
+                # A plain raise hide, this path and preserves the traceback
+                # of the call that was originally given.
+                raise
         return ret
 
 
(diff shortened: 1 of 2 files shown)


rocky:
It looks like there are still some bugs: 

```
Throw need not occur lexically inside Catch:

In[1]:=	f[x_] := (If[x < 0, Throw["negative"]]; Sqrt[x])	https://wolfram.com/xid/0y76co-bdikhw
In[2]:=	Catch[Sum[f[i], {i, 5, -5, -1}]]	https://wolfram.com/xid/0y76co-npb522
Out[2]=	negative (* comes out 0 instead *)
```

I'll look at this more later.

rocky:
> It looks like there are still some bugs:
> 
> ```
> Throw need not occur lexically inside Catch:
> 
> In[1]:=	f[x_] := (If[x < 0, Throw["negative"]]; Sqrt[x])	
> In[2]:=	Catch[Sum[f[i], {i, 5, -5, -1}]]	
> Out[2]=	negative (* comes out 0 instead *)
> ```
> 
> I'll look at this more later.

mathics/builtin/control.py
```diff
@@ -759,9 +760,8 @@ def apply1(self, expr, evaluation):
         except WLThrowInterrupt as e:
             return e.value
         return ret
- 
+
     def apply3(self, expr, form, f, evaluation):
-        'Catch[expr_, form_, f__:Identity]'
```
mmatera, line 764: Why did you remove this line? This was for stsblish the default handler function to the identity. Shouldn't it be there?

mathics/builtin/control.py
```diff
@@ -771,7 +771,9 @@ def apply3(self, expr, form, f, evaluation):
             if match.is_true():
                 return Expression(f, e.value)
             else:
-                raise e
+                # A plain raise hide, this path and preserves the traceback
+                # of the call that was originally given.
+                raise
```
mmatera, line 776: OK, I es not aware of how this idiom works

mmatera (approved):

mathics/builtin/control.py
rocky: Ooops - definitely a mistake.  It should be back in d0be8daf   Thanks for catching.

mmatera: OK. Then, let's merge this. Over the weekend I can try to figure out what happens with the throw inside called functions

PR #1028, opened 2020-11-26
More tests, More doc, less duplication

rocky:


test/test_evaluation.py
```diff
@@ -46,7 +46,7 @@
         ),
     ],
 )
-def check_evaluation(str_expr: str, str_expected: str, message=""):
```
rocky, line 49: This was a plain bug I introduced a while a go. The name has to start with `test_` in order for `py.test` to notice it is a test case.

PR #1030, opened 2020-11-26
Add Trace option to Get

rocky:
`Get["fn", Trace->True]` will show lines as they are read.

This isn't ideal, but it is a start. I would also like to do the same for Import.


GarkGarcia (approved):

mmatera (approved):
Great Idea!

PR #1032, opened 2020-11-27
Add contributed package RSolve

rocky:


mmatera (approved):
Does this package work as it is? Very nice!

GarkGarcia (approved):

rocky:
> Does this package work as it is? Very nice!

It is hard to tell because there don't seem to be any tests listed. Testing doesn't seem to be part of the Wolfram Language Package culture. 

So I imagine some things work and some things don't work. GeneratingFunction and SeriesTerm (now called SeriesCoefficient) do not work, but don't error either. I think at least one RSolve example that isn't builtin works. 

But given the choice of leave out or go with and have need drive improvements, I'd prefer the latter.

mmatera:
> > Does this package work as it is? Very nice!
> 
> It is hard to tell because there don't seem to be any tests listed. Testing doesn't seem to be part of the Wolfram Language Package culture.
> 
> So I imagine some things work and some things don't work. GeneratingFunction and SeriesTerm (now called SeriesCoefficient) do not work, but don't error either. I think at least one RSolve example that isn't builtin works.
> 
> But given the choice of leave out or go with and have need drive improvements, I'd prefer the latter.

I think the same, I just was curious. Let's merge this, then.

PR #1038, opened 2020-12-08
Tuples[expr_, n_] -> Tuples[expr_, n_Integer]...

rocky:
since that's what we currently handle.

In the future we should expand to Tuples[{list1, list2, ...}]

Some white space and blackening done too.


rocky:
Needs revising. Will do sometime after the 1.1.1 release.

PR #1044, opened 2020-12-10
Symbol("True') and Symbol("False')

rocky:
There were numerous complicated ways we were tesing for Symbol("True") and Symbol("False"). 

This simplfies the code. Some more files are blackened in the process.

After this is done, probalby Symbol("Null") is next.


rocky:
Fixes #1044

PR #1060, opened 2020-12-24
Improve to python

rocky:


rocky:
Code seems to be incorporated already.

PR #1066, opened 2020-12-25
Convert the wiki to Sphinx-docs

GarkGarcia:
@rocky Sorry for taking so long for this. I've been very busy this semester. This is a follow up to https://github.com/mathics/Mathics/pull/999#issuecomment-735493279. I've written a guide on how to write builtins and I've also converted multiple articles from the wiki to RST and Sphinx.

Hope you're all doing well and mary christmas! I plan on giving Mathics more attention in the following months.


rocky:
@GarkGarcia I understand "very busy" very well, so of course it is expected that at such times you may drop off here. 

It is good to hear from you, whenever you are free. 

Wow - there is a lot here. There is also a lot that could be better organized and expanded. 

I'll probably make a few changes and put this under read the docs.

rocky:
@GarkGarcia @mmatera 

Recently I have been moving things _out_ of Mathics. Most recently is the docker image creation. Here is why I think that is better outside. 

*I have noticed that basically *after* a mathics release I have to then update the Docker file, but the release has already been released. 
* The docker image really pulls togther: 
   * Mathics3
   * Mathics-Django 
   * Mathicsscript
   * pymathics-natlang
   * pymathics-graph
   * iwolfram (in the future perhaps)
   * matplotlib backend (in the future perhaps)
* The docker file is not getting large and complicated. And every change that is commited we are running a 13-minute test of stuff that has nothing to do with docker. On the other hand docker should have its own tests and those could take a long time too.

Something similar is happening here. Every commit wi'll have a test of Mathics core which isn't needed. And when when Mathics is built we run the sphinx developer tests which are probably independent. 

Therefore I think we might start splitting this into its own directory on Mathics3. Thoughts? 

Also, over the weekend I was planning on committing #999 

Your thoughts?

GarkGarcia:
> Therefore I think we might start splitting this into its own directory on Mathics3. Thoughts?

Yeah, makes sense. We don't want to run all tests before we commit a change to the developer documentation. Having this in a separate repository seems a bit weird to me, but we can test it and see if it works.

> Also, over the weekend I was planning on committing #999
> 
> Your thoughts?

I've seen you've merged it yesterday. I think that's great!

@rocky I've only now realized that there is already some Sphinx documentation out there (https://mathics-development-guide.readthedocs.io/en/latest/#). I believe this is stored at https://github.com/Mathics3/mathics-development-guide, am I right? I'll convert this to a PR on https://github.com/Mathics3/mathics-development-guide.

> It is good to hear from you, whenever you are free.

It's good to hear from you guys too 😁️

GarkGarcia:
> @rocky I've only now realized that there is already some Sphinx documentation out there (https://mathics-development-guide.readthedocs.io/en/latest/#). I believe this is stored at https://github.com/Mathics3/mathics-development-guide, am I right? I'll convert this to a PR on https://github.com/Mathics3/mathics-development-guide.

Ohh, I see, you've already done this for me. Great work!

GarkGarcia:
@rocky I'll go over the wiki and see if there's anything missing in [`mathics-development-guide`](https://github.com/Mathics3/mathics-development-guide). We should probably remove the wiki after we finish this.

rocky:
> We should probably remove the wiki after we finish this.

@GarkGarcia I added a comment to this effect in https://github.com/mathics/Mathics/wiki . However yes, when things settle down we should. I haven't figured out how to solicit changes in readthedocs. I know some docs have a "edit me on github" links.

PR #1068, opened 2020-12-26
WIP Add Show[ ]

rocky:


GarkGarcia (approved):

wuyudi:
Will it be better to use f-string instead of `%` in formatting string? (Maybe just taste?)

rocky:
@wuyudi Sure - if you would like to do this please do so. 

This code predates the use of f-strings. Now that the code is 3.6 or better, we can start using f-strings. 

The thing that bothers me more about this though is that it is not processing the Graphics options like it should. 

Therefore we see stuff like this: 

Mathics-Show-Bug

rocky:
> The thing that bothers me more about this though is that it is not processing the Graphics options like it should.
> 
> Therefore we see stuff like this:
> 
> Mathics-Show-Bug

This was fixed in commit 392be84. I also committed to master directly inadvertently, but since this is beneficial I don't see any harm.

PR #1079, opened 2020-12-29
Start Insert[]. Fix Do[] for negative increment

rocky:
These bug fixes were a result of expanding Combinatorica example testing to page 28.

### DIFF
--- a/mathics/builtin/lists.py
+++ b/mathics/builtin/lists.py
@@ -2128,7 +2128,6 @@ class Range(Builtin):
             index += di
         return Expression('List', *result)
 
-
 class _IterationFunction(Builtin):
     """
     >> Sum[k, {k, Range[5]}]
@@ -2242,8 +2241,9 @@ class _IterationFunction(Builtin):
         di = di.evaluate(evaluation)
 
         result = []
+        compare_type = "GreaterEqual" if Expression('Less', di, Integer(0)).evaluate(evaluation).to_python() else "LessEqual"
         while True:
-            cont = Expression('LessEqual', index, imax).evaluate(evaluation)
+            cont = Expression(compare_type, index, imax).evaluate(evaluation)
             if cont == Symbol('False'):
                 break
             if not cont.is_true():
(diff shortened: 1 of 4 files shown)


mathics/builtin/lists.py
```diff
@@ -2548,11 +2548,35 @@ def parts():
...
+    >> Insert[{a,b,c,d,e}, x, 3]
+     = {a, b, x, c, d, e}
+
+    >> Insert[{a,b,c,d,e}, x, -2]
+     = {a, b, c, d, x, e}
+    """
+    def apply(self, expr, elem, n, evaluation):
+        'Insert[expr_List, elem_, n_Integer]'
+
+        py_n = n.to_python()
```
GarkGarcia, line 2567: We could probably use `n.get_int_value()` in here (it's a bit safer than `to_python`, but it isn't strictly necessary because `n` is guaranteed to be an integer).

test/helper.py
```diff
@@ -11,7 +12,9 @@ def check_evaluation(str_expr: str, str_expected: str, message=""):
     result = session.evaluate(str_expr)
     expected = session.evaluate(str_expected)
 
+    print(time.asctime())
```
GarkGarcia, line 15: Are these for debugging?

rocky: Yes, it is for seeing where the time is getting spent. There are probably more elaborate ways. If you want to do something more sophisticated with elapsed time, please go ahead.

This test is starting to take a long time and will grow a lot more. 

Since py.test doesn't show this output unless you use `-s` my thought rigtht now is to leave all of this in since wanting to know where the time is spent will be an on-going question. 

Also in the back of my mind is to have some flag to separate quick from long-running tests and this could be something could be done here with verbose logging. 

Again, if this is stuff you feel strongly about feel free to adjust the code.

mathics/builtin/lists.py
rocky: I thiink we should be using `to_python()` for a number of reasons:

- it requires less thought to get it right. (How many times in PR's have I seen the comment that there is `get_int_value()` instead of ... this is also an indicator that something is not simple )
- it is more robust when thing change.

rocky:
Just got word from Steve Skiena that he has tweeted about this. Since this greatly improves Combinatorica coverage, I am merging  and will put out another docker image soon.

PR #1080, opened 2020-12-29
Implementing TimeConstrained and TimeRemaining

### LINKED ISSUE #1059: Implement TimeConstrained
See https://reference.wolfram.com/language/ref/TimeConstrained.html

This is needed in Rubi.

mmatera:
Fixes #1059

### DIFF
--- a/mathics/core/evaluation.py
+++ b/mathics/core/evaluation.py
@@ -114,6 +114,19 @@ def run_with_timeout_and_stack(request, timeout, evaluation):
     thread = Thread(target=_thread_target, args=(request, queue))
     thread.start()
 
+    # Thead join(timeout) can leave zombie threads (we are the parent)
+    # when a time out occurs, but the thread hasn't terminated.  See
+    # https://docs.python.org/3/library/multiprocessing.shared_memory.html
+    # for a detailed discussion of this.
+    #
+    # To reduce this problem, we make use of specific properties of
+    # the Mathics evaluator: if we set "evaluation.stopped", and then
+    # at various points of evaluation check_stopped() is called an the
+    # thread will raise a TimeoutInterrupt.
+    #
+    # However this still will not terminate long-running processes
+    # in Sympy or or libraries called by Mathics that might hang or run
+    # for a long time.
     thread.join(timeout)
     if thread.is_alive():
         evaluation.timeout = True
--- a/mathics/core/evaluation.py
+++ b/mathics/core/evaluation.py
@@ -11,7 +11,7 @@ from threading import Thread, stack_size as set_thread_stack_size
 from typing import Tuple
 
 from mathics import settings
-from mathics.core.expression import ensure_context, KeyComparable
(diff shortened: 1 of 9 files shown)


mmatera:
I get an error in the github workflow, but it seems to be related to the particular test I did: maybe I hit an issue in sympy.

rocky:
> I get an error in the github workflow, but it seems to be related to the particular test I did: maybe I hit an issue in sympy.

Will look at this soon to see if I can make headway.

rocky:
Based on the latest results, I now _do_ think there is a bug here with some interaction between the new code and the thread started. 

I have a branch with the thread code redone as multiprocessing, however that doesn't work right now. The queue needs to be shared between the two processes. (In python 3.8 there is https://docs.python.org/3/library/multiprocessing.shared_memory.html)

I've spent more time than I can afford on this right now. So it will have to wait. 

I'll commit in yet another branch the multiprocessing version of this...

rocky (commented):
How does this kill the remaining threads like the comment suggests? 

Also how does this deal with the concern in https://docs.python.org/3/library/multiprocessing.shared_memory.html)?

mmatera:
According to the material that you passed me, it seems that the best fitting solution would be to use threads.  My idea about overcoming the threat of zombie threads would be to use the fact that each time the mathics interpreter makes a step, it checks if the evaluation was stopped. Then, we can use this to finish the zombie threads.

rocky:
> According to the material that you passed me, it seems that the best fitting solution would be to use threads. My idea about overcoming the threat of zombie threads would be to use the fact that each time the mathics interpreter makes a step, it checks if the evaluation was stopped. Then, we can use this to finish the zombie threads.

Good! This makes a lot of sense - thanks!

We should probably but the above information and maybe a link to the problem as comments in the code.

mmatera:
I think now this is ready to merge. A remaining problem is that certain tasks (like dsolve, and fullsimplify) from sympy can not be stopped with this mechanism, but then we have two alternatives: 1) ask for implementing timeouts in sympy and 2) reimplement certain time-consuming routines in sympy inside mathics.

mathics/core/evaluation.py
```diff
@@ -116,6 +116,11 @@ def run_with_timeout_and_stack(request, timeout):
 
     thread.join(timeout)
     if thread.is_alive():
```
rocky, line 118: What is going on here is not obvious: that we are doing setting these variables because the behavior of Mathics interpretation looks for those specifically at certain points and uses them to abort execution. And the part about how this fails for long-running sympy (or numpy, etc) operations. 

And the link to the place that describes why a simple `join(timeout)` is not sufficient. 

I'd be happy to make a stab at this if you'd like.

rocky:
Ok - thanks for doing this done. I guess we should come back to figure out how to deal with long sympy calculations later. 

The only thing I'd personally like to see is to add some comment describing what's up the code around the `join(timeout)`.

PR #1086, opened 2020-12-31
Start a settings file...

rocky:
Put in it MathicsVersion usage
Add TraceGet for tracing Get
definition.py: Sets MathicsVersion

We could probably put $UseSansSerif in here as well.

### DIFF
--- a/mathics/autoload/settings.m
+++ b/mathics/autoload/settings.m
@@ -1,0 +1,10 @@
+(*****************************)
+(* Settings for Mathics Core *)
+(*****************************)
+
+Settings`$TraceGet::usage = "If this Boolean variable is set True, 'Get' traces the lines it reads that start a new expression";
+Settings`$TraceGet = False
+Unprotect[Settings`$TraceGet]
+
+System`MathicsVersion::usage = "This string is the version of Mathics we are running."
+(* System`MathicsVersion = "xxx" - set inside Mathics *)
--- a/mathics/builtin/files.py
+++ b/mathics/builtin/files.py
@@ -2156,23 +2156,24 @@ class Get(PrefixOperator):
         "Trace": "False",
     }
 
-    def check_options(self, options):
-        # Options
-        # TODO Proper error messages
-
-        result = {}
-        if options["System`Trace"].to_python():
-            result["TraceFn"] = print
-        else:
-            result["TraceFn"] = None
-
-        return result
-
     def apply(self, path, evaluation, options):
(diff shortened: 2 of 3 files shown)


mathics/autoload/settings.m
```diff
@@ -0,0 +1,10 @@
+(*****************************)
+(* Settings for Mathics Core *)
+(*****************************)
+
+Settings`$TraceGet::usage = "If this Boolean variable is set True, 'Get' traces the lines it reads that start a new expression";
+Settings`$TraceGet = False
+Unprotect[Settings`$TraceGet]
```
mmatera, line 7: Shouldn't be before the assignment?

mmatera (approved):

mathics/autoload/settings.m
rocky: When this ends, I TraceFout should not be protected. 

There is a bug somewhere where Mathics protects a lot of symbos it shoudn't. I see this a lot when tryimng to import Mathics pacages with ``Get`` or ``Needs``. The thing is that I just haven't been able to understand what exactly is wrong.

mmatera: The thing is: you set the value of the variable and, after that, you Unprotect it. If in the line 6 you can assign, then in line 7 you are Unprotect`ing a variable which was Unprotect'ed already.

By default, all the Builtins are protected, but I remember you had proposed to implement something like an "Unprotected" property to avoid that behavior in some specific cases.

mmatera: By the way, to allow to trace imports is very useful! Thanks!

rocky: Thanks. There is a  _lot_ more we can do in the way of making things easier to debug. 

I'd like to follow the WL `Trace` functions, but since I don't have Mathematica installed anywhere, I am not sure how to follow that closely.  

I suspect some sorts of debuggers can and are written purely in WL, but right now we don't have a `Dialog[]` function.

GarkGarcia (approved):

mathics/autoload/settings.m
rocky: > By default, all the Builtins are protected, but I remember you had proposed to implement something like an "Unprotected" property to avoid that behavior in some specific cases.

Yes, that has been done. You add "Unprotected" to the set in the class variable "attributes". But that only works if there is a Python class, which isn't the case here. Perhaps we should add it and then remove this from the settings file?

PR #1088, opened 2020-12-31
implementing functions from WMA 10

mmatera:
I just borrow the implementation of `FirstCase`, `Key`, `Lookup` and `Failure`  from CellsToTeX.


rocky (approved):
LGTM  - Thanks for doing this

I can't wait to start being able to use tooltips and the other new options (which I guess are not implemented). 

It is also nice to have more version 10 features included.

PR #1095, opened 2021-01-03
Add from_python option and bug fix

rocky:
Add Mathics option OmitSurroundingQuotes and omit_surrounding_quotes in from_python to indicate whether we want to add surrounding quotes when converting from python.

Also we had a bug in formatting from a python string to Mathics String. Previously we slapped a double quote (") around the outside. However that assumes internally there isn't a double quote.

Instead we now use Python's repr() which handles this.

However it can produce a string that starts out with a single quote ('). But we shouldn't explect the output of PythonForm to be something that can be then input into Mathics.

### DIFF
--- a/mathics/builtin/inout.py
+++ b/mathics/builtin/inout.py
@@ -670,7 +670,7 @@ class InterpretationBox(Builtin):
     """
     <dl>
     <dt>'InterpretationBox[{...}]'
-        <dd> is a low-level box construct that displays as 
+        <dd> is a low-level box construct that displays as
     boxes but is interpreted on input as expr.
     </dl>
     """
@@ -706,7 +706,7 @@ class TagBox(Builtin):
     """
     <dl>
     <dt>'TagBox[boxes, tag]'
-        <dd> is a low-level box construct that displays as 
+        <dd> is a low-level box construct that displays as
     boxes but is interpreted on input as expr
     </dl>
     """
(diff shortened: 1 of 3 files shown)


GarkGarcia (approved):

GarkGarcia:
> Also we had a bug in formatting from a python string to Mathics String. Previously we slapped a double quote (") around the outside. However that assumes internally there isn't a double quote.
> 
> Instead we now use Python's repr() which handles this.

That's quite interesting. Using `repr` is an improvement, but I feel like we should look into how are special characters supposed to be quoted according to the WL specification. Are [named characters](https://reference.wolfram.com/language/guide/ListingOfNamedCharacters.html) used for this? If so, it feels like the tables from https://github.com/Mathics3/mathics-development-guide/pull/6 could be helpful in here.

Anyway, I wouldn't worry about this too much in this PR, we can address this issue later.

rocky:
The use of repr is to take some string in WL and turn it into a string in Python. Generally the meaning of alll of those special symbols as operators and such doesn't mean anything inside a string, or I wouldn't know or pretent to know what the intent is here when some Unicode character is used inside a string. I think I'd need to see a real example of the use in order to figure what the right thing to do would be. 

In any case, there is a problem with this, probalby not related to that repr and I didn't have time and won't have time to investigate.

If someone wants to just pull out the repr part and make a new PR with that, please go ahead!

rocky:
Needs to be rethought.

PR #1101, opened 2021-01-08
Fix bug where Mathics failed to sort a list o eigenvalues

GarkGarcia:
You can reproduce the bug by trying to evaluate `{{-8, 12, 4}, {12, -20, 0}, {4, 0, -2}} // Eigenvalues`. Mathics crashes because it tries to sort the results but it fails for some reason (in this case I think the values are so complicated that Mathics fails to compare them effectively).

This PR fixes this issue. By the way, for some reason SymPy is returning the wrong answer in this case: `{{-8, 12, 4}, {12, -20, 0}, {4, 0, -2}}` is a [Hermitan matrix](https://en.wikipedia.org/wiki/Hermitian_matrix), so it's eigenvalues should be real numbers (but SymPy returns complex numbers with non-zero imaginary parts).


rocky:
Seems interesting. I'll try to look at over the weekend.

rocky:
I have looked over to understand this better. The bug fixes are fine.

As to the results, what could be done here is to use [numpy's routines instead](https://numpy.org/doc/stable/reference/generated/numpy.linalg.eig.html) -- after checking that the values are all real and. 

Here is what you'd get: 

```
>> import numpy as np
>> xx = np.array([[-8.0, 12.0, 4.0],[12.0, -20.0, 0.0], [4.0, 0.0, -2.0]], ndmin=2)
>> from numpy import linalg as LA
>>> w, v = LA.eig(xx)
>>> v
array([[ 0.53280038, -0.67737661, -0.50723238],
       [-0.84213356, -0.36542763, -0.3965775 ],
       [-0.08327559, -0.63845405,  0.76514156]])
```

rocky (approved):
LGTM

PR #1103, opened 2021-01-09
Celltotexfixes rebase

rocky:
I had or am having a very hard time rebasing to master.

This might be close though. If you can work from here and add what's missing we might be close.


rocky:
Sigh. This just doesn't work. 

I think we should just close the not only this PR but the other one too.

PR #1106, opened 2021-01-10
mpmath fns and more...

rocky:
This probably should be spit up.

The main part of it adds more mpmath constants and DRY's code
In particular

* the EulerGamma constant was added.
* Classes MPMathConstant and NumpyMathConstant were added
* We start to keep track of dictionaries for mapping between names
  automatically now. More will be added later
* `get_constant()` added. When looking up a constant we can decide
  whether to evaluate the value or leave in symbolic form.

Misc things not so directly related to the above...

* test.py --sections now can take a list of sections separated with commas (and no space). For example:
```
python test.py --sections=Catalan,Pi,GoldenRatio
```

* super(<ClassName>, self) ->  super() a number of places
* admin testing uses 3.7.9 now
* add `admin-tools/time-mathmp-sympy-fns.py` program to time sympy vs. mpmath

### DIFF
--- a/admin-tools/pyenv-versions
+++ b/admin-tools/pyenv-versions
@@ -5,4 +5,4 @@ if [[ $0 == ${BASH_SOURCE[0]} ]] ; then
     echo "This script should be *sourced* rather than run directly through bash"
     exit 1
 fi
-export PYVERSIONS='3.6.12 3.7.9 3.8.6 3.9.0'
+export PYVERSIONS='3.6.12 3.7.9 3.8.7 3.9.1'
--- a/admin-tools/time-mathmp-sympy-fns.py
+++ b/admin-tools/time-mathmp-sympy-fns.py
@@ -1,0 +1,56 @@
+#!/usr/bin/env python
+"""
+Program to time mpmath pi vs sympy pi
+"""
+
+from timeit import timeit
+import mpmath
+import math
+import sympy
+
+PRECISION = 100
+mpmath.mp.dps = PRECISION
+ITERATIONS = 2000
+
+# print(mpmath.pi, "\n")
+
+def math_pi():
+    return math.pi
+
+
+def mpmath_pi():
+    return mpmath.pi
+
+
+def sympy_pi():
+    return sympy.pi.n(PRECISION)
+
+
+def mpmath_e():
+    return mpmath.e
+
+
+def sympy_e():
+    return sympy.E.n(PRECISION)
+
+
+def mpmath_degree():
+    return mpmath.degree
+
+
+def sympy_degree():
+    return (sympy.pi / 180).n(PRECISION)
+
+
+# print(timeit(math_pi, number=ITERATIONS))
+
+for sympy_fn, mpmath_fn, fn_name in (
+    (sympy_pi, mpmath_pi, "pi"),
+    (sympy_e, mpmath_e, "e"),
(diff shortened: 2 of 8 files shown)


GarkGarcia (approved):
LGTM

mmatera (approved):
Looks great!

rocky:
Ok. Let's do it then. 

Let me also explain that the motivation for starting some of this is as I try to explain how to interact with Mathics from Python. 

In doing that I see that this is another big area where things could be simpler and more powerful. 

There was a longer term goal I had mentioned of interoperability between Python, Sympy, mpmath, numpy in that you might want to start off in Mathics and convert to one of these. Or vice versa. 

One other thing that is worth mentioning is that if we can use mpmath functions more often there seems to be a huge win. 

What is dragging us down right now is that relations in trig functions have to be preserved. So the approximation for Degree has to match the approximation for Pi, Sin, Cos, and so on.  This is another area where I think there is some work to be done.

mmatera:
> Ok. Let's do it then.
> 
> Let me also explain that the motivation for starting some of this is as I try to explain how to interact with Mathics from Python.
> 
> In doing that I see that this is another big area where things could be simpler and more powerful.
> 
> There was a longer term goal I had mentioned of interoperability between Python, Sympy, mpmath, numpy in that you might want to start off in Mathics and convert to one of these. Or vice versa.
> 
> One other thing that is worth mentioning is that if we can use mpmath functions more often there seems to be a huge win.
> 
> What is dragging us down right now is that relations in trig functions have to be preserved. So the approximation for Degree has to match the approximation for Pi, Sin, Cos, and so on. This is another area where I think there is some work to be done.

Indeed, that's great. It will be also very important for example in linear algebra functions. Something that I was thinking is about the internal representation of arrays and graphics, for which ``Expression`` objects maybe are not the most efficient option.

rocky:
If something is pure data like arrays and graphics operations we can add `JSONFormat`. And then using that
Implement [ExpressionJSON](https://reference.wolfram.com/language/ref/format/ExpressionJSON.html).

I've been thinking about the xxxFormat, internally to_xxx, especially with respect to Python-like data structures. 

Here are my thoughts . The xxxFormat seems to be something you run in Mathics. The output type is something easily renderable by a front end. Often it is text but sometimes it can be graphics. 

The to_xxx routines are the underlying routines that create the formats. 

Is others understanding of what's  up here? 

Also, before I lose track, here are timings of mpmath vs sympy for 3 constant values: 

```
$ python ./time-mathmp-sympy-fns.py 
pi:
0.00028181797824800014 seconds for mpmath
0.03673283010721207 seconds for sympy

 3.141592653589793238462643383279502884197169399375105820974944592307816406286208998628034825342117068 
 3.141592653589793238462643383279502884197169399375105820974944592307816406286208998628034825342117068

e:
0.0002758570481091738 seconds for mpmath
0.039352256106212735 seconds for sympy

 2.718281828459045235360287471352662497757247093699959574966967627724076630353547594571382178525166427 
 2.718281828459045235360287471352662497757247093699959574966967627724076630353547594571382178525166427

degree:
0.0002791371662169695 seconds for mpmath
0.1811701429542154 seconds for sympy
```
 0.01745329251994329576923690768488612713442871888541725456097191440171009114603449443682241569634509482 
 0.01745329251994329576923690768488612713442871888541725456097191440171009114603449443682241569634509482

Issue #1107, opened 2021-01-11
Named Character don't really work


GarkGarcia:
After finishing the work at https://github.com/mathics/Mathics/pull/1077 and https://github.com/Mathics3/mathicsscript/pull/9 I went out and did a little test in mathicsscript:

```
Mathicscript: 1.1.2, Mathics 2.0.0dev
on CPython 3.6.9 (default, Oct  8 2020, 12:12:24)
using SymPy 1.7.1, mpmath 1.1.0

Copyright (C) 2011-2020 The Mathics Team.
This program comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under certain conditions.
See the documentation for the full license.

Quit by pressing CONTROL-D

In[1]:= \[Theta] = 4
Out[1]= θ == 4

In[2]:= a = b
Out[2]= a == b

In[3]:= "\[DifferentialD] is the differential"
Out[3]=  is the differential
```

There are a couple of issues in here. First of all, it looks like `=` is replaced by `` in mathicsscript. It looks like this is a mathicsscript-specific issue, but I haven't updated Mathics-Django since https://github.com/mathics/Mathics/pull/1077 was merged and I got the following when I tried updating it now:

```
python3 setup.py install
/home/pablo/.local/lib/python3.6/site-packages/setuptools/dist.py:452: UserWarning: Normalizing '2.0.0dev' to '2.0.0.dev0'
  warnings.warn(tmpl.format(**locals()))
running install
running bdist_egg
running egg_info
writing Mathics3.egg-info/PKG-INFO
writing dependency_links to Mathics3.egg-info/dependency_links.txt
writing entry points to Mathics3.egg-info/entry_points.txt
writing requirements to Mathics3.egg-info/requires.txt
writing top-level names to Mathics3.egg-info/top_level.txt
error: package directory 'mathics/algorithm' does not exist
Makefile:42: recipe for target 'install' failed
```

Anyway, this is probably a silly mistake in mathicsscript and it's probably my fault. I'll investigate this.

What's more concerning is the fact that event thought mathicsscript correctly convert the named characters to their corresponding unicode representations, the kernel doesn't seem to know what they are and they aren't properly display. According to [Wolfram's listing of named character](https://reference.wolfram.com/language/guide/ListingOfNamedCharacters.html):

[...]

PR #1108, opened 2021-01-11
Fix bug that converts `=` to `\uf7d9` in `replace_unicode_with_wl`

GarkGarcia:
This a fix for the first issue described in #1107. The issue was my fault, it's an error in the `UNICODE_TO_WL` dict.


GarkGarcia:
This should be followed by an update in https://github.com/Mathics3/mathics-development-guide/blob/master/resources/unicode-to-wl-conversion.csv.

PR #1114, opened 2021-01-16
Use Mathics Plus when we can...

### LINKED ISSUE #1112: Sum not expanding iteration variable inside its 1st argument function
In `Sum[`_f_, `{`  _i_ , _imax_ `}]` the (and the form where _imin_ is also given), the value of _i_ isn't getting expanded inside _f()_. 

Here is an example:

```
$ mathicsscript -e 'j, {j, 3}]'
6
$ mathicsscript -e 'Sum[Identity[j], {j, 3}]'
3 j
```

rocky:
Starts to fix #1112 but doesn't entirely fix the more general problem

### DIFF
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -1939,7 +1939,7 @@ class Sum(_IterationFunction, SympyFunction):
      = 0
 
     >> (-1 + a^n) Sum[a^(k n), {k, 0, m-1}] // Simplify
-     = -1 + (a ^ n) ^ m
+     = Piecewise[{{m (-1 + a ^ n), a ^ n == 1}, {-1 + (a ^ n) ^ m, True}}]
 
     Infinite sums:
     >> Sum[1 / 2 ^ i, {i, 1, Infinity}]
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -41,7 +41,7 @@ from mathics.core.expression import (
 from mathics.core.numbers import min_prec, dps, SpecialValueError
 
 from mathics.builtin.lists import _IterationFunction
-from mathics.core.convert import from_sympy
+from mathics.core.convert import from_sympy, SympyExpression
 
 class _MPMathFunction(SympyFunction):
 
@@ -1950,19 +1950,12 @@ class Sum(_IterationFunction, SympyFunction):
     #> a=Sum[x^k*Sum[y^l,{l,0,4}],{k,0,4}]]
      : "a=Sum[x^k*Sum[y^l,{l,0,4}],{k,0,4}]" cannot be followed by "]" (line 1 of "<test>").
 
-    ## Issue431
-    #> Sum[2^(-i), {i, 1, \\[Infinity]}]
-     = 1
-
-    ## Issue302
+    ## Issue #302
+    ## The sum should not converge since the first term is 1/0.
     #> Sum[i / Log[i], {i, 1, Infinity}]
      = Sum[i / Log[i], {i, 1, Infinity}]
     #> Sum[Cos[Pi i], {i, 1, Infinity}]
      = Sum[Cos[Pi i], {i, 1, Infinity}]
-
-    ## Combinatorica V0.9 issue in computing NumberofInvolutions
-    >> Sum[k!, {k, 0, Quotient[4, 2]}]
-     = 4
     """
 
     # Do not throw warning message for symbolic iteration bounds
@@ -1985,7 +1978,7 @@ class Sum(_IterationFunction, SympyFunction):
     def get_result(self, items):
         return Expression("Plus", *items)
 
-    def to_sympy(self, expr, **kwargs):
+    def to_sympy(self, expr, **kwargs) -> SympyExpression:
         """
         Perform summation via sympy.summation
         """
@@ -1993,7 +1986,10 @@ class Sum(_IterationFunction, SympyFunction):
             index = expr.leaves[1]
             arg_kwargs = kwargs.copy()
             arg_kwargs["convert_all_global_functions"] = True
-            arg = expr.leaves[0].to_sympy(**arg_kwargs)
+            f_sympy = expr.leaves[0].to_sympy(**arg_kwargs)
+            if f_sympy is None:
+                return
+
             evaluation = kwargs.get("evaluation", None)
 
             # Handle summation parameters: variable, min, max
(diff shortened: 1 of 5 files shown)


mathics/builtin/arithmetic.py
```diff
@@ -2003,14 +1999,33 @@ def to_sympy(self, expr, **kwargs):
...
+                # in its first argument, the function paramameter.
+                # For example in Sum[Identity[x], {x, 3}], sympy.summation can't
+                # does't evaluate Indentity[x].
+
+                if all(hasattr(i, "is_integer") and i.is_integer for i in bounds[1:]):
+                    # When we have integer bounds, it is better to not use Sympy but
+                    # use Mathics evaluation. We turn:
+                    # Sum[f[x], {<limits>}] into
+                    #   MathicsSum[Table[f[x], {<limits>}]]
+                    # where MathicsSum is self.get_result() our Iteration iterator.
```
mmatera, line 2021: Bmathics suu

rocky: I don't understand.

mathics/builtin/arithmetic.py
```diff
@@ -2003,14 +1999,33 @@ def to_sympy(self, expr, **kwargs):
             if evaluation:
                 # Min and max might be Mathics expressions. If so, evaluate them.
                 for i in (1, 2):
-                    expr = var_min_max[i]
+                    min_max_expr = var_min_max[i]
                     if not isinstance(expr, Symbol):
-                        expr = expr.evaluate(evaluation)
-                        value = expr.to_sympy(**kwargs)
+                        min_max_expr_eval = min_max_expr.evaluate(evaluation)
+                        value = min_max_expr_eval.to_sympy(**kwargs)
```
mmatera, line 2005: apart from the change in the name of the variables, is here something else?

mathics/builtin/arithmetic.py
mmatera: Sorry, I started writing in my cellphone, then the window got closed, and it seems that a random text was sent... :)

mmatera: What I was actually wondering is about what kind of expressions we should send to sympy. Should we tag expressions that have a sympy form, to decide if send them or not?

mathics/core/expression.py
```diff
@@ -88,23 +88,26 @@ def __str__(self) -> str:
 
 
 def from_python(arg):
+    """Converts a Python expression into a Mathics expression.
+
+    TODO: I think there are number of subtleties to be explained here.
+    In particlar the expression might have come form Sympy so we
```
mmatera, line 94: "In particular"?

mmatera (approved):

mathics/builtin/arithmetic.py
rocky: At one oint the mpmath and the sympy math were in the same function which made it _necessary_ to change the names; I needed the expr before evaluation and the min_max_expr before evaluation.

After they were split, in theory this didn't need to change. However, I think it a bad practice, which I hope we can stop doing ,to take  a variable of one type, e.g. Expression and convert it to another type e.g. SympyExpression as was happening in the function. I am not sure it was happening here, but it might have been.

mathics/builtin/arithmetic.py
rocky: It is a bit of a research problem 
* when something  _needs_ to be sent to sympy
* when it _could_ work either way, and
* when it needs or is better done using Mathics Plus

Right now I know that if the bounds are not integers, those need to be sent to sympy's Sum. 

Note that just because we are using Mathics' Plus, that doesn't mean that Sympy is not involved. The terms that Plus is summing could very easily be Sympy symbolc expressions can materialized on demand, like Pi.

So in sum (pun iintended), Sympy's Sum is needed and most useful when the value is some sort of sophisticated transformation that needs to look at the entire sequence of terms in order to make a decision and there are no Mathics variables that can be expanded. For the most part we shouldn't use Sympy Sum when Mathics Plus will do the trick. Mathics Plus is also probably faster.

Sympy Sum can't expand Mathics variables that have balues because that would mean calling back Mathics to get the value and there currently is no mechanism as far as I know for doing that. 

It is this aspect that has been the source of the bugs we've been seeing.

As was recently done for Eigenvalues, a "Method" option can be added to specify which approach to use, However when no opiton was supplied need to do the best we can.

mathics/core/expression.py
rocky: Typo fixed - thanks.

mathics/builtin/arithmetic.py
mmatera: My idea would be, add a method sympy_native():bool on the class Expression. Then, for constants (Pi, E), atoms, and numeric types we set this method to return True, and for elementary (sympy) functions, the method would return True if the arguments are all sympy_native(). Then, in operations like ``Sum``, and in functions like `to_sympy` you could check sympy_native and decide how to proceed.

rocky:
> What I was actually wondering is about what kind of expressions we should send to sympy. Should we tag expressions that have a sympy form, to decide if send them or not?

I am not sure if you mean here or both here and in general. In either case I don't know and suspect that many times this isn't going to be easy to answer if it can be answered at all.

Therefore some sort of hinting method like `Method` in WL has for Eigenvalue is one direction to explore.

mathics/builtin/arithmetic.py
rocky: This whole area is a mess. What is needed to think this out better. Maybe coming up with a number of specific examples would help.

I think what you want to know about an expression is whether it has unbound variables, which is the more general property that matters at least in Sum.

If an unbound variable, what kind? (Sympy variable, Module local varable? global variable?)

A bound value can still be symbolic as far as sympy, numpy, or mathmp are concerned. Pi and E are bound but symbolic. (In mathmp these  are not thought of a symbolc but rather arbitrary precision real). But these kinds of values can be materialized to some approximation to sort of Numeric value in Mathics. In general it is better to keep the value unmaterialized as long a possible, so that it can be used symbolicly in mpmath, Sympy, Numpy or Mathics.

mmatera: The thing is, we could have in mathics our own set of rules to deal with unbounded sums. The true advantage of using sympy is that the effort to keep the database of conversion rules for sums (and products) are not duplicated. But this makes sense for expressions that have a direct equivalent as pure sympy expressions.  So, I guess the right algorithm would be
* Check if the iterator is finite. In that case, return Plus/@Table[expr,it]
* otherwise, check if expr.to_sympy() gives a non ``None`` value. In that case, sent it to sympy
* If not, just let mathics to evaluate it with its internal rules. If there is no rule, then leave it unevaluated.

rocky: A lot of this and what you write before makes a lot of sense. 

Note though that: 

> * otherwise, check if expr.to_sympy() gives a non None value.

In general, this isn't good enough. Before the change expr.to_sympy() I believe was giving a non None value. The problem was however that SymPy sum had no way of materializing the iteration value of Sum. 

That said, since the step is tried second after the first succeeds, in practice will work, although it isn't bug free.

I've run out of time I can work on this in the short term. If you are up to it and want to continue please do!

GarkGarcia (approved):

mmatera (commented):
I think that sympy() should return a non None value just in those cases where all the symbols are native or can be interpreted as variables or formal functions. My question is if it is more efficient to look at whether an expression can be translated to sympy, or to do the transformation and see if it fails.

mathics/builtin/arithmetic.py
```diff
@@ -2003,14 +1999,33 @@ def to_sympy(self, expr, **kwargs):
...
+                # sympy.summation() won't be able to handle Mathics functions in
+                # in its first argument, the function paramameter.
+                # For example in Sum[Identity[x], {x, 3}], sympy.summation can't
+                # does't evaluate Indentity[x].
+
+                if all(hasattr(i, "is_integer") and i.is_integer for i in bounds[1:]):
+                    # When we have integer bounds, it is better to not use Sympy but
+                    # use Mathics evaluation. We turn:
+                    # Sum[f[x], {<limits>}] into
+                    #   MathicsSum[Table[f[x], {<limits>}]]
```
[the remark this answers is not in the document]
rocky: I'd prefer using the designated iterator method. That way Plus (which is the choice for now but might not always be) appears once.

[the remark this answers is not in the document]
mmatera: So, the idea is to introduce a `MathicsSum` symbol, as an alias of Plus@@?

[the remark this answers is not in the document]
rocky: Honestly, I think you two probably know more about this than I. 

My thoughts were from an engineering standpoint: 
this is some sort of iterator class, there is an iteration method, and in some particular situations the iteration method is preferred over Sympy's routine. 

Is this there a better way to implement this? I don't know. Is there a better way to express what is desired in implementation such as WL's Method option? (I don't know but I am leaning toward adding options Method options for "numpy", "mpmath", and "Sympy" - however you tell me what you think.)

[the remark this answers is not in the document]
rocky: BTW, one thing that has been in the back of my mind has been to updated the FIXME comment to better summarize and describe what we now understand. I won't have time for this until maybe the weekend, so you all want to update that or revise this please do.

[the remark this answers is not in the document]
mmatera: > My thoughts were from an engineering standpoint:
> this is some sort of iterator class, there is an iteration method, and in some particular situations the iteration method is preferred over Sympy's routine.

It is, except where there are closed forms for sums

> Is this there a better way to implement this? I don't know. Is there a better way to express what is desired in implementation such as WL's Method option? (I don't know but I am leaning toward adding options Method options for "numpy", "mpmath", and "Sympy" - however you tell me what you think.)

As a first stage, I think that the "manual" choice is fine. On the other hand, I guess that to decide which method to use automatically would not be very complicated.

PR #1136, opened 2021-01-28
Added calls to replace_wl_with_plain_text in boxes_to_text and boxes_to_xml for String and Symbol

GarkGarcia:
This is a follow up to https://github.com/orgs/mathics/teams/maintainers/discussions/20/comments/32. This PR makes it so that named characters are only replaced selectively (inside of strings or symbols). This makes calling `replace_wl_with_plain_text` from the clients unnecessary (we only need to set `use_unicode` field of `Evaluation` appropriately). 

After this is merged, PRs to the clients will follow.

### DIFF
--- a/mathics/core/evaluation.py
+++ b/mathics/core/evaluation.py
@@ -233,7 +233,8 @@ class Evaluation(object):
             definitions=None,
             output=None,
             format="text",
-            catch_interrupt=True
+            catch_interrupt=True,
+            use_unicode=True
     ) -> None:
         from mathics.core.definitions import Definitions
         from mathics.core.expression import Symbol
@@ -247,6 +248,7 @@ class Evaluation(object):
         self.stopped = False
         self.out = []
         self.output = output if output else Output()
+        self.use_unicode = use_unicode
         self.listeners = {}
         self.options = None
         self.predetermined_out = None
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -14,7 +14,7 @@ import typing
 from typing import Any
 from itertools import chain
 from bisect import bisect_left
-
+from mathics_scanner.characters import replace_wl_with_plain_text
 
 from mathics.core.numbers import get_type, dps, prec, min_prec, machine_precision
 from mathics.core.convert import sympy_symbol_prefix, SympyExpression
@@ -1793,7 +1793,16 @@ class Symbol(Atom):
         return Symbol(self.name)
 
     def boxes_to_text(self, **options) -> str:
-        return str(self.name)
+        name = str(self.name)
+
+        if "evaluation" in options:
+            e = options["evaluation"]
+            return replace_wl_with_plain_text(name, use_unicode=e.use_unicode)
+        else:
+            return name
+
+    def boxes_to_xml(self, **options) -> str:
+        return replace_wl_with_plain_text(str(self.name))
 
     def atom_to_boxes(self, f, evaluation) -> 'String':
         return String(evaluation.definitions.shorten_name(self.name))
@@ -2528,6 +2537,10 @@ class String(Atom):
             value.startswith('"') and value.endswith('"')):
             value = value[1:-1]
 
+        if "evaluation" in options:
+            e = options["evaluation"]
+            value = replace_wl_with_plain_text(value, e.use_unicode)
+
         return value
 
     def boxes_to_xml(self, show_string_characters=False, **options) -> str:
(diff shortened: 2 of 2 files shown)


mathics/core/evaluation.py
```diff
@@ -233,7 +233,8 @@ def __init__(
             definitions=None,
             output=None,
             format="text",
-            catch_interrupt=True
+            catch_interrupt=True,
+            use_unicode=True
```
rocky, line 237: I think this is worng. This is a property of formatting, not evaluation.

GarkGarcia: @rocky I agree, but I don't know where to place it.

GarkGarcia:
I am concerned about the implications of this PR. Maybe it's better to leave the conversion to the clients, even if we have to traverse the AST twice.

GarkGarcia:
> I am concerned about the implications of this PR. Maybe it's better to leave the conversion to the clients, even if we have to traverse the AST twice.

@rocky @mmatera Opinions?

mmatera:
I found similar problems working with the graph engine. Later I send you a resume of how I think we can face this

mathics/core/expression.py
```diff
@@ -1793,7 +1793,16 @@ def do_copy(self) -> 'Symbol':
...
+        name = str(self.name)
+
+        if "evaluation" in options:
+            e = options["evaluation"]
+            return replace_wl_with_plain_text(name, use_unicode=e.use_unicode)
+        else:
+            return name
+
+    def boxes_to_xml(self, **options) -> str:
+        return replace_wl_with_plain_text(str(self.name))
```
mmatera, line 1805: Shouldn't this depend on the option passed to the Evaluation object?

mathics/core/expression.py
```diff
@@ -2543,7 +2556,7 @@ def boxes_to_xml(self, show_string_characters=False, **options) -> str:
         text = self.value
 
         def render(format, string):
-            encoded_text = encode_mathml(string)
+            encoded_text = encode_mathml(replace_wl_with_plain_text(string))
```
mmatera, line 2559: The  same that in Symbol

mmatera (changes requested):

mmatera:
Well, this took to me more time than I had expected. After working on #1140, and having a closer look at how to implement the proposal in this PR, I think that @GarkGarcia is right: the best choice seems to add a parameter in Evaluation.
Let me explain why, and what could be the alternatives. 
In #1140 I implemented a way to overwrite the default behavior of MakeBoxes, in a way that a client can redefine them. This mechanism could be used also to deal with Strings: In the client, we could define a custom MakeBoxes for String expressions, that converts or not between the different string representations (wolfram-like utf8, standard utf8, ASCII named characters). However, this solution is difficult to implement for general Symbol and operators.
 
A second possibility would be to hack the output through the  $PrePrint symbol. This is what I did in IWolfram. I used this approach in IWolfram because I needed to make it work also with a WMA kernel, and there I do not have access to the internals. So, if a user tries to overwrite $PrePrint, the interface would get screwed.
 
On the other hand, the Evaluation class, already controls the formatting step. So it would be natural to introduce a property that controls how the final string should be printed. There would be two different ways to control from the front-end the behavior of Evaluation. One is what @GarkGarcia did: just add a string parameter. We could also allow for this parameter a custom call-back function, that takes the standard (WL)UTF8 and process it in the most convenient way for the front-end.

rocky:
> Well, this took to me more time than I had expected. After working on #1140, and having a closer look at how to implement the proposal in this PR, I think that @GarkGarcia is right: the best choice seems to add a parameter in Evaluation.
> Let me explain why, and what could be the alternatives.
> In #1140 I implemented a way to overwrite the default behavior of MakeBoxes, in a way that a client can redefine them. This mechanism could be used also to deal with Strings: In the client, we could define a custom MakeBoxes for String expressions, that converts or not between the different string representations (wolfram-like utf8, standard utf8, ASCII named characters). However, this solution is difficult to implement for general Symbol and operators.
> 
> A second possibility would be to hack the output through the $PrePrint symbol. This is what I did in IWolfram. I used this approach in IWolfram because I needed to make it work also with a WMA kernel, and there I do not have access to the internals. So, if a user tries to overwrite $PrePrint, the interface would get screwed.
> 
> On the other hand, the Evaluation class, already controls the formatting step. So it would be natural to introduce a property that controls how the final string should be printed. There would be two different ways to control from the front-end the behavior of Evaluation. One is what @GarkGarcia did: just add a string parameter. We could also allow for this parameter a custom call-back function, that takes the standard (WL)UTF8 and process it in the most convenient way for the front-end.

`replace_wl_with_plain_text()` is not something we want in the long run because it can't see the structure inside and that is important. Down the line, it would needs to be integrated better and driven by the format routine.

mmatera:
I am working on a better proposition. Once I finish with the checks, I am going to do a PR to this branch.

PR #1137, opened 2021-01-28
Small fixes in the pymathics mechanism 

mmatera:
Working in the matplotlib module I found certain deficiencies in the current implementation of LoadModule. Here I put those with the simplest solution.

### DIFF
--- a/mathics/builtin/assignment.py
+++ b/mathics/builtin/assignment.py
@@ -1748,20 +1748,17 @@ class LoadModule(Builtin):
         except ImportError as e:
             evaluation.message(self.get_name(), 'notfound', module)
             return SymbolFailed
-        except PyMathicsLoadException as e:
-            evaluation.message(self.get_name(), 'notmathicslib', module)
-            return SymbolFailed
-        else:
-            # Add PyMathics` to $ContextPath so that when user don't
-            # have to qualify PyMathics variables and functions,
-            # as the those in the module just loaded.
 
-            # Following the example of $ContextPath in the WL
-            # reference manual where PackletManager appears first in
-            # the list, it seems to be preferable to add this PyMathics
-            # at the beginning.
-            context_path = evaluation.definitions.get_context_path()
-            if "PyMathics`" not in context_path:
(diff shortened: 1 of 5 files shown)


mathics/builtin/base.py
```diff
@@ -60,7 +60,8 @@ def has_option(options, name, evaluation):
 
 class Builtin(object):
     name: typing.Optional[str] = None
-    context = "System`"
+    # context = "System`"
```
GarkGarcia, line 63: This should be cleaned before merging.

mathics/builtin/assignment.py
```diff
@@ -1748,20 +1748,17 @@ def apply(self, module, evaluation):
         except ImportError as e:
             evaluation.message(self.get_name(), 'notfound', module)
             return SymbolFailed
-        except PyMathicsLoadException as e:
```
GarkGarcia, line 1751: Why was this block removed?

mathics/core/definitions.py
```diff
@@ -81,11 +81,13 @@ def __init__(
                             module, remove_on_quit=False
                         )
                     except PyMathicsLoadException as e:
-                        print(e.module + " is not a valid pymathics module.")
-                        continue
+                        raise
```
GarkGarcia, line 84: I imagine this was changed for debugging (?). Anyway, we need to clean that too before merging.

GarkGarcia:
Besides https://github.com/mathics/Mathics/pulls#discussion_r566975813 this LGTM

mathics/builtin/assignment.py
mmatera: Because it is better to catch it in the apply method of LoadModule, in order to be able of sending an error message using the method in evaluation.

mathics/core/definitions.py
mmatera: Because it is better to catch it in the apply method of LoadModule, in order to be able of sending an error message using the method in evaluation.

rocky (approved):
LGTM

PR #1138, opened 2021-01-30
dry module loading

rocky:


rocky:
@mmatera Ok. I think this will work. There seems to be a bug probably uncovered with increased combinatorica testing and possibly old Python 2.7 behavior where you could call reveresed() with a generator. 

This has nothing to do with the changes here. Probably I will apply this bug fix to master later on.

PR #1140, opened 2021-01-30
Makeboxes overhaul

mmatera:
In this PR, I propose some changes in the way that the underlining MakeBoxes mechanism works. In Mathics 1.1 (and the current master branch) there are BoxConstruct classes, hard-linked to the interpreter, that defines how  (the `FullForm` of ) an expression is translated to a set of rules to build the output, that can be processed by the front-end.
In WMA, the interpreter is designed in a way that MakeBoxes can be redefined to adapt the output to different front-ends. 
With the proposed changes, the behavior of MakeBoxes in Mathics would be closer to the WMA, as  described in [Wolfram Reference: Textual Input and Output ](https://reference.wolfram.com/language/tutorial/TextualInputAndOutput.html#27901)
This would allow for example, to overload the MakeBoxes for different classes of expressions (Graphics, Strings, etc) by the Pymathics mechanism, in a way that different front-ends build and render boxes in the most suitable way, according to the capabilities of the interface.

This will help also to make a transparent interface for the Matplotlib pymathics package.


rocky (commented):
Overall, I very much like the idea and think all of this will be much much appreciated. 

And I totally agree that MakeBox has been too tied in with XML and/or other things in the core and we would benefit if as we add other kinds of formatting they too had a default makebox mechanism when the user hasn't overwritten this. 

I have looked the Makeboxes section in  [Wolfram Reference: Textual Input and Output](https://reference.wolfram.com/language/tutorial/TextualInputAndOutput.html#27901) and I think I understand what's up better. 

However I'd like to think about this and try the code more. It is also possible that another organization might facilitate things here, but I am really speculating and don't really know. 

In my opinion rather than try to rush this to release, I'd prefer spending a little more time to think this over and try it.

mathics/builtin/graphics.py
```diff
@@ -3229,7 +3236,7 @@ def boxes_to_xml(self, leaves, **options):
             " ".join("%f" % t for t in (xmin, ymin, w, h)),
             svg,
         )
-
+        print("svg=", svg_xml)
```
rocky, line 3239: Probably intended to remove this print?

mathics/builtin/graphics.py
```diff
@@ -2598,8 +2601,8 @@ def to_svg(self):
         content = self.content.boxes_to_xml(evaluation=self.graphics.evaluation)
         style = create_css(font_color=self.color)
         svg = (
-            '<foreignObject x="%f" y="%f" ox="%f" oy="%f" style="%s">'
-            "<math>%s</math></foreignObject>"
+            '<foreignObject x="%f" y="%f" ox="%f" oy="%f" width="100" height="100" style="%s">'
```
rocky, line 2604: The magic width and height values 100 should probably be options that can be set to 100 by default.

What do you think?

rocky (commented):
Overall I really like this. Thanks so much for doing it.

The main question I have is, how can I see the new wonderful behavior. 

Is there any test I can run to see the difference?

mathics/builtin/graphics.py
mmatera: I just added these parameters to see if it was a problem of boundaries. It wasn't...

mmatera:
> Overall I really like this. Thanks so much for doing it.
Your welcome!
>The main question I have is, how can I see the new wonderful behavior? Would you describe something please?
At this point, the idea is that you cannot see it from mathics-core: everything should work as before. To see the differences, you would need to see in the code, or trying to write a pymathics module that uses this new behaviour.

> Is there any test I can run to see the difference?
In principle, if you load (my version of) pymathics-matplotlib, you would see the difference (in one case it works, otherwise no).

rocky:
Should we be able to write a test that alters `Makebox[]` in a valid way and see the change in behavior there?

mmatera:
> Overall I really like this. Thanks so much for doing it.
> 
> The main question I have is, how can I see the new wonderful behavior? Would you describe something please?
> 
> Is there any test I can run to see the difference?
I added some basic tests in te 

> Overall I really like this. Thanks so much for doing it.
> 
> The main question I have is, how can I see the new wonderful behavior? Would you describe something please?
> 
> Is there any test I can run to see the difference?

@rocky, please look at  test/test_custom_boxconstruct.py to see some very basic tests. 

If there are no extra thoughts, observations, or complaints, I would like to merge this during the weekend to finish implementing the pymathics-matplotlib and pymathics-asy.

mathics/builtin/base.py
```diff
@@ -362,7 +362,7 @@ def get_option_string(self, *params):
         return None, s
 
 
-class InstancableBuiltin(Builtin):
+class InstanceableBuiltin(Builtin):
```
rocky, line 365: This name feels a bit weird to me.  In a PR off of this I might have changed the name. I don't remember. Right now I can't come up with another name.

rocky:
> > Overall I really like this. Thanks so much for doing it.
> > The main question I have is, how can I see the new wonderful behavior? Would you describe something please?
> > Is there any test I can run to see the difference?
> > I added some basic tests in te
> 
> > Overall I really like this. Thanks so much for doing it.
> > The main question I have is, how can I see the new wonderful behavior? Would you describe something please?
> > Is there any test I can run to see the difference?
> 
> @rocky, please look at test/test_custom_boxconstruct.py to see some very basic tests.
> 
> If there are no extra thoughts, observations, or complaints, I would like to merge this during the weekend to finish implementing the pymathics-matplotlib and pymathics-asy.

Will do with pleasure, but I don't I'll have time to look at before the weekend.

mathics/builtin/base.py
rocky: Ok. I can't come up with another name and it problably isn't that important either. 

Sorry for the delay - thanks for your patience.

PR #1144, opened 2021-02-01
Add Method option to N[]

rocky:
and some other small tweaks. 

I view this as a small follow-on to the constants.py work. In fact, I was reminded of this when going over the release notes. Specifying the method for contants this way is more more useful than the option setting. 

It also points out a descrepency, in the precision and in how numpy works - it doesn't have precison; at least not as flexiibly as mpmath and sympy. 

Probably after this release some of htis will be addressed. I need to google what's up with the precision for mpmath and numpy, and I may ask on forums, or stackoverflow.

I will have probably add a few tests using these options later and before release.


mmatera:
@rocky, Numbers in WL have ``Precision``, which is indicated, for example, as
```3.212`3```
If it is not specified, precision is equal to MachinePrecision:
```
In[1]:= Precision[1]
Out[1]= Infinity

In[2]:= Precision[1.2]
Out[2]= MachinePrecision

In[3]:= Precision[Pi]
Out[3]= Infinity

In[4]:= Precision[3.14`2]
Out[4]= 2.

In[5]:= N[MachinePrecision]
Out[5]= 15.9546

```
For numbers/expressions ``a`` with a ``Precision[a]<MachinePrecision``, I would use the numpy version of ``N[a]``, and then convert to a fixed precision number, if Precision != MachinePrecision.  For numbers/expressions with ``Precision[a]>MachinePrecision``, the fixed precision routine should be used always, in order to get a consistent result.
(At least, this is how it works in WMA)

**Comment by rocky** [comment]
@mmatera Ok - thanks for the information. If you or @GarkGarcia are up to fixing this and it is clear how to do, please do. Otherwise, I'll look at the upcoming weekened.

**Comment by mmatera** [comment]
@rocky I think this is ready: this does not involve very drastic changes but includes a fix for Information that you started and I completed, that would be great to have inside the release. Please have a look.

**Review by mmatera — APPROVED** [review]

**Comment by mmatera** [comment]
Just one thing: I marked the test in ``Information`` as private (``#>``) because they weren't properly handled by the doc builder. This could be improved in a future version, but as by now just affects  just a single symbol, I thought we can leave it as it is now, to avoid larger changes.

File: mathics/builtin/constants.py [hunk 1]
```diff
@@ -43,7 +47,16 @@ def mp_convert_constant(obj, **kwargs):
 
 
 def numpy_constant(name: str, d=None) -> float:
-    return getattr(numpy, name)
+    if d:
+        # by mmatera: Here I have a question:
+        # 0.0123`2 should be rounded to
+        # 0.01 or to 0.0123?
```
rocky, line 53: I don't know the answer either. 

Let's leave this for later though...

rocky (commented):
@mmatera 

Is this okay to go in now?

PR #1147, opened 2021-02-06
Fix PossibleZeroQ

mmatera:
* a fraction is zero if the numerator is zero and the denominator is not zero. 
* fix MachineReal.is_approx_zero


mmatera:
> @mmatera - sorry for changing master here. We don't want what is in master here and below with respect to removing the PossibleQ tests which should now work.

If at the end everything works, it is OK. So, what do you think we need to change here?

rocky:
> > @mmatera - sorry for changing master here. We don't want what is in master here and below with respect to removing the PossibleQ tests which should now work.
> 
> If at the end everything works, it is OK. So, what do you think we need to change here?

Commit d686e8f897ea1fa6edec923d41edf3cb3f048bd9 where tests get added back in is what I meant. 

Please check over and if this is okay, just merge in - thanks.

PR #1148, opened 2021-02-07
Opt e

rocky:
`matics -e` showing just the calculation is more useful for embed use. It matches what we have over in mathicsscript too.


mmatera (approved):
LGTM

PR #1152, opened 2021-02-13
Fix pymathics context

mmatera:
@rocky , This PR fixes several bugs in the pymathics mechanism. With this, graph works again (see https://github.com/Mathics3/mathicsscript/issues/23 )


rocky (approved):
Looked at in more detail. Although I see this addresses scoping, I don't understand it all that much. 

However I did test it on mathicsscript and mathics-django and it fixes these issues. 

Thanks! 

I have merged this to rc1 (dropping an unneeded import that my editor was telling me about.)

rocky:
Since rc1 is imminent, probably we can merge rc1 into master after release and this should show up as more or less a noop.

PR #1154, opened 2021-02-14
Fixing Information

mmatera:
Here, there are some improvements in the way ``Information`` symbol works, coming from PR #1144. Over this, there is also an improvement regarding the connection between Definition objects and Builtin objects, allowing to a better decoupling between the mathics core module and the loading of ``pymathics`` modules: docstrings are now acceded throw the ``builtin`` property of the definition object, instead of looking at the mathics.builtin.builtins dictionary.

### DIFF
--- a/mathics/builtin/assignment.py
+++ b/mathics/builtin/assignment.py
@@ -5,7 +5,7 @@
 import mathics.builtin
 from mathics.builtin.base import (
     Builtin, BinaryOperator, PostfixOperator, PrefixOperator)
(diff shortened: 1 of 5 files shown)


rocky (commented):
Merged into rc1. As soon this clears CI I will be releasing Mathics. 

@mmatera if you have a chance also look at mathics-django rc1 which has the changes merged but one of the files was reformated using "black" so I needed to convert characters into the Python3 Unicode equivalents.

(Down the line all of this will be replaced by using tables in mathics-scanner)

PR #1155, opened 2021-02-14
decoupling from mathics.builtin.builtins

mmatera:
This PR has the aim of continuing to remove hard-linked objects in the mathics core. In particular, this removes mathics.builtin.builtins
This dictionary was used in some formating operations, the definition of certain Test symbols, and in the doctest mechanism.  The problem with this comes up when we implemented the pymathics mechanism. Then, to fulfill the functionality of true built-in symbols, the global dictionary should be modified, which also has consequences if we want to make to coexist two instances of MathicsSession objects. 
In any case, with the present changes, this global object would be not necessary anymore.

### DIFF
--- a/mathics/builtin/system.py
+++ b/mathics/builtin/system.py
@@ -407,7 +407,11 @@ class UserName(Predefined):
     name = "$UserName"
 
     def evaluate(self, evaluation) -> String:
-        return String(os.getlogin())
+        try:
+            return String(os.getlogin())
+        except:
+            import pwd
+            return String(pwd.getpwuid(os.getuid())[0])
 
 
 class Version(Predefined):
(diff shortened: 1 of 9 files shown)


mathics/builtin/system.py
```diff
@@ -407,7 +407,11 @@ class UserName(Predefined):
     name = "$UserName"
 
     def evaluate(self, evaluation) -> String:
-        return String(os.getlogin())
+        try:
+            return String(os.getlogin())
+        except:
+            import pwd
+            return String(pwd.getpwuid(os.getuid())[0])
```
rocky, line 414: @mmatera his should be a separate PR and merged in.

mmatera: See PR  #1163.  If pass CI, let's merge it.

mmatera:
Already included in PR #1140

PR #1159, opened 2021-02-18
Unprotect $InstallationDirectory

rocky:
Do any of these also need to be unprotected?:

@mmatera  @GarkGarcia $BaseDirectory $HomeDirectory $InitialDirectory $UserBaseDirectory

### DIFF
--- a/mathics/builtin/files.py
+++ b/mathics/builtin/files.py
@@ -240,7 +240,7 @@ class InitialDirectory(Predefined):
 class InstallationDirectory(Predefined):
     """
     <dl>
-    <dt>'$InstallationDirectory'
+      <dt>'$InstallationDirectory'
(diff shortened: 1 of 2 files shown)


rocky:
Not sure why this is failing because it works when I run locally.

mmatera:
$InstallationDirectory is the place where the Kernel is installed. Why would you like to overwrite it?

mmatera:
@rocky, I do not see why you think that these symbols should be unprotected. In principle, all of them represent fixed places  in the filesystem. Why would you like to overwrite them?

rocky:
This PR was in response to something someone asked for. I can't find the issue or comment, and I think we had decided that this wasn't the right thing to do anyway. 

So closing.

PR #1160, opened 2021-02-18
basic implementation of NIntegrate

mmatera:
Here it is a basic implementation of NIntegrate, working both on intervals and d-dimensional regions.
If scipy.integrate is available, uses quad as a default integrator, and allows to  call other routines (romberg, and  nquad, but others can be included later). Otherwise, or if these routines fails, a native Simpson's rule adaptative  routine is included, and set as default if scipy is not available.
With #1161, I expect that this also works for complex valued functions.


mathics/builtin/compilation.py
```diff
@@ -50,10 +49,10 @@ class Compile(Builtin):
     #> cf[0, -2]
      = 0.5
 
-    Loops and variable assignments are not yet supported
+    Loops and variable assignments are supported as python (not llvmlite)
+    functions
     >> Compile[{{a, _Integer}, {b, _Integer}}, While[b != 0, {a, b} = {b, Mod[a, b]}]; a]       (* GCD of a, b *)
-     : Expression While[b != 0, {a, b} = {b, Mod[a, b]}] ; a could not be compiled.
-     = Function[{Global`a, Global`b}, While[b != 0, {a, b} = {b, Mod[a, b]}] ; a]
+     =  CompiledFunction[{a, b}, a, -CompiledCode-]
```
rocky, line 55: When I assign this to a variable and run that on anythng anything other than 0 as the `b` argument this hangs. I am guessing this definition isn't quite right?

rocky (commented):
Overall I like this! Thanks! 

On additional suggestion is to change CompiledCode#__str__ to report _which_ variation of compilation was used now that several methods are tried.

mathics/builtin/compilation.py
rocky: One other thing I note is that when I look compilation.py pyflakes is reporting a number of small lint fixes.

rocky (commented):
I like this and it is much needed. Thanks!

mathics/builtin/compilation.py
mmatera: Yes to both things: we need at some point to review the implementation of `Compile` in a closer way. However, this is a task for a couple of weekends. What motivated me to dig into this right now is that I needed this work for the `NIntegrate` implementation.

mmatera:
> I like this and it is much needed. Thanks!
> 
> As suggestion for improvement if you are up for it is to add in `CompiledCode#__str__` an indication of what kind of compiled code we have now that there are several methods.
Thanks!  This last commit should attend to your suggestion.

PR #1161, opened 2021-02-20
Implementing Compile for non compilable llvmlite functions as python functions

mmatera:
In this PR, some ideas to improve the support of ``System`Compile`` are implemented, in a way more generic expressions can be converted to a python callable object, including complex valued functions. 
This mechanism is also used to implement a better behaviour for to_python over ``System`Function`` expressions: when called on these kind of expressions, to_python returns a python callable object which can be passed to things like optimizers and integrators.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,8 +21,12 @@ Enhancements
 Miscellanea
 +++++++++++
 
-A pass was made to improve Microsoft Windows compatability and testing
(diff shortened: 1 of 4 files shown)


rocky (approved):
LGTM - thanks for doing this!

I took the liberty of rebasing off of master (there were merge conflicts) and squashed some commits.

PR #1162, opened 2021-02-21
Minor tweaks...

rocky:
Back to you @mmatera for some small changes.

### DIFF
--- a/mathics/builtin/base.py
+++ b/mathics/builtin/base.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 
 import mpmath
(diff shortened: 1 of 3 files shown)


mmatera (commented):
@rocky I took note of your comments and I have incorporated some of them to #1140. If you agree, we can close this PR.

What remains open, and I think that would deserve some extra discussion, is the role of `InstanceableBuiltin` / `CustomizableBuiltin`  and `AtomBuiltin` in the Mathics'  class hierarchy.   Both classes refer to objects that both define an interface and also hold "live data", so represent both an Expression and a Python object that contain non-trivial data structures associated with that expression, that we shouldn't duplicate.

rocky:
Sure - close the PR is the useful ideas have been incorporated and there is
no benefit of having this any more.

As for the name, let me think of something. The name Instanceable feels
really awkward. I am sure there must be something more appropriate and less
awkward.

And as you say there might be a totally different organization for this.
But I don't have time to think  about it right now.

PR #1164, opened 2021-02-21
fix rational is_zero

mmatera:
Rational numbers are not created if the denominator is zero...


rocky:
Looks fine to me. Merge if this is ready.

Issue #1171, opened 2021-03-04
physchemdata module does not load.


mmatera:
The file `mathics/data/element.csv` is not available anymore, so this module never loads. Maybe we can put this module also as a pymathics module, and restore there this file, or look for a smarter way to load this file. This issue was detected using #1172.

mmatera:
Fixed in #1171

Issue #1169, opened 2021-03-02
mathics crashes with KeyError: 'System`MakeBoxes'


vsht:
Hi,

I apologize if I overlooked something, but on my system (Fedora 32) upon compiling Mathics
from git via `pip3 install -e . --user`, running `mathics` leads to the following error

```
$ mathics
/media/Data/Projects/Other/Mathics/mathics/builtin/datentime.py:130: DeprecationWarning: invalid escape sequence \$
  """
Traceback (most recent call last):
  File "/home/vs/.local/bin/mathics", line 11, in <module>
    load_entry_point('Mathics3', 'console_scripts', 'mathics')()
  File "/media/Data/Projects/Other/Mathics/mathics/main.py", line 280, in main
    definitions = Definitions(add_builtin=True, extension_modules=extension_modules)
  File "/media/Data/Projects/Other/Mathics/mathics/core/definitions.py", line 77, in __init__
    contribute(self)
  File "/media/Data/Projects/Other/Mathics/mathics/builtin/__init__.py", line 137, in contribute
    builtins["System`MakeBoxes"].contribute(definitions)
KeyError: 'System`MakeBoxes'
```

The same happens if I install mathics directly via pip using
```
pip install Mathics3  --user
pip install Mathics-Django  --user
pip install mathicsscript  --user
pip install pymathics-natlang  --user
pip install pymathics-graph  --user
```
which is what I initially did. Am I doing something wrong?

Cheers,
Vladyslav

rocky:
TL;DR remove _all_ mathics packages before installing first.

```
pip uninstall pymathics-graph
pip uninstall pymathics-natlang
pip uninstall mathicsscript
pip uninstall Mathics-Django
pip uninstall Mathics3
```

then check that everything is gone including what is in `/local/Data/Projects/Other`.

I will guess that you have a mixture of mathics installed using `--user`. Under what directory is this installing to? 

I see above `/home/vs/.local` and `/media/Data/Projects/Other`.  And if you are getting a deprecation warning you may have an older version. I am not getting a deprecation warning. What version of Python do you have installed? `python -V` will show that.

In the short term there are going to be API mismatches as we try to reorganize things. In theory the package versioning should catch this. In practice things are a bit too complex and mistakes in the past have been made (which can't be fixed if you have those installed).  Therefore it is best to delete old versions before installing a newer or different version.

vsht:
Sorry for the confusion. I didn't have any Mathics versions installed before installing via pip.
[...]

PR #1172, opened 2021-03-04
Check fails in loading modules.

mmatera:
With this PR, I try to detect problems associated with version mismatch (See #1169 for an example).


mmatera:
Now, this also fixes #1171.

rocky (approved):
I tried this locally and it looked good. 

About the import line. I added a `# noqa` See https://stackoverflow.com/questions/45346575/what-does-noqa-mean-in-python-comments

Also I ran black over the files and removed the lines that start:

```
#/usr/bin/env python3
```

which is _not_ appropriate here. These belong only on command-line files. 

@mmatera I had originally wanted to keep this separate but there were just all these conflicts when creating the branch. Also this change is a bit more than I would have liked.

Please check it over and if ok merge in.

PR #1173, opened 2021-03-04
Filenames implement

mmatera:
This PR adds `FileNames`, a symbol required in the `FeynCalc` package (see #1169)

PR #1176, opened 2021-03-07
Improve Windows coverage and fix Windows tests

rocky:
files.py: implement OperatingSystem option


rocky:
@mmatera there are some fixes in here that might be split ou

In `mathics/Mathics/mathics/builtin/importexport.py` there were dictionaries with duplicate keys. 
If you use a linter like pyflakes this will catch this kind of error. 

Also,  the black formatter (`pip install black`) can assist in finding mistakes like this and it would be very good to use when you have huge lists like this. 

I may try to get actions CI working on this later.

mmatera:
By now, looks very good! 

> Also, the black formatter (`pip install black`) can assist in finding mistakes like this and it would be very good to use when you have huge lists like this.

Indeed, today I installed pyflakes, and I found this issue. Thanks for solving it!

mmatera:
Also, maybe we can merge this branch with #1173 with has some similar contributions.

mathics/builtin/files.py
```diff
@@ -2531,21 +2518,19 @@ class ToFileName(Builtin):
...
-     = dir1/dir2/file
+     = dir1...dir2...file
 
     >> ToFileName["dir1", "file"]
-     = dir1/file
+     = dir1...file
 
     >> ToFileName[{"dir1", "dir2", "dir3"}]
-     = dir1/dir2/dir3
+     = dir1...dir2...dir3
```
mmatera, line 2533: `...` would be also a good way to deal with digits beyond accuracy in numeric tests!

mathics/builtin/datentime.py
```diff
@@ -126,62 +127,63 @@ def apply(self, evaluation):
...
-        try:
-            t = float(t.to_python())
-            evaluation.timeout_queue.append((t, datetime.now().timestamp()))
-            request = lambda : expr.evaluate(evaluation)
-            res = run_with_timeout_and_stack(request, t, evaluation)
-        except TimeoutInterrupt:
-            evaluation.timeout_queue.pop()
-            return failexpr.evaluate(evaluation)
-        except:
+if sys.platform != "win32":
```
mmatera, line 130: Is  TimeConstrained what is failing? or is the expression we have chosen to test it?

mmatera (commented):
LGTM

mathics/builtin/datentime.py
rocky: I believe the implementation is Unix/POSIX specific. I haven't investigated though.

rocky:
@mmatera this has been most not fun to work on. I have spent way too much time on this, so I am stopping work on this. 

Also I am merging this in because coping with the conflicts with master, in particular `bultins/importexport` were also a big chore. I resolved these mostly in favor of master at the expense of losing some fixes here. 

In general, it appears that doc tests should not rely on unicode output such as were found in Base64Encode and Print. Instead put those under pytest and test for character-set capability. 

In sum although this is incomplete, it is a start.

mmatera:
@rocky, thanks for all this work! I will think about how to solve the code page issue in the docstrings.

mmatera:
@rocky,completely agree with that criteria. However, I think that it is possible to implement tests using unicode / the character table of the platform. In any case, I think that  your work on improving compatibility improves a lot the quality of the code that we have in mathics.

PR #1177, opened 2021-03-08
ByteArray implementation enhancements

mmatera:
This PR fix some issues of the current implementation of ByteArray, and covers new functionalities like conversions to Normal List objects.

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -2697,33 +2697,33 @@ class ByteArrayAtom(Atom):
         elif type(value) is list:
             self.value = bytearray(list)
         elif type(value) is str:
-            self.value = base64.b64encode(bytearray(value, 'utf8'))
+            self.value = base64.b64decode(value)
         else:
             raise Exception("value does not belongs to a valid type")
         return self
 
     def __str__(self) -> str:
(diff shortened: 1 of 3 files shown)


mathics/core/expression.py
```diff
@@ -2697,33 +2697,33 @@ def __new__(cls, value):
...
             raise Exception("value does not belongs to a valid type")
         return self
 
     def __str__(self) -> str:
-        return '"' + base64.b64encode(self.value).decode('utf8') + '"'
+        return base64.b64encode(self.value).decode('utf8')
 
     def boxes_to_text(self, **options) -> str:
-        return '"' + base64.b64encode(self.value).decode('utf8') + '"'
+        return '"' + self.__str__() + '"'
```
rocky, line 2709: Note: think this is the same thing as `str(self)`

mmatera: Indeed, `self.__str__()` and `str(self)` should be the same thing with a different number of keystrokes. I thought this way was more "explicit", but as you prefer...

rocky: It was just an observation. Do any way you wish.

PR #1178, opened 2021-03-08
Unicode replace strings symbols mmatera

mmatera:
This implements some of the changes I think are necessary to make this PR acceptable:
* Always check if `Evaluation.use_unicode` is set to a boolean value, and pass this value to `replace_wl_with_plain_text`
* Allows the possibility to set `Evaluation.use_unicode=None`. In this case, replace_wl_with_plain_text is not called. Now the default is `True`. But if we put `None` as the default, the default behavior would be the current (without this PR).

Something that I didn't implement, but is a possibility, is to accept a function in `use_unicode`, in a way that the control of how the strings are formated be completely determined in the frontends.

### DIFF
--- a/mathics/builtin/logic.py
+++ b/mathics/builtin/logic.py
@@ -137,7 +137,7 @@ class Implies(BinaryOperator):
     If an expression does not evaluate to 'True' or 'False', 'Implies'
     returns a result in symbolic form:
     >> Implies[a, Implies[b, Implies[True, c]]]
-     = a \uF523 b \uF523 c
+     = a ⟹ b ⟹ c
     """
 
     operator = "\uF523"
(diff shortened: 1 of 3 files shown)


mmatera:
@rocky, @GarkGarcia 
Please have a look when you are available.

rocky (changes requested):
Sorry for the delay. I wanted to understand the bigger picture here and this took time. 

A bad reaction I am getting to our current code base is that we have hacks upon hacks and this one extends that. Rather I'd like to see the problem thought about for how to do this properly. 

The problem with a "short-term" fix like this is that when we do get around to fixing this, we'll have to undo more stuff than if we just had not added the hacks in the first place. 

The code that is wrong and needs fixing is https://github.com/mathics/Mathics/blob/master/mathics/builtin/base.py#L514-L519

Instead of substituting a string that has to be undone later, I think we need be calling a _function_ like `ToString[]` which consults [$CharacterEncoding](https://reference.wolfram.com/language/ref/$CharacterEncoding.html).

TiagoCavalcante:
@mmatera what is the status of this?

mmatera:
> @mmatera what is the status of this?

This is a draft with one idea about how to handle the formatting of strings.  I think now is outdated.

PR #1183, opened 2021-03-11
handle nonutf8 environments in test.py

mmatera:
Just in case, this seems to be the proper way to handle this, if we keep some of these examples in the docstrings


rocky (changes requested):
Sorry for the delay in looking at. 

What I now have come to realize is that what we need in [`$CharacterEncoding`](https://reference.wolfram.com/language/ref/$CharacterEncoding.html) and [`SystemCharacterEncoding`](https://reference.wolfram.com/language/ref/$SystemCharacterEncoding.html) 

`SystemCharacterEncoding` would be set in the front end from `sys.getdefaultencoding()`

I don't see any tests of of the usefulness of this code.  In the past we've gotten into the situation where code got into master, but it wasn't working and this forced us to fix up the code that should have been gone over better before merge.  

Let's not get into that situation again. Especially since I d not think this is where we should be focusing our time on. 

Rather, I'd like to see CellsToTeX finshined (including having it package ready so that we can just add ``Needs["CellsToTeX`"]``

mmatera:
> Sorry for the delay in looking at.
> 
> What I now have come to realize is that what we need is [`$CharacterEncoding`](https://reference.wolfram.com/language/ref/$CharacterEncoding.html) and [`SystemCharacterEncoding`](https://reference.wolfram.com/language/ref/$SystemCharacterEncoding.html)
> 
OK, this makes sense. 
> `SystemCharacterEncoding` would be set in the front end from `sys.getdefaultencoding()`
> 
> I don't see any tests of of the usefulness of this code. In the past we've gotten into the situation where code got into master, but it wasn't working and this forced us to fix up the code that should have been gone over better before merge.
> 
> Let's not get into that situation again. Especially since I d not think this is where we should be focusing our time on.
> 
OK, with this idea I think I could write something more definitive.

> Rather, I'd like to see CellsToTeX finshined (including having it package ready so that we can just add `` Needs["CellsToTeX`"] ``
That's in the way, but this code is particularly complicated because demands implement a lot of features. I think that once we have Export and ExportString functions we will have a big step forward on this.

mmatera:
Update:
This is the code in ToString:
```
    def apply(self, value, evaluation):
        'ToString[value_]'

        text = value.format(evaluation, 'System`OutputForm').boxes_to_text(
            evaluation=evaluation)
        return String(text)
```
So, it seems that the right strategy would be, in ``boxes_to_*`` to add an option controlling the encoding and, according to the encoding, apply the replace function.
Then, we also need to change in test.py the comparison: we should also evaluate the expected output. I can do this during this week.

rocky:
> Update:
> This is the code in ToString:
> 
> ```
>     def apply(self, value, evaluation):
>         'ToString[value_]'
> 
>         text = value.format(evaluation, 'System`OutputForm').boxes_to_text(
>             evaluation=evaluation)
>         return String(text)
> ```
> 
> So, it seems that the right strategy would be, in `boxes_to_*` to add an option controlling the encoding and, according to the encoding, apply the replace function.
> Then, we also need to change in test.py the comparison: we should also evaluate the expected output. I can do this during this week.

One little thing that shoudl be done earliy is to add `$CharacterEncoding` and `SystemCharacterEncoding` into Mathics. That would be consulted in `ToString[]` when no encoding is given as an option.

rocky:
@mmatera What does `ToString["->", CharacterEncoding->"Unicode"]` produce? 

More generally what's the WL function that converts ASCII combinations of operators like ">=" to their unicode equivalent?

rocky:
It occurs to me that the _Form_ parameter in `ToString[]` can interact with the encoding. In particular if you want TeXForm output, then the encoding is ASCII, not unicode.

mmatera:
> @mmatera What does `ToString["->", CharacterEncoding->"Unicode"]` produce?
> 
> More generally what's the WL function that converts ASCII combinations of operators like ">=" to their unicode equivalent?

In master, `CharacterEncoding-> ` is not taken into account. #1188 implements a first step in choosing the character encoding.

rocky:
> In master, CharacterEncoding-> is not taken into account. 

I meant in WL not Mathics.

mmatera:
> > In master, CharacterEncoding-> is not taken into account.
> 
> I meant in WL not Mathics.

This is in WolframScript:
```
 In[1]:=  A="∫ f  x"
 Out[1]= ∫ f  x


 In[2]:=  ToString[A]
 Out[2]= ∫ f  x


 In[3]:= ToString[A,CharacterEncoding->"ASCII"]
 Out[3]= \[Integral] f d x


 In[4]:= ToString[A,CharacterEncoding->"UTF-8"]
 Out[4]= â« f ï x


 In[5]:= ToString[A,CharacterEncoding->"Unicode"]
 Out[5]= ∫ f  x



 In[7]:= $SystemCharacterEncoding
 Out[7]= UTF-8
`

**Comment by rocky** [comment]
Ok. But in these examples you start with some non-ASCII unicode. I was asking about more than one ASCII character getting represented as a singe unicode characters. Like `->` or `>=`.

**Comment by mmatera** [comment]
> Ok. But in these examples you start with some non-ASCII unicode. I was asking about more than one ASCII character getting represented as a singe unicode characters. Like `->` or `>=`.

OK, that never happens, at least in my WolframScript. Maybe I should look what happens in the notebook interface...

**Comment by mmatera** [comment]
I just checked it: ToString["\[LeftArrow]",CharacterEncoding->(whatever)] gives you 
"\[LeftArrow]"

**Comment by rocky** [comment]
But is there a WL function that  WL converts _pairs_ of ASCII characters that make up operators, such as those given before, into _single_ unicode character, short of scanning and parsing?

Again what does `ToString["->", CharacterEncoding->"Unicode"]` produce?

**Comment by rocky** [comment]
And also you examples the other direction don't cover the information I was asking about. There is no multi-character ascii representation of Integral. The example in the integral with _d_ fo _dx_ might be an example, but it is a little ambiguous.

I asked specifically about `>=` and `->` specifically. What's the story with these? Thanks.

**Comment by mmatera** [comment]
> And also your examples the other direction don't cover the information I was asking about. There is no multi-character ascii representation of Integral. The example in the integral with _d_ in _dx_ to an ASCII d might be an example, but it is a little ambiguous.
> 
> ~I asked specifically about `>=` and `->` specifically. What's the story with these?~ Change that. I guess you are saying `->` becomes some sort of arrow. However if there is TeXForm added _then_ it would have to be `->`. Right?
> 
> Thanks.

I just give an example with the named character I have in mind.
```
 In[1]:= "\[GreaterEqual]"
 Out[1]= ≥

 In[2]:= ToString[">=",CharacterEncoding->"Unicode"]
 Out[2]= >=
```
and so one... Named characters are interpreted when you enter the string, not when you evaluate it. The same with the short-hand notation -> ,  >= , etc.

In anycase, we need the opposite: given a character, convert it to a printable form.

rocky:
Superceded by #1188

PR #1187, opened 2021-03-14
Poke's fast import

mmatera:
@poke1024 @rocky  This is a rebase of the poke's branch for delayed import of built-in modules. Thit is working, and loads very fast! Also adds some code for doing tests for the load time of different modules.


mmatera:
I wonder if it would be hard to implement the same for pymathics modules

rocky:
I am not seeing any difference: 

```
# Current
$ time mathics -e '1+2'
3

real	0m3.526s
user	0m3.913s
sys	0m2.076s

# delayed import
$ time mathics -e '1+2'
3

real	0m3.851s
user	0m4.296s
sys	0m2.038s

# Second run
$ time mathics -e '1+2'
3

real	0m3.508s
user	0m3.915s
sys	0m2.071s
```

mmatera:
Indeed, the rate is about 3.5 sec/ 3.7 sec of the master initialization time,  in my machine. However, this could be useful for debug (by setting `BENCHMARK_STARTUP=True`  in `mathics.settings.py`).

rocky:
> However, this could be useful for debug (by setting BENCHMARK_STARTUP=True in mathics.settings.py).

There are a ton of tools for benchmarking Python programs that don't require changing the code drastically like this.

Those are what I'd start out with in benchmarking.

PR #1188, opened 2021-03-14
Add  $SystemCharacterEncoding and CharacterEncoding in boxes_to_text and ToString[]

mmatera:
A first try about how ToString should manage the `CharacterEncoding` option (see #1183)

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -486,7 +486,7 @@ class BaseExpression(KeyComparable):
         finally:
             evaluation.dec_recursion_depth()
 
-    def format(self, evaluation, form) -> typing.Union["Expression", "Symbol"]:
+    def format(self, evaluation, form, **kwargs) -> typing.Union["Expression", "Symbol"]:
         """
         Applies formats associated to the expression, and then calls Makeboxes
         """
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -17,6 +17,7 @@ from mathics.core.numbers import get_type, dps, prec, min_prec, machine_precisio
 from mathics.core.convert import sympy_symbol_prefix, SympyExpression
 import base64
 
+from mathics_scanner.characters import replace_wl_with_plain_text
 
 def fully_qualified_symbol_name(name) -> bool:
     return (
@@ -1904,7 +1905,9 @@ class Symbol(Atom):
         return Symbol(self.name)
 
     def boxes_to_text(self, **options) -> str:
-        return str(self.name)
+        if options.get("encoding", "UTF-8") in ("UTF-8", "Unicode"):
+            return str(self.name)
+        return replace_wl_with_plain_text(str(self.name), False)
 
     def atom_to_boxes(self, f, evaluation) -> "String":
         return String(evaluation.definitions.shorten_name(self.name))
@@ -2723,7 +2726,9 @@ class String(Atom):
         ):
             value = value[1:-1]
 
-        return value
+        if options.get("encoding", "UTF-8") in ("UTF-8", "Unicode"):
+            return value
+        return replace_wl_with_plain_text(value, False)
 
     def boxes_to_xml(self, show_string_characters=False, **options) -> str:
         from mathics.core.parser import is_symbol_name
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -17,7 +17,6 @@ from mathics.core.numbers import get_type, dps, prec, min_prec, machine_precisio
 from mathics.core.convert import sympy_symbol_prefix, SympyExpression
 import base64
 
-from mathics_scanner.characters import replace_wl_with_plain_text
 
 def fully_qualified_symbol_name(name) -> bool:
     return (
@@ -127,7 +126,12 @@ def from_python(arg):
         #     return Symbol(arg)
     elif isinstance(arg, dict):
         entries = [
-            Expression("Rule", from_python(key), from_python(arg[key]),) for key in arg
+            Expression(
+                "Rule",
+                from_python(key),
+                from_python(arg[key]),
+            )
+            for key in arg
         ]
         return Expression(SymbolList, *entries)
     elif isinstance(arg, BaseExpression):
@@ -486,7 +490,9 @@ class BaseExpression(KeyComparable):
         finally:
             evaluation.dec_recursion_depth()
 
-    def format(self, evaluation, form, **kwargs) -> typing.Union["Expression", "Symbol"]:
+    def format(
+        self, evaluation, form, **kwargs
+    ) -> typing.Union["Expression", "Symbol"]:
         """
         Applies formats associated to the expression, and then calls Makeboxes
         """
@@ -508,7 +514,7 @@ class BaseExpression(KeyComparable):
     def get_option_values(self, evaluation, allow_symbols=False, stop_on_error=True):
         options = self
         if options.has_form("List", None):
-            options = options.flatten(Symbol("List"))
+            options = options.flatten(SymbolList)
             values = options.leaves
         else:
             values = [options]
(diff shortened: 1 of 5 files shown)


mathics/core/expression.py
```diff
@@ -1774,17 +1781,21 @@ def thread(self, evaluation, head=None) -> typing.Tuple[bool, "Expression"]:
...
-            "Subtract",
-            "Minus",
-            "Power",
-            "Abs",
-            "Divide",
-            "Sin",
-        ) and all(leaf.is_numeric() for leaf in self._leaves)
+        return (
+            self._head.get_name()
+            in system_symbols(
```
mmatera, line 1786: Actually, this should ask to the Definitions object if the head corresponds to a NumericFunction symbol.

mmatera (commented):
@rocky: After our talk yesterday, I was trying to figure out how to put the encoding parameter in the Makeboxes (format) function instead of on the boxes_to_text method.  In the last approach, you start with an expression, you call MakeBoxes, and then you have expressions consisting of `*Box` expressions with arguments that are all strings (including operators). Then, it is more or less easy to change Unicode internal characters to something that you can represent in your frontend. Also is something that makes sense if you think about what string encoding means.

The other way around requires a lot more work, and I am not sure that is indeed correct. To do this, you need to change the way in which the formatting of operators is currently implemented. In the current implementation, each operator defines the format (including the characters that represent the operator) before the Definitions object is built. For this reason, it cannot ask for which encoding use. To change to an scheme where the character representing the operator is defined on runtime, would require a big refactory, and at the end, we would get something less efficient, and without any improvement in functionality regarding the other approach. 

The unique important think to take care of is that each (unicode) character in the internal representation has a proper form in system codepage, and that you can always introduce the corresponding character, in the system codepage, and convert it at some point to the internal codepage. Having the \[NamedCharacter] this is always ensured. So, I believe that the best way to face this is the way in which I had implemented before, but replacing the replace function with a one with more complete support for other codepages.

mmatera:
BTW, it worths to note that in WMA, when a .m or a .nb file is saved, characters are stored in the `\[NamedCharacter]` (ascii) form.

mathics/core/expression.py
rocky: Good observation. But let's address this in a separate PR.

mmatera:
Another insight: 
```
wolframscript -c 'ToString[s \[LessEqual] b, CharacterEncoding->"Cualquiera"]'

Get::noopen: Cannot open /opt/Wolfram/CDPlayer/11.3/SystemFiles/CharacterEncodings/Cualquiera.m.
s <= b
```
So, WMA seems to make the encoding doing a byte by byte change, according to the content of the files in the `CharacterEncodings` folder. For example, the file associated to `ISOLatin2.m` contains:

```
(* "ISOLatin2 *)
{"8Bit",
  {{161, "\:0104"}, {162, "\:02d8"}, {163, "\:0141"}, {165, "\:013d"}, 
   {166, "\:015a"}, {169, "\:0160"}, {170, "\:015e"}, {171, "\:0164"}, 
   {172, "\:0179"}, {174, "\:017d"}, {175, "\:017b"}, {177, "\:0105"}, 
   {178, "\:02db"}, {179, "\:0142"}, {181, "\:013e"}, {182, "\:015b"}, 
   {183, "\:02c7"}, {185, "\:0161"}, {186, "\:015f"}, {187, "\:0165"}, 
   {188, "\:017a"}, {189, "\:02dd"}, {190, "\:017e"}, {191, "\:017c"}, 
   {192, "\:0154"}, {195, "\:0102"}, {197, "\:0139"}, {198, "\:0106"}, 
   {200, "\:010c"}, {202, "\:0118"}, {204, "\:011a"}, {207, "\:010e"}, 
   {208, "\:0110"}, {209, "\:0143"}, {210, "\:0147"}, {213, "\:0150"}, 
   {216, "\:0158"}, {217, "\:016e"}, {219, "\:0170"}, {222, "\:0162"}, 
   {224, "\:0155"}, {227, "\:0103"}, {229, "\:013a"}, {230, "\:0107"}, 
   {232, "\:010d"}, {234, "\:0119"}, {236, "\:011b"}, {239, "\:010f"}, 
   {240, "\:0111"}, {241, "\:0144"}, {242, "\:0148"}, {245, "\:0151"}, 
   {248, "\:0159"}, {249, "\:016f"}, {251, "\:0171"}, {254, "\:0163"}, 
   {255, "\:02d9"}
}}
```

rocky:
> 
> 
> @rocky: After our talk yesterday, I was trying to figure out how to put the encoding parameter in the Makeboxes (format) function instead of on the boxes_to_text method. In the last approach, you start with an expression, you call MakeBoxes, and then you have expressions consisting of `*Box` expressions with arguments that are all strings (including operators). Then, it is more or less easy to change Unicode internal characters to something that you can represent in your frontend. Also is something that makes sense if you think about what string encoding means.

The front-end can still call whatever  conversion functions it wants to. That function is still there in the scanner. 

Doing it here though is wrong, however tempting, and makes a mess when it isn't desired.
> 

> The other way around requires a lot more work, and I am not sure that is indeed correct. 

Yes, I am aware that it is a more work. But adding a bit of code now that has to be undone later isn't a good idea. 

At any rate, we wanted to get a release out sooner than later.  So let's do that.

rocky:
https://github.com/mathics/Mathics/pull/1188#issuecomment-808958029

Ok. Let's come back to this after the release. What's here I don't think does harm like it did before and certainly is less code.

mmatera:
> > @rocky: After our talk yesterday, I was trying to figure out how to put the encoding parameter in the Makeboxes (format) function instead of on the boxes_to_text method. In the last approach, you start with an expression, you call MakeBoxes, and then you have expressions consisting of `*Box` expressions with arguments that are all strings (including operators). Then, it is more or less easy to change Unicode internal characters to something that you can represent in your frontend. Also is something that makes sense if you think about what string encoding means.
> 
> The front-end can still call whatever conversion functions it wants to. That function in the scanner modules is still there and can be imported.

The problem is that you need to call this function BEFORE calling boxes_to_tex/boxes_to_xml, which is something that is done internally by the ```evaluate``` method of the  ```Evaluation```, which is called by the front-ends, and which returns a string, not a Box expression.


> 
> Doing it here though is wrong, however tempting, and makes a mess when it isn't desired.
> 
OK, until yesterday, I agree with you in this. However, I found some evidence that WMA does this in this way, and I am pretty sure that they do that for a reason.
> > 
> 
> > The other way around requires a lot more work, and I am not sure that is indeed correct.
> 
> Yes, I am aware that it is a more work. But adding a bit of code now that has to be undone later isn't a good idea.
> 
OK, but again, the motivation for doing in the other way is not how fast we get something working, but because after looking at this hard, I think that is the better approach.

> At any rate, we wanted to get a release out sooner than later. So let's do that.

OK. Let's keep this as it is now, and improve this afterwards.

rocky:
> The problem is that you need to call this function BEFORE calling boxes_to_tex/boxes_to_xml,

It is hard to argue about this because in fact both are wrong. Yes I suppose it would be "less wrong" to do that. But either way, it really shouldn't be done. 

> However, I found some evidence that WMA does this in this way, and I am pretty sure that they do that for a reason.

I think you are mistaken that WMA does it this way. It might not do it the way that this branch works either, but again I think overall this does the lesser damage.

What we wanted was $SystemCharacterEncoding and a ToString[] that that could be iterated on. 

> OK. Let's keep this as it is now, and improve this afterwards.

Thanks. We can come back to this after release.

rocky:
> BTW, it worths to note that in WMA, when a .m or a .nb file is saved, characters are stored in the `\[NamedCharacter]` (ascii) form.

What this says to me, and what I do in mathicsscript is save the Mathics Expression form as well as the printed form. With this, if I need to write it out, the Expression can be called through a different formatter. 

And even if that isn't done (but it should be), there are maps from unicode symbol to named character.

PR #1189, opened 2021-03-16
No longer need PyYAML. Simpfiy Windows CI test

rocky:


mmatera (approved):
I put the explicit list of packages to see if some of them produce troubles. Now I think it is safe to remove them

PR #1191, opened 2021-03-20
Try simplying SympyFunction builtins

rocky:
Would have expected to catch/DRY more items here. But I am optimistic
that by starting this as we add more Sympy functions we will be making
more use of this.

I think it is better to have one way that is used regularly than open-code this
a number of places.


mmatera (approved):
LGTM

mmatera:
BTW, if you have time, something that is missing in Mathics is the `Series` and `Collect` builtins. Sympy has these functions, if you have time, it would be great to try this mechanism with these functions.

rocky:
Ok - will look at when I get a chance. Added https://github.com/mathics/Mathics/issues/1193 as a reminder. 

Would really like to see CellsToTeX merged into master. What's up with that?

mmatera:
> Ok - will look at when I get a chance. Added #1193 as a reminder.
> 
> Would really like to see CellsToTeX merged into master. What's up with that?

Celltotex requires improving the support for Export. This is where I was putting most of my recent work. Once the support for this be estabilized, I can continue with celltotex.

PR #1192, opened 2021-03-20
More descriptive and general in prep for profiling

rocky:
Memoization isn't haven't a big effect although I do see it is doing something. More important from a maintainablilty

is the blackening of some files and clearification with by starting to rename `apply_2`, etc with more descriptive name and to use
"%(name)s" so that when things move around we don't have to change that.

### DIFF
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -66,6 +66,7 @@ class _MPMathFunction(SympyFunction):
 
     nargs = 1
 
+    @lru_cache(maxsize=1024)
     def get_mpmath_function(self, args):
         if self.mpmath_name is None or len(args) != self.nargs:
             return None
@@ -75,7 +76,7 @@ class _MPMathFunction(SympyFunction):
         "%(name)s[z__]"
 
         args = z.numerify(evaluation).get_sequence()
-        mpmath_function = self.get_mpmath_function(args)
+        mpmath_function = self.get_mpmath_function(tuple(args))
         result = None
 
         # if no arguments are inexact attempt to use sympy
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -10,6 +10,7 @@ Basic arithmetic functions, including complex number arithmetic.
 from mathics.version import __version__  # noqa used in loading to check consistency.
 import sympy
 import mpmath
+from functools import lru_cache
 
 from mathics.builtin.base import (
     Builtin,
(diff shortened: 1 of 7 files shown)


mmatera:
@rocky, about memoization, the kernel has already implemented an expression cache, in a way that if you call several times the same (mpmath/numpy/whatever) function, the result is read from a lookup table instead of calling apply_ methods.

mmatera:
With the blackening, and using more descriptive names, I completely agree.

rocky:
> about memoization, the kernel has already implemented an expression cache, in a way that if you call several times the same (mpmath/numpy/whatever) function, the result is read from a lookup table instead of calling apply_ methods.

@mmatera - Let me check this by timing... 

Without: 
```
In[1]:= Timing[RealDigits[Pi, 10, 1, -10000]]
Out[1]= {4.51084, {{8}, -9999}}

In[2]:= Timing[RealDigits[Pi, 10, 1, -10000]]
Out[2]= {6.65212, {{8}, -9999}}
```

with: 
```In[1]:= Timing[RealDigits[Pi, 10, 1, -10000]]
Out[1]= {4.50568, {{8}, -9999}}

In[2]:= Timing[RealDigits[Pi, 10, 1, -10000]]
Out[2]= {3.432, {{8}, -9999}}
```

Please doublecheck though...

mmatera:
> > about memoization, the kernel has already implemented an expression cache, in a way that if you call several times the same (mpmath/numpy/whatever) function, the result is read from a lookup table instead of calling apply_ methods.
> 
> @mmatera - Let me check this by timing...
> 
> Without:
> 
> ```
> In[1]:= Timing[RealDigits[Pi, 10, 1, -10000]]
> Out[1]= {4.51084, {{8}, -9999}}
> 
> In[2]:= Timing[RealDigits[Pi, 10, 1, -10000]]
> Out[2]= {6.65212, {{8}, -9999}}
> ```
> 
> with:
> 
> ```
> Out[1]= {4.50568, {{8}, -9999}}
> 
> In[2]:= Timing[RealDigits[Pi, 10, 1, -10000]]
> Out[2]= {3.432, {{8}, -9999}}
> ```
> 
> Please doublecheck though...

OK, looks appealing. However, this seems to indicate that the expression cache is not working properly...
This weekend I'm a little busy. I can look at this over the next week.

rocky:
> However, this seems to indicate that the expression cache is not working properly...
> This weekend I'm a little busy. I can look at this over the next week.

No problem. There is a lot of evidence to suggest that caching at a higher level would be a big win. 

Possibliy related by possibily different: CProfile data suggests that most of the time is spent higher up such as in pattern matching. If that is so, caching at a higher level would be a big win. Alternatively there's probably somehting at the pattern matching that could be done to speed things up, but  I haven't been able to get the right level of granularity in pattern.py:394 `match_leaf()`  to understand what's up.

mmatera (approved):
LGTM

rocky:
@mmatera We should merge this and continue with other performance efforts, right?

mmatera:
> @mmatera We should merge this and continue with other performance efforts, right?

Yes, I think that the changes here help and do not hurt future improvements. I would focus the efforts on the ExpressionCache before tackling specific things like numerics.

PR #1196, opened 2021-03-24
Make a pass over documents

rocky:
Include deletion in conjunction of worksheets.
Go over typographical setting in various new docs.


mmatera (approved):
LGTM

PR #1199, opened 2021-03-25
Knot theory fixes

mmatera:


rocky:
@mmatera  think I understand how to write the type table-lookup code I was describing in a rough way. 

What would help though is are some useful tests that come up.

mathics/builtin/files.py
```diff
@@ -712,7 +715,17 @@ def apply(self, channel, types, evaluation, options):
 
         # Wrap types in a list (if it isn't already one)
         if not types.has_form("List", None):
+            if types.get_head_name() == "System`Hold":
```
rocky, line 718: This seems hoaky since the speck seems to allow any function and not just `Hold[]` right?

rocky (changes requested):
Please let's start splitting this up into the individual pieces and create separate PRs for this.

For example Setting System`Notebooks is a thing in of itself and not even needed for Feyncalc.

That could be merged in by itself.

The problem with these growing grab-bag PRs is that they are impossible to review and the quality can suffer because we can't focus on individual features. 

And then they are continually merged (when rebasing instead of merging would lead to more comprehensible results)

mathics/builtin/strings.py
```diff
@@ -829,6 +829,9 @@ class StringSplit(Builtin):
         <dd>splits $s$ at the delimiter $d$.
     <dt>'StringSplit[$s$, {"$d1$", "$d2$", ...}]'
         <dd>splits $s$ using multiple delimiters.
+    <dt>'StringSplit[{$s_1$, $s_2, $\ldots$}, {"$d1$", "$d2$", ...}]'
```
rocky, line 855: Note that ldots doesn't render in the Django interface.

mmatera:
Spli

> Please let's start splitting this up into the individual pieces and create separate PRs for this.
> 
> For example Setting System`Notebooks is a thing in of itself and not even needed for Feyncalc.
> 
> That could be merged in by itself.
> 
> The problem with these growing grab-bag PRs is that they are impossible to review and the quality can suffer because we can't focus on individual features.
> 
> And then they are continually merged (when rebasing instead of merging would lead to more comprehensible results)

Splitted on #1219, #1209 #1216, #1217

rocky:
Superceded by other PRs.

PR #1204, opened 2021-03-25
Singletonize Symbols frequently used in Head 

mmatera:
This is an attempt to reduce the overhead in comparisons and pattern matching. @rocky, could you profile this against master?

### DIFF
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -31,6 +31,7 @@ from mathics.core.expression import (
     Real,
     String,
     Symbol,
+    SymbolN,
     SymbolFalse,
     SymbolNull,
     SymbolTrue,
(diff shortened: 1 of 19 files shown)


rocky:
> could you profile this against master?

@mmatera Will try to get to over the weekend.

rocky:
@mmatera I can't detect a change with this over master. Still, the PR moves us is in the right direction. 

To show this and get a feel for what we _should_ instead be focusing on to gain performance here is a flamegraph of some code
mathics-profile

If you want the raw data which is more useful let me know. github doesn't want me to paste these here.

rocky:
@mmatera there were merge conficts in the branch. I rebased. We are seeing a Windows problem that probably has nothing to do with this. 

Still, given that this isn't a big win, let's hold off on this until after the release. It is also likely that when ToString comparisons get in to master this will correct itself.

PR #1208, opened 2021-03-29
mglyph -> mtext broke some images

rocky:
@mmatera I used [`git bisect`](https://thoughtbot.com/blog/git-bisect) and tracked the graphics misbehaving to 616dbbae9c9480878c0752488724a74b61640b99  and specificlly the one-line change here. 

What does this change make possible? If we can't figure out something that accomodates both concerns, maybe we can add a setting so both ways are possible.

BTW This PR is the kind of thing I was mentioning about large PR's that are hard to follow. 

The title of  #1173 is "adds FileNames, a symbol required in the FeynCalc package"  and I don't see how this change which is about graphics has anything to do with ByteArray or FileNames. Even the commit message "fix write"  isn't suggestive.

### DIFF
--- a/mathics/builtin/graphics.py
+++ b/mathics/builtin/graphics.py
@@ -3226,14 +3226,15 @@ clip(%s);
             svg,
         )
 
-        return (
-            '<mtext><img width="%dpx" height="%dpx" src="data:image/svg+xml;base64,%s"/></mtext>'
-            % (
+        # mglyph, which is what we have been using, is bad because MathML standard changed.
+        # metext does not work beacause the way in which we produce the svg images is also based on this outdated mglyph behaviour.
+        # format = "<mtext><img width="%dpx" height="%dpx" src="data:image/svg+xml;base64,%s"/></mtext>"
+        template = '<mglyph width="%dpx" height="%dpx" src="data:image/svg+xml;base64,%s"/>'
+        return template % (
                 int(width),
                 int(height),
                 base64.b64encode(svg_xml.encode("utf8")).decode("utf8"),
             )
-        )
 
     def axis_ticks(self, xmin, xmax):
         def round_to_zero(value):


mmatera:
The problem here is that both ways are broken: mglyph does not work anymore because MathML standard changed. The mtext way does not work because the way in which we produce the svg images is also based on this outdated mglyph behaviour. However, the mtext version allows us to patch this with the pymathics modules we are developing, while the mglyph doesn't.

mathics/builtin/graphics.py
```diff
@@ -3227,7 +3227,7 @@ def boxes_to_xml(self, leaves=None, **options):
         )
 
         return (
-            '<mtext><img width="%dpx" height="%dpx" src="data:image/svg+xml;base64,%s"/></mtext>'
+            '<mglyph width="%dpx" height="%dpx" src="data:image/svg+xml;base64,%s"/>'
```
mmatera, line 3230: Could you just leave the <mtext> line as a comment? Apart from that, go ahead.

mmatera (changes requested):
leave the  <mtext>  line as a comment.

mathics/builtin/graphics.py
rocky: Sure adding information about what's up and possible change and the pros and cons of that is a great idea.

Some that could be improved is that instead of this getting broken and _then_ we start to add what was wrong why it was changed it would be better done when the change went in. And not mixed inside a whole bunch of other unrelated stuff. 

I've added what you report so far, but I am still a bit hazy on what's up. 

> mglyph does not work anymore because MathML standard changed. 

Please elaborate here. It does work in the sense that it is producing images. What in the MathML standard changed? Maybe links to the change would be helpful. 

> The mtext way does not work because the way in which we produce the svg images is also based on this outdated mglyph behaviour.

Ok, so then the change using mtext is incomplete and I wonder why then we were using it. 

> However, the mtext version allows us to patch this with the pymathics modules we are developing, while the mglyph doesn't.

I don't understand this either. If you can patch mtext you should be able to patch mglyph in the same way. In fact it feels like patching mglyphs is more selective than patching mtexts because mglyph is probably only used for graphics while mtext is used for plain text as well.

mmatera: > Sure adding information about what's up and possible change and the pros and cons of that is a great idea.
> 
> Some that could be improved is that instead of this getting broken and _then_ we start to add what was wrong why it was changed it would be better done when the change went in. And not mixed inside a whole bunch of other unrelated stuff.
> 
> I've added what you report so far, but I am still a bit hazy on what's up.
> 
> > mglyph does not work anymore because MathML standard changed.
> 
> Please elaborate here. It does work in the sense that it is producing images. What in the MathML standard changed? Maybe links to the change would be helpful.
>

OK. Let's elaborate on this a little bit. A long time ago, when the first implementation of the graphics Mathics engine was defined, MathML standard allowed to nest MathML code inside a glyph, so you can render a MathML code, which contains a glyph (graphics) that contains inside MathML code. This was great because it allowed mimicking exactly what happens in graphics in WMA. This works while the Mathjax version shipped with Mathics supported the feature. 
As SVG supported also this feature, I had the idea that we can export SVG graphics just by taking the XML form of the output of graphics, by just hacking the header. And we all were very happy with our solution.

However, at some point, this feature allowing all of this was discontinued in MathML and then in Mathjax. Then, someone updated the MathML backend shipped with Mathics, and then the graphics engine got broken: labels (which are included as MathML code inside the glyph) were not rendered anymore. 

Then, the question was, what would be the most rational way to implement the same functionality with the current standards? The right way would be to change  `<glyph>` by something that allows embedding an image. Here is where `<mtext>` enters in the game: it allows to include plane HTML inside MathML code. The problem is that the mechanism to generate images still relayed on the mathics XML generator, which now produces broken SVG files. Then, what we need to fix is the generation of the SVG images, not going back to glyph.

However, since the code that fixes the export to SVG also fixes the MakeBoxes[Graphics... ], your change back to broken code does not put in risk the future solution. 

 
> > The mtext way does not work because the way in which we produce the svg images is also based on this outdated mglyph behaviour.
> 
> Ok, so then the change using mtext is incomplete and I wonder why then we were using it.
> 


As I explained before, `<mtext>` is not the root of the problem: the problem is that we are not able by now to produce the svg image that this tag requires.

> > However, the mtext version allows us to patch this with the pymathics modules we are developing, while the mglyph doesn't.
> 
> I don't understand this either. If you can patch mtext you should be able to patch mglyph in the same way. In fact it feels like patching mglyphs is more selective than patching mtexts because mglyph is probably only used for graphics while mtext is used for plain text as well.


After looking at the standard, I arrived at the conclusion that `<mtext>` is more suitable than `<mglyph>`. Maybe I am wrong, or there is another better option, but it was not just a quick patch...

rocky: Many thanks for the explanation. I really appreciate understanding what's going on and think I now understand the complexities better. 

> I had the idea that we can export SVG graphics just by taking the XML form of the output of graphics, by just hacking the header. And we all were very happy with our solution.

We've gone over this before: XML is not SVG or MathML. if you want to export SVG, I would imagine you'd create a to_svg routine. Underneath SVG's are text but they have their own format. (For the record, I have never been happy with calling SVG and MathML XML.) 

> After looking at the standard, I arrived at the conclusion that <mtext> is more suitable than <mglyph>. 

I am not opposed to this. But it has to work probably and right now it doesn't. I tried before adding width and length to the mtext, but that didn't do anything. 

> Maybe I am wrong, or there is another better option, but it was not just a quick patch...

It feels to me like there must be a better approach because Django should not have to render everything as MathML. Manytimes when the top-level object is an SVG  it would be nice to reder it as SVG. And we've run into the problem that we can't render HTML links either. 

I don't doubt this isn't a quick patch.  My frustration right now is that there are tons of these quick fixes such as the stuff for handling unicode operators, and when we add more quick fix patches that are not well thought out but just sort of work most of the time for now, then I at least have a harder time finding where to remove that when we finally start to move towards a more complete solution. 

Sorry to rant. This aspect has been the most unfun part of this code base.

PR #1209, opened 2021-03-30
Fixcomparisons

mmatera:
In this PR I included several fixes for the Equal symbol. This pass all the tests I could imagine, but the implementation is not very elegant...


mmatera:
Well, here I will try to explain what is what we have now, why I do not think that we can have a more general framework to avoid the if's and what I think could be improved in the future.

`Equal` tries to determine if two or more expressions are "equivalent". So, two expression objects with the same head and the same leaves are "equal", as it can be check by calling the  `same` method of  `BaseExpression`.
When the expressions are not the same (let's say `Pi` and `4 ArcTan[1]`), then the specialization is needed:

* For a couple of exact numbers, (integers, rational), or complex numbers with exact real and imaginary parts, the comparison should be exact, and resolved by `sympy` by exact methods.
* For pairs of atoms  (numbers, `Image`, and any other possible object that should be handled as a single thing), except symbols (which are also atoms) comparison is performed using the `__eq__` method of the corresponding classes.
* The only case in which a couple of different symbols should be explicitly compared are `True!=False` and `False!=True`.
* Lists are equal if they are element by element.
* As equality of two compiled functions are defined by their first two arguments, but not by the result of the compilation, we need to handle in an especial way too.
* The rest of the cases that can be compared are expressions that can be evaluated to numbers, or expressions involving `Infinity`, `ComplexInfinity`, and `DirectedInfinity`. This should rely on `sympy`. However, Sympy does not handle properly expressions with complex values or infinity values, so these special cases must be preprocessed.


Also, we need to take into account that this routine is one in which relies on many of the computations of the kernel, so it should be as fast as possible.


I analyzed the possibility of handling all these cases with a dictionary, as @rocky suggested. However, as the objects are of different classes (symbols, atoms of different types, expressions) such an approach would require an important overhead. For example, to compare two real numbers, we would need first to generate their "head" symbols, compare them, and then compare the numbers. With the current approach, we just need to check that the expressions are atoms and rely on the comparison of real numbers.

What is actually susceptible to improvements are the routines where do_compare are used: `_EqualityOperator.apply` and `_EqualityOperator.apply_other`, where for example in checking equality, all the possible permutations of members are verified. Also, we would gain something in performance by singletonizing the heads of built-in objects.
On the other hand, to help the readability of the code, maybe we should comment better what is supposed that the code does on each step, and maybe also split do_compare into smaller functions.

rocky:
@mmatera For me, none of this was either surprising or unexpected and I had imagined how such things would be taken care of. 

That said, I clearly didn't convey this very well. And it is good that with these details in hand when we redo things, it will make be clearer as to why. (And probably in the design in my mind I may have omitted a couple of steps). 

However I won't be able to address this in any detail until the weekend.

mmatera:
I added a link to the previous comment as a comment in `do_compare()`. The idea is not to review this right now but to have a better idea of what is the next step.

test/test_compare.py
```diff
@@ -1,23 +1,644 @@
 # -*- coding: utf-8 -*-
```
mmatera, line 1: This file includes all the tests...

mmatera (commented):
I didn't realize that blacken each file I change actually produces a lot of changes that difficult the revision. Bellow there is a detail of what I actually change here.

mathics/builtin/arithmetic.py
```diff
@@ -1135,12 +1135,16 @@ class DirectedInfinity(SympyFunction):
     }
 
     def to_sympy(self, expr, **kwargs):
-        if len(expr.leaves) == 1:
+        if len(expr._leaves) >= 1:
```
rocky, line 1138: Why do we need ">=" instead of "=="? 

https://reference.wolfram.com/language/ref/DirectedInfinity.html indicates there is either zero or one leaves only?

mmatera: For upward compatibility, and because I was too lazy to check the documentation :)

PR #1210, opened 2021-03-30
Fix to svg

mmatera:
This PR fixes 
* the internal generation of SVG images from the internal generator (to_svg) including text inside (like labels)
* the Export of graphics (and symbols) to a valid SVG.
This is achieved by using `boxes_to_text` instead of `boxes_to_xml` in the `InsetBox.to_svg` method. 

This still does not fixes the issue with the Django's interface: if we use `<mtext><\mtext>`, images are not properly bounded in the front end. On the other hand, this trick does not helps if we use `<mglyph ...>`: boundaries are correct in this case, but labels does not appears. I keep this last form, that can be handled making small modifications to `mathics.js` in the Django frontend (the change in  the corresponding PR).

image

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,7 @@ Enhancements
 * Improvements in handling products with infinite factors: ``0 Infinity``-> ``Indeterminate``, and ``expr Infinity``-> ``DirectedInfinite[expr]`.
 * ``SetDelayed`` now accept several conditions impossed both at LHS as well as RHS.
 
+  
 Bug fixes
 +++++++++
 
@@ -53,6 +54,13 @@ Miscellanea
 Windows under MSYS.
 * Include numpy version in version string. Show in CLI
 * Small CLI tweaks ``--colors=None`` added to match mathicsscript.
+* In the ``BaseExpression`` and derivated classes, the method ``boxes_to_xml`` now are called ``boxes_to_mathml``.
+* In the ``format`` method of the class ``Evaluation``,  the builtin ``ToString`` is called instead of  ``boxes_to_text``
+  In order to control the final form of boxes from the user space in specific symbols and contexts.
+* ``GraphicsBox`` now have two methods:  ``to_svg`` and  ``to_mathml``. The first produces SVG plain text while the second produces ``<mglyph ...>``
+  tags with base64 encoded svgs.
+* Improving the support for ``Inset`` and  ``InsetBox``.
+
 
 2.0.0
 -----
--- a/mathics/autoload/formats/SVG/Export.m
+++ b/mathics/autoload/formats/SVG/Export.m
@@ -4,13 +4,17 @@ Begin["System`Convert`TextDump`"]
 
 
 SVGExport[filename_, expr_, opts___] := 
-  Module[{strm, data}, 
+  Module[{strm, data, p, q, expr2}, 
     strm = OpenWrite[filename];
     If[strm === $Failed, Return[$Failed]];
-    If[System`$UseSansSerif,
-	    data = StringTake[ToString[MathMLForm[expr]],{23,-8}],
-	    data = StringTake[ToString[MathMLForm[expr]],{23,-8}]];
-    WriteString[strm, "<svg>" <> data <> "</svg>"];
+    expr2 = If[Head[expr]=!=System`Graphics, System`Graphics[{System`Inset[ToString[expr]]}], expr];
+	 expr2= MathMLForm[expr2];
+    data=ToString[expr2];
+    p = StringPosition[data, "data:image/svg+xml;base64"][[1]][[2]];
+    (*Let's assume that the end of the string is reached just before the last quote. *)
+    q = StringPosition[data,"\""][[-1]][[-2]];
+    data = StringTake[data  ,{p+2,q-1}];
+    WriteString[strm, System`Convert`B64Dump`B64Decode[data]];
     Close[strm];
   ]
 
--- a/mathics/builtin/graphics.py
+++ b/mathics/builtin/graphics.py
@@ -119,25 +119,26 @@ def cut(value):
     return value
 
 
-def create_css(edge_color=None, face_color=None, stroke_width=None, font_color=None):
+def create_css(edge_color=None, face_color=None, stroke_width=None, font_color=None, opacity=1.0):
     css = []
     if edge_color is not None:
-        color, opacity = edge_color.to_css()
+        color, stroke_opacity = edge_color.to_css()
         css.append("stroke: %s" % color)
-        css.append("stroke-opacity: %s" % opacity)
+        css.append("stroke-opacity: %s" % stroke_opacity)
     else:
         css.append("stroke: none")
     if stroke_width is not None:
         css.append("stroke-width: %fpx" % stroke_width)
     if face_color is not None:
-        color, opacity = face_color.to_css()
+        color, fill_opacity = face_color.to_css()
         css.append("fill: %s" % color)
-        css.append("fill-opacity: %s" % opacity)
+        css.append("fill-opacity: %s" % fill_opacity)
     else:
         css.append("fill: none")
     if font_color is not None:
-        color, opacity = font_color.to_css()
+        color, _ = font_color.to_css()
         css.append("color: %s" % color)
+    css.append("opacity: %s" % opacity)
     return "; ".join(css)
 
 
@@ -294,8 +295,8 @@ def _CMC_distance(lab1, lab2, l, c):
 
 def _extract_graphics(graphics, format, evaluation):
     graphics_box = Expression(SymbolMakeBoxes, graphics).evaluate(evaluation)
-    builtin = GraphicsBox(expression=False)
-    elements, calc_dimensions = builtin._prepare_elements(
+    # builtin = GraphicsBox(expression=False)
+    elements, calc_dimensions = graphics_box._prepare_elements(
         graphics_box.leaves, {"evaluation": evaluation}, neg_y=True
     )
     xmin, xmax, ymin, ymax, _, _, _, _ = calc_dimensions()
@@ -498,7 +499,14 @@ class Graphics(Builtin):
                     for atom in atoms
                 ):
                     if head == "System`Inset":
-                        n_leaves = [content.leaves[0]] + [
+                        inset = content.leaves[0]
+                        if inset.get_head_name() == "System`Graphics":
+                            opts = {}
+                            # opts = dict(opt._leaves[0].name:opt_leaves[1]   for opt in  inset._leaves[1:])
+                            inset = self.apply_makeboxes(
+                                inset._leaves[0], evaluation, opts
+                            )
+                        n_leaves = [inset] + [
                             Expression(SymbolN, leaf).evaluate(evaluation)
                             for leaf in content.leaves[1:]
                         ]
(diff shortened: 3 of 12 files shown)


rocky:
@mmatera I looked at this a little bit and following the logic in this traceback I see that the state of things is in a mess that I don't understand.

```
->0 boxes_to_xml(self=<String: "20">, show_string_characters=False, **options={'evaluation': <mathics.core.evaluatio...)
     called from file '/src/external-vcs/github/mathics/Mathics/mathics/core/expression.py' at line 2766
##1 to_svg(self=<mathics.builtin.graphics.InsetBox object at 0x7f9c859ec278>)
     called from file '/src/external-vcs/github/mathics/Mathics/mathics/builtin/graphics.py' at line 2590
##2 <genexpr>(.0=<list_iterator object at 0x7f9c859ec4e0>)
     called from file '/src/external-vcs/github/mathics/Mathics/mathics/builtin/graphics.py' at line 2912
##3 to_svg(self=<mathics.builtin.graphics.GraphicsElements object at 0x7f9c859b7710>)
     called from file '/src/external-vcs/github/mathics/Mathics/mathics/builtin/graphics.py' at line 2912
##4 boxes_to_xml(self=<mathics.builtin.graphics.GraphicsBox object at 0x7f9c859d6a58>, leaves=(<Expression: System`L...)
     called from file '/src/external-vcs/github/mathics/Mathics/mathics/builtin/graphics.py' at line 3201
##5 apply_mathml(self=<mathics.builtin.inout.MathMLForm object at 0x7f9ca50d0d30>, expr=<Expression: System`Standard...)
     called from file '/src/external-vcs/github/mathics/Mathics/mathics/builtin/inout.py' at line 2087
....
```

Starting from the bottom at frame 5...
The apply_mathml is fine. It's where it decides to call GraphicsBox#boxes_to_xml that is odd. More clear would be a boxes_to_mathml() method because that is the kind of xml we want and I don't see how any other kind of XML is going to make things nicer. In fact the change here suggests that sometimes we don't want boxes_to_xml but boxes_to_text.

Then inside Graphics#`boxes_to_xml` immediately the implication is that the only kind if xml thing for a graphics box is an SVG!?  Well I suppose right now that is the only thing we have implemented. 

I see what looks like a hack `self._prepare_elements()` which does operator to unicode symbol because somehow if I am in a Graphics#boxes_to_xml I clearly must know that I can handle unicode?  Wrong!

Ths works only because of particular kinds of situtations we have right now. Django (for example) handles unicode and it only knows about MathML, and the only `Graphics#boxes_to_xml()` is that which is handled by MathML. So because of this particular chain of events we currently have, there is the assumption that anything that gets to `Graphics#boxes_to_xml`l must understand unicode.

But if we think in terms of modularity: just convert graphics boxes to xml (which doesn't make a lot of sense) or convert boxes to mathml (which makes more sense) and ask does this mean mathml must handle unicode? it is less clear. Maybe handling mathml requires handling unicode. I don't know, but even if it does, is this really the right place to do this. I doubt it.

And so when  the `boxes_to_xml` to `boxes_to_text`  is changed you lose that feature. But since it was a hack anyway it won't matter when other things are fixed up. But I hope you understand now why adding hacks that have occured in the past to substitute unicode characters I have a strong reaction to. 

Who would think to look `Graphiss#boxes_to_xml` for conversion of ascii operators to unicode? 

Instead of boxes_to_xml or boxes_to_text what I would expect is boxes_to_mathml. If there is some xml property that is used by mathml it can subclass or delegate xml (same as xml could subclass or delegate to text)

mathics/builtin/graphics.py
```diff
@@ -2602,7 +2602,7 @@ def to_svg(self):
         # svg = (
         #    '<foreignObject x="%f" y="%f" ox="%f" oy="%f" style="%s">'
         #    "<math>%s</math></foreignObject>")
-        svg = ('<text x="%f" y="%f" ox="%f" oy="%f" style="%s">' "%s" "</text>") % (
+        svg = ('<text x="%f" y="%f" ox="%f" oy="%f" style="text-anchor:middle; dominant-baseline:middle; %s">' "%s" "</text>") % (
```
rocky, line 2605: I suspect to do a good job a function is going to have to get called width and hight information and maybe other information about the SVG.

mmatera:
OK, I agree with much of what you said above, and I think that the key to unentangled all this mess is to modularize the code.
On the other hand, the idea is to do that without breaking the code, in a way that users are motivated to use it and contribute with more code and testing. This is why despite I think that this code for graphics would be removed as soon as possible, I tried to fix it to a minimally usable state. This is what this PR is about.  In the middle, I learned a little more about how graphics are working in the Django frontend.
Also, I have now more clear that what we need is to rethink how formatting is entangled with evaluation, but this is something to discuss in another comment.

mmatera:
Regarding this, please notice that most of the changes are related to blacken the files. I will mark specifically which are the changes I implement in the code.

mathics/autoload/formats/SVG/Export.m
```diff
@@ -4,13 +4,15 @@
...
 SVGExport[filename_, expr_, opts___] := 
-  Module[{strm, data}, 
+  Module[{strm, data, p, expr2}, 
     strm = OpenWrite[filename];
     If[strm === $Failed, Return[$Failed]];
-    If[System`$UseSansSerif,
-	    data = StringTake[ToString[MathMLForm[expr]],{23,-8}],
-	    data = StringTake[ToString[MathMLForm[expr]],{23,-8}]];
-    WriteString[strm, "<svg>" <> data <> "</svg>"];
+    expr2 = If[Head[expr]=!=System`Graphics, System`Graphics[{System`Inset[ToString[expr]]}], expr];
```
mmatera, line 10: I change this to fix the Export to SVG functionality, in a way that is more independent on how we format the return string in boxes_to_xml

mathics/builtin/graphics.py
mmatera: This is the main change I made here: just use boxes_to_text instead boxes_to_xml.

mathics/autoload/formats/SVG/Export.m
```diff
@@ -4,13 +4,15 @@
...
+	 expr2= MathMLForm[expr2];
+    data=ToString[expr2];
+    p = StringPosition[data, "data:image/svg+xml;base64"][[1]][[2]];
+    data = StringTake[data  ,{p+2,-19}];
```
rocky, line 14: Magic numbers like `p+2`, `-19` worry me and are likely to be fragile.

Or expressions like `[[1]][[2]]`. What do these mean and do?

rocky:
Two additional things I would like to see, but they might be added as follow-on PRs:

* rename `Graphics#boxes_to_xml` to `Graphics#boxes_to_mathml` since that is what it is. 
* Add the Matrix example shown in the screenshot to the Django gallery.

mathics/builtin/graphics.py
rocky: > 
> 
> I suspect to do a good job a function is going to have to get called width and height information and maybe other graphics information that might be in the SVG in complicated form but in the Mathics Graphics object in less complicated form.

I may be wrong here. Possibly https://github.com/mathics/Mathics/blob/5aa0367be03d9e912f5c082fae33abeec87b1650/mathics/builtin/graphics.py#L3355-L3356 is the culprit. The size should be determined by the number of items in xmin to xmax and ymin to ymax.

mathics/autoload/formats/SVG/Export.m
mmatera: When you ask for the MathMLForm, the kernel returns you a mglyph, with the svg that you want encoded in b64.
To get the svg, we need to get the string with the svg encoded, which lies to bytes after the place where `data:image/svg+xml;base64` is found. Also, you need to remove the tail of `"><...>` to get something that you can decode.

mathics/test.py
```diff
@@ -25,6 +25,7 @@
```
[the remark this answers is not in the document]
mmatera: Agree: this shouldn't be here...

mmatera (commented):
> Two additional things I would like to see, but they might be added as follow-on PRs:
> 
>     * rename `Graphics#boxes_to_xml` to `Graphics#boxes_to_mathml` since that is what it is.
> 
You can't now: all the makeboxes scheme assumes that your boxes have a method `boxes_to_xml`. We should replace this name everywere.

>     * Add the Matrix example shown in the screenshot to the Django gallery.
Done in the corresponding PR in mathics-django

mmatera:
I insist with this: eventually all this code is going to be replaced by other code based on mpl. What I am trying here is to get a baseline of functionality, that should be supported by the mpl code.

rocky:
> I insist with this: eventually all this code is going to be replaced by other code based on mpl. What I am trying here is to get a baseline of functionality, that should be supported by the mpl code.

Ok. Then I have marked this as a draft. I tried pulling the code earlier and at least in the Django interface there is no difference. I guess that is intentional since this can't work there?

mmatera:
@rocky , you need both to use this branch of mathics, together with this [PR 28 ](https://github.com/Mathics3/mathics-django/tree/fixing-graphics2d) branch of Mathics-Django

mmatera:
For me, the plan is to merge this into master, merge the PR 28 in mathics-django now, with this ugly internal API on the side of mathics. Then, the idea would be to improve the internal API of mathics, renaming the functions that format the expressions. This does not affect the way in which the available clients get then formatted output (through the evaluation object), so this PR could be shipped in the next release.
On the other hand, I am working on a proposal for such a change in the API, but I want to have a working version before share it. This new API would be included in a future release (Let's say, passing to a 2.1 /2.2).

mmatera:
@rocky: I changed here the name ``boxes_to_xml`` to ``boxes_to_mathml`` according to your suggestion. In the end, I realized this does not impact too strongly on the API. I also changed ``boxes_to_text`` by a call to ``ToString`` in the evaluation object. If this passes the tests, I guess it covers all your comments.

mathics/builtin/graphics.py
rocky: @mmatera I spent some time looking again at this, and I am now pretty sure that it is not every "InsetBox" that you want to change but instead the thing that is calling this which is called "create_axes".

mmatera: Actually, using text or mathml should depend on what is the Expression: for graphics objects (points, lines, disks,...) should be mathml or svg while for anything else, text.

mmatera: @rocky, my last attempt on this:  ``Inset`` was not implemented, except in the case of strings and expressions (labels, including axes labels). With the current implementation, MathML support is not possible. I added reduced support for ``Graphics`` insets, but full support would require more work than what I plan to invest in this. 
If this works, the next step is to work out the MPL package, in a way that all that we have here will be implemented there.

CHANGES.rst
```diff
@@ -53,6 +54,13 @@ Miscellanea
 Windows under MSYS.
 * Include numpy version in version string. Show in CLI
 * Small CLI tweaks ``--colors=None`` added to match mathicsscript.
+* In the ``BaseExpression`` and derivated classes, the method ``boxes_to_xml`` now are called ``boxes_to_mathml``.
+* In the ``format`` method of the class ``Evaluation``,  the builtin ``ToString`` is called instead of  ``boxes_to_text``
+  in order to control the final form of boxes from the user space in specific symbols and contexts.
```
rocky, line 59: in order -> In order
`to_svg` -> ``to_svg``w
plane -> plain

rocky (approved):
Great work! LGTM

PR #1211, opened 2021-03-30
blacken all the files... That makes sense?

mmatera:
This PR actually shouldn't change anything, except to make the code more uniform. @rocky, that makes sense, or it would screw all the other PRs?


rocky:
If this succeeds let's merge it. Then _rebase_ the outstanding PR's on top of this merged master.

Thanks.

mmatera:
> If this succeeds let's merge it. Then _rebase_ the outstanding PR's on top of this merged master.
> 
> Thanks.

The only thing that bothers me with black is the use of `"` for strings, which forces to escape all the `"` in Mathics String expressions...

rocky:
> > If this succeeds let's merge it. Then _rebase_ the outstanding PR's on top of this merged master.
> > Thanks.
> 
> The only thing that bothers me with black is the use of `"` for strings, which forces to escape all the `"` in Mathics String expressions...

If internally there is a double quote, then black will not change the outer quote whatever it is " or ''' or """

PR #1104, opened 2021-01-09
WIP: fixes to make cellstotex to work

mmatera:
Tidier version of PR #995

PR #1215, opened 2021-03-31
WIP: CellsToTeX support

mmatera:
rebase of #1104


rocky:
Two things that would be super to do. 

1. Squash all of the commits down into those commits that are meaningful and not part of something else. This is done using `git rebase -i b0e63ed5eb23fd6c4b9baf922695128405a1de20`
2. Include a copy of CellsToTeX in packages with whatever version changes are useful. We were given the go ahead to do so.

mmatera:
> Two things that would be super to do.
> 
>     1. Squash the 75 or so  commits down into those commits that are meaningful and not part of something else. This is done using `git rebase -i b0e63ed5eb23fd6c4b9baf922695128405a1de20`
> 
>     2. Include a copy of CellsToTeX in packages with whatever version changes are useful. We were given the go ahead to do so.

From this PR, the only thing that deserves to be conserved is the test_cellstotex.py file. Now I just copy it in a clean branch of master.

Regarding of leaving a copy, as this is a package that is automatically downloaded from internet, I think that is not a good idea. What maybe makes sense is to have a "loader" that before loading from internet, fix the version of the language (now you need to do that by hand before loading).

test/test_cellstotex.py
```diff
@@ -0,0 +1,107 @@
...
+    expected2 = _evaluate('{"CellToTeX", "CellsToTeXException", "CellsToTeXPreamble"}')
+    print(result2)
+    assert result2 == expected2
+
+
+@pytest.mark.skipif(not url_reachable, reason="skipping since we can't reach %s" % external_url)
+@pytest.mark.xfail
+def test_load_and_run():
+    print("load and run")
+    str_expected0 = "None"
```
rocky, line 59: All of this should be redone using the helper. 

Also in the future tests like this should be moved to mathics_omnibus.

PR #1220, opened 2021-03-31
Fix D and Derivative

mmatera:
This PR fixes another bug coming from a circular definition in the definition of `Derivative`.


mmatera:
superseeded by #1280

PR #1216, opened 2021-03-31
`StringSplit`: adding support for handling lists as a first argument

mmatera:
From #1199

PR #1132, opened 2021-01-26
Graphics overhault

mmatera:
This tries to merge and update PR  #635   @poke1024, would you mind having a look over this?

PR #1223, opened 2021-04-03
Svg axis and text opacity

rocky:
@mmatera with a change to the text opacity, I think we get most of the axis ugliness addressed. Also there is one slight adjustment to how svg `<text ..../>`  is set up. 

Let me know what you think.

### DIFF
--- a/mathics/builtin/graphics.py
+++ b/mathics/builtin/graphics.py
@@ -2617,15 +2617,17 @@ class InsetBox(_GraphicsElement):
             content = self.content.to_svg(noheader=True, offset=(x, y))
             svg = "\n" + content + "\n"
         else:
-            style = create_css(
+            css_style = create_css(
                 font_color=self.color, edge_color=self.color, face_color=self.color, opacity=self.opacity
             )
+            text_pos_opts = f'x="{x}" y="{y}" ox="{self.opos[0]}" oy="{self.opos[1]}"'
+            # FIXME: don't hard code text_style_opts, but allow these to be adjustable.
+            text_style_opts = "text-anchor:middle; dominant-baseline:middle;"
             content = self.content.boxes_to_text(evaluation=self.graphics.evaluation)
             svg = (
-                '<text x="%f" y="%f" ox="%f" oy="%f" style="text-anchor:middle; dominant-baseline:middle; %s">'
-                "%s"
-                "</text>"
-            ) % (x, y, self.opos[0], self.opos[1], style, content,)
+                f'<text {text_pos_opts} style="{text_style_opts} {css_style}">{content}</text>'
+            )
+
         # content = self.content.boxes_to_mathml(evaluation=self.graphics.evaluation)
         # style = create_css(font_color=self.color)
         # svg = (
--- a/mathics/builtin/graphics.py
+++ b/mathics/builtin/graphics.py
@@ -119,25 +119,26 @@ def cut(value):
     return value
 
 
-def create_css(edge_color=None, face_color=None, stroke_width=None, font_color=None):
+def create_css(edge_color=None, face_color=None, stroke_width=None, font_color=None, opacity=1.0):
     css = []
     if edge_color is not None:
-        color, opacity = edge_color.to_css()
+        color, stroke_opacity = edge_color.to_css()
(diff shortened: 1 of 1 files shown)


mmatera:
OK, I like this, but actually, to improve the axis labels the best would be to implement the Align property, in a way that the labels of the y-axis be aligned to the right, and the labels of the x axis be aligned to the top.

rocky:
> OK, I like this, but actually, to improve the axis labels the best would be to implement the Align property, in a way that the labels of the y-axis be aligned to the right, and the labels of the x axis be aligned to the top.

Absolutely, but let me give the broader context around why I did this and did this now. 

At this point, I'd like to have a release in about 24hours and I am trying to understand what wil make the cut and what won't.

I was borderline on this, until I spent some time with the code to glean how much of a mess we woudl have to undo in the future, how ugly it looks, and the benefit. With the merge in (thank you), I think this should go in. I would like some of the commits (like the one where you are just removing print statements that in my opinion shouldn't have been commited in the first place) squashed. 

* #1216 I think is okay to go in, right? 
* #1215 I think is sort of okay, but I'd like a version of CellsToTeX in packages even if not all of it works. The author says that's okay. If you are not up for doing the work, I'll shepard this part in
* #1213 I guess is okay, right? 
* #1212 iis okay, right? 
* 1217 is ready? 

The PRs labled "needs work" I would like deferred
* #1222 
* #1214 
* #1199

These appear to as brorken: 

* #1220 
* #1132 (and those with numbers lower than that likewise)

These I would like deferred to another release
* #1178
* #1209

Since there is already a lot of changes here including some big ones, if there is _anything_ or can be deferred I would prefer they get deferred. I would rather have two smaller releases than one big one.

Your thoughts?

mmatera:
> 
> @mmatera At this point, I'd like to have a release in about 24hours and I am trying to understand what will make the cut and what won't.
> 

Completely agree. I think with the transparency it actually looks nicer...

> I was borderline on this, until I spent some time with the code to glean how much of a mess we would have to undo in the future, how ugly it looks, and the benefit. With the merge in (thank you), I think this should go in. I would like some of the commits (like the one where you are just removing print statements that in my opinion shouldn't have been commited in the first place) squashed.

I guess I did it the squash...

Regarding the rest:

#1216
#1213
#1212
#1217
are ready. #1215 is ready, in the sense that the test is ready. If you can also ship the package, go ahead, but still, I do not have any subset of the package working, just tests that eventually will detect when the package does something interesting.

I would also keep #1209, because it actually improves a lot the support of the kernel in comparisons. I guess it could be a little bit better implemented, but not a lot. 

#1199  collect some of the previous PRs, so you do not need to merge it.

mmatera:
Regarding #1220, this is something that we should fix as soon as possible after we finish with the release since many other things depend on the ability to handle derivatives.

rocky:
> > @mmatera At this point, I'd like to have a release in about 24hours and I am trying to understand what will make the cut and what won't.
> 
> Completely agree. I think with the transparency it actually looks nicer...
> 
> > I was borderline on this, until I spent some time with the code to glean how much of a mess we would have to undo in the future, how ugly it looks, and the benefit. With the merge in (thank you), I think this should go in. I would like some of the commits (like the one where you are just removing print statements that in my opinion shouldn't have been commited in the first place) squashed.
> 
> I guess I did it the squash...
> 
> Regarding the rest:
> 
> #1216
> #1213
> #1212
> #1217
> are ready. 

Merged now.

#1215 is ready, in the sense that the test is ready. If you can also ship the package, go ahead, but still, I do not have any subset of the package working, just tests that eventually will detect when the package does something interesting.
> 

I just looked at in more detail. #1215 not ready. 

> I would also keep #1209, because it actually improves a lot the support of the kernel in comparisons. I guess it could be a little bit better implemented, but not a lot.

I am not going to have time to go over #1209. From what I saw before it needs a refactor to make it more scalable, and at the same time we can increase the number of passing test cases.

So let's defer and after the release, let's work on this. 

I will add a section in the release notes what's will be coming soon. 


> 
> #1199 collect some of the previous PRs, so you do not need to merge it.

It has been closed.

PR #1225, opened 2021-04-04
Misc small changes

rocky:
* Nicer Plot3D example
* Add description of MaxRecursion
* Note Plot3D options


mathics/main.py
```diff
@@ -242,6 +242,7 @@ def main() -> int:
 
     argparser.add_argument(
         "--execute",
+        "-c",
```
mmatera, line 245: In WolframScript, "-c" has the meaning of '-e'

mmatera (approved):

mathics/main.py
rocky: Ok, removed then. I am sure though I saw someone use "-c" some place.

PR #1228, opened 2021-04-05
Support Table[expr, n]

rocky:


rocky:
> Apart from the need of fixing the test in `Information`, this LGTM.

It will be on me to come up with a _unit_ test to test this. 

I tried earlier but the subtleties of getting formatting to evaluation were eluding me in a unit test.

As a doc test, we have seen that this test is very fragile in that the text has to match.

However the assumed empty lines with blanks has been a constant source of annoyance.

mmatera:
> > Apart from the need of fixing the test in `Information`, this LGTM.
> 
> It will be on me to come up with a _unit_ test to test this.
> 
> I tried earlier but the subtleties of getting formatting to evaluation were eluding me in a unit test.
> 
> As a doc test, we have seen that this test is very fragile in that the text has to match.
> 
> However the assumed empty lines with blanks has been a constant source of annoyance.

Yes, I have faced the same problem several times. Doctests check the output in a way that the "expected" output is the output in the (text only) front-end, while pytest checks the output as Python objects. The last approach has the problem that does not check the messages and print strings, just the output. I would like to have all the test engines based on MathicsSession, but then we need to take some definitions in its design: when "evaluation" returns a Python object and when a formated string, and how the messages are exposed.

For this, if you  want, just comment out the broken test, or add the corresponding lines in the Information Doctest (it is not a very drastic change). Then, we can try to work out the global problem in another PR.

Issue #90, opened 2013-02-05
Attributes for Predefined symbols


sn6uv:
When the attributes of a `Predefined` symbol are left undefined (or are defined as an empty tuple) this seems to get overridden with `{Protected}`. See, for example:

```
>> Attributes[$InstallationDirectory]
 = {Protected}
```

which should give 

```
In[1]:= Attributes[$InstallationDirectory]
Out[1]= {}
```

despite being defined as such in `builtin.files`.

Related: #85

tueda:
Any updates? Some existing Mathematica packages may try to overwrite `$Path` for adding their subdirectories and fail due to this bug.

rocky:
When not in Django, this works.

PR #1229, opened 2021-04-05
Unprotect $InstallationDirectory

rocky:
Note: Unprotect doesn't seem to work for $Path, $InstallationDirectory,
Makeboxes if not others in Django. What's up here?

But let's get this in and continue from here. 

See #90


mmatera (approved):
OK, I do not see why ``$InstallationDirectory`` should be modifiable, but it is in WMA.

Issue #1235, opened 2021-04-06
Arguments of FindRoot are not evaluated


soehms:
Using predefined arguments gives:

```
In[4]:= v1 := Exp[x] - 3x == 0
In[5]:= v2 := {x, 2}
In[6]:= FindRoot[v1, v2]
Out[6]= FindRoot[v1, v2]
```
Using explicit arguments works:

```
In[7]:= FindRoot[Exp[x] - 3x == 0, {x, 2}]
Out[7]= {x -> 1.51213}
```

Shouldn't `Out[6]` be the same as `Out[7]`?

rocky:
#1236 starts to address some of this, but unless someone else gets interested in it (@soehms maybe you can start here and work more on it?), from my standpoint this will have to wait until what is needed for FeynCalc to work is more complete, since that probably will have more of a public impact. 

But here is what is going on. 

Currently the function form that is matched is `FindRoot[f_, {x_, x0_}]`  which is expecting the second argument to be written out as a list of two items.  So it is not so much an _evaluation_ problem so much as a pattern matching problem which is causing the code to handle `FindRoot[]` not to get triggered. Or rather that is just _one_ of the problems. 

In the FindRoot-expansion` branch I created to start addressing this, I added code to match the situation here where the second parameter is a variable containing a list. 

However there is transformation rule that looks like this: `FindRoot[lhs_ == rhs_, {x_, xs_}]": "FindRoot[lhs - rhs, {x, xs}]`.

In other words, a function whose top-level operation is `Equal[]` in them are changed into  subtraction.

The code I wrote detects that but isn't doing the transformation in code properly. It might be a mistake I've made. But as I said, that someone else should look at and move forward at some point. 

So to sum up... 

```
n[4]:= v1 := Exp[x] - 3x  (* no: "== 0" which is the same thing *)
In[5]:= v2 := {x, 2}
In[6]:= FindRoot[v1, v2]
```
now works. Although adding the `== 0` doesn't yet.  A different error, a recursion error, is given for this.

rocky:
I should also mention that mpmath also provides a `findroot()` function and it might be that the iterative code we have here.
Also [sympy's nsolve seems to use the mpmath findroot](https://docs.sympy.org/latest/modules/solvers/solvers.html). We may hook into that via the Mathics `Solve` function.

mmatera:
[...]

PR #1236, opened 2021-04-06
Start to more Handle FindRoot forms...

rocky:
In particular where the second parameter is a variable containing a list of two items, _x_,  _x0_. 

We are still failing when the 1st parameter is a variable expression involving `Equal[]`.

See #1235

### DIFF
--- a/test/test_calculus.py
+++ b/test/test_calculus.py
@@ -24,5 +24,10 @@ def test_calculus():
             "{{a -> 1}}",
             "Issue #1168",
         ),
+        (
+            "v1 := Exp[x] - 3x; v2 = {x, 2}; FindRoot[v1, v2]",
+            "{x->1.51213}",
+            "Issue #1235",
+        ),
     ):
         check_evaluation(str_expr, str_expected, message)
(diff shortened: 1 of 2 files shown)


mmatera (approved):
LGTM

test/test_calculus.py
```diff
@@ -24,5 +24,10 @@ def test_calculus():
             "{{a -> 1}}",
             "Issue #1168",
         ),
+        (
+            "v1 := Exp[x] - 3x; v2 = {x, 2}; v2 = {x, 2}; FindRoot[v1, v2]",
```
mmatera, line 28: why v2 is defined two times?

rocky: This was a cut and paste typo. Should be fixed in  98221899

rocky:
BTW this doesn't completely address #1235

mmatera:
OK, then let's merge this...

mmatera:
> BTW this doesn't completely address #1235

OK, but it is in the right direction. On the other hand, we should try this now, after the implementation of the other changes in Set and Equal

PR #1237, opened 2021-04-07
Series

### LINKED ISSUE #1193: Add WL function Series[]
* [WL Series](https://reference.wolfram.com/language/ref/Series.html)
* [Sympy Series](https://docs.sympy.org/latest/modules/core.html?highlight=series#sympy.core.expr.Expr.series)

mmatera:
This is a first attempt to fix #1193 

Includes both #1220 #1233


mathics/builtin/calculus.py
```diff
@@ -314,6 +392,7 @@ def __init__(self, *args, **kwargs):
         super(Derivative, self).__init__(*args, **kwargs)
 
     def to_sympy(self, expr, **kwargs):
+        print("calling to_sympy")
```
rocky, line 395: We probably want to remove here. 

Ideally I'd like to figure out a way where we could not have our commit history fill of this is the place where we removed the debug print. Squashing commits might be one way.

mathics/builtin/calculus.py
```diff
@@ -273,9 +348,12 @@ class Derivative(PostfixOperator, SympyFunction):
             r'    RowBox[{"(", Sequence @@ Riffle[{n}, ","], ")"}]]]]'
         ),
         "MakeBoxes[Derivative[n:1|2][f_], form:OutputForm]": """RowBox[{MakeBoxes[f, form], If[n==1, "'", "''"]}]""",
+        # The following rules should be applied in the apply method, instead of relying on the pattern matching
+        # mechanism.
         "Derivative[0...][f_]": "f",
         "Derivative[n__Integer][Derivative[m__Integer][f_]] /; Length[{m}] "
         "== Length[{n}]": "Derivative[Sequence @@ ({n} + {m})][f]",
+        # This would require at least some comments...
```
rocky, line 356: Ok - I guess this isn't finished? Or haven't decided what to do?

mathics/builtin/calculus.py
mmatera: I squashed this, but it seems that this one escaped to my sight... In any case, I found that this function is never called, except in some tests...

mathics/builtin/calculus.py
mmatera: Here I am not sure what to do. As it is, it works with the functions I imagined it makes sense to try, but I would like to have something clearer here. Also, I would like to avoid relying on the pattern matching mechanism when it could be handled by a more or less concise code.

mmatera: By now, and to avoid making this PR too fat, I would merge this as it is, and then continue making improvementsments in another PR.

PR #1238, opened 2021-04-07
Knot theory fixes

mmatera:
Here I included the package KnotTheory from [http://katlas.org/wiki/Setup](http://katlas.org/wiki/Setup). Not sure if we want to ship this with Mathics, or just add a link, but I made this to do tests.


soehms:
> @soehms, please contact the authors to see if they have any problem with having a copy of the code here, or if they want to have a more detailed mention in the credits.

Sure, I will take that! Many thanks again, for working on this!

soehms:
> @soehms, please contact the authors ...

I have received an answer from Dror Bar-Natan, right now:

> to see if they have any problem with having a copy of the code here

There is no problem!

> or if they want to have a more detailed mention in the credits

Just include a link to http://katlas.org.

Also, he would agree if KnotTheory were to be a Mathics package.

PR #1241, opened 2021-04-08
adding  ``$BaseDirectory`` and ``$UserBaseDirectory`` . Required by Feyncalc

mmatera:
By default, I set ``$BaseDirectory`` equal to ``$InstallationDirectory``, and ``$UserBaseDirectory`` as ``~/.mathics``. Maybe we need to ensure that such a directory exists when a ``Definitions`` is created. On the other hand, maybe we could choose a better place for this configuration directory.


rocky (approved):
Seems straightforward to me.

Issue #1234, opened 2021-04-06
Read[StringToStream["..."], Hold[Expression]]


rocky:
Split off from #1174

* https://github.com/mathics/Mathics/issues/1174#issuecomment-806073268
* https://github.com/mathics/Mathics/issues/1174#issuecomment-806604428

@mmatera is this something you are going to handle? If not, let me know and I will look at it.

This seems to be the next main obsticle in getting #1174 working.

mmatera:
I am trying to fix the issue with the ``Derivative``s.  If you want, go ahead!

vsht:
Let me know once you have branch I can test my codes against.

mmatera:
Addressed by #1242.

rocky:
Think this is addressed. If not reopen with comments.

PR #1242, opened 2021-04-08
Processing comments in Read Expression

mmatera:
Just removes the warning when Read receives a comment without an expression.


mmatera:
Address #1234

PR #1246, opened 2021-04-09
Sparsearray

mmatera:
Basic implementation of SparseArray

TODO:
* Operations between sparse objects.
* Conversion to the format for numeric libraries.
* Optimize the representation of the non-vanishing elements.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,7 @@ New builtins
 * ``NIntegrate``
 * ``PartitionsP``
 * ``$Notebooks``
+* ``SparseArray``
 
 Enhancements
 ++++++++++++
--- a/mathics/builtin/sparse.py
+++ b/mathics/builtin/sparse.py
@@ -1,0 +1,205 @@
+# -*- coding: utf-8 -*-
+
+"""
+SparseArray Functions
+"""
+
+
+from mathics.version import __version__  # noqa used in loading to check consistency.
+from mathics.builtin.lists import walk_parts
+
+from mathics.builtin.base import (
+    Builtin,
+    Test,
+    InvalidLevelspecError,
+    BinaryOperator,
+    PartError,
+    PartDepthError,
+    PartRangeError,
+    Predefined,
+    SympyFunction,
+)
+from mathics.builtin.scoping import dynamic_scoping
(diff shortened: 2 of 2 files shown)


rocky:
@mmatera I would like some context here. Is this necessary for Feyncalc or knot theory? 

I look at Scipy and see they have a [sparse arrays (many kinds)](https://docs.scipy.org/doc/scipy/reference/sparse.html) already, and I wonder if just hooking them in wouldn't be a better start.

mmatera:
> @mmatera I would like some context here. Is this necessary for Feyncalc or knot theory?
> 
> I look at Scipy and see they have a [sparse arrays (many kinds)](https://docs.scipy.org/doc/scipy/reference/sparse.html) already, and I wonder if just hooking them in wouldn't be a better start.

I started with this because is necessary for implementing CoefficientArray, which is necessary for feyncalc,and several other things. Also, SparseArray is a very important pattern for many other stuff, like eigensolvers.

mmatera:
Regarding on the proposed implementation, what we need to change if we want to delegate the manipulation to sympy is the format of the data block, which could be implemented as an InstanceableBuiltin.

rocky:
@mmatera Thanks for the context, background and information. 

Some broad general thoughts, but I leave the ultimate decision of how to handle to you. 

It feels to me that too often we are rushing things when we need not. I am definitely guilty of this sometimes, because I just often don't have time. So what you see is that I'll leave unfinished PR's out there for a while. (It doesn't bother me though.) 

And sometimes I do commit things that are minimal. 

But here, this feels like and important thing to get sort of right. And if it is not too difficult to do in the way that we will ultimately use, that in my opinion we should just spend the extra time and start out with the scipy sparse arrray, (even if it is a mimimal use).

mmatera:
This is a very basic implementation, that helps to move forward. Much of this PR is about the interface, and that shouldnt change in the near future. What it can be improved, and what is needed for implementing operations with sparse objects, is to improve the storage (last element) of the SparseArray object. This storage now is implemented as a list of rules, which was the way in which wma did this in earlier versions. In the future, the storage should be implemented in a way more friendly to the libraries which should carry most of the computing.

PR #1247, opened 2021-04-10
Improve ToExpression parsing of multi-line input

### LINKED ISSUE #1245: Multi-line string parsing inside `ToExpression[x, InputForm]`
It appears that Mma is somewhat more fault tolerant when it comes to missing  semicolons, cf. e.g. the following
admittedly sloppy code

```
foo::usage =
"foo is a foo."

bar::usage =
"bar is a bar."

Begin["`Package`"]

End[]

Begin["`Private`"]

Options[foo] = {
  "FooOpt" ->  True
}

foo[x_]:=x^2;
```
Despite of the semicolons missing at many places, Mma obviously figures out that a CR accounts for a semicolon here. So it happily accepts
```
myFileAsString = Import["testFile.m", "Text"];   
ToExpression[myFileAsString] 
```
Mathics doesn't
```
In[1]:= myFileAsString = Import["testFile.m", "Text"];   
In[2]:= ToExpression[myFileAsString] 
SetDelayed::write: Tag Times in {FooOpt -> True} foo[x_] is Protected.
Set::write: Tag Times in bar is a bar. Begin[`Package`] End[] Begin[`Private`] Options[foo] is Protected.
Set::write: Tag Times in foo is a foo. bar::usage is Protected.
In[3]:= 
```
For compatibility with other packages it would be good to emulate Mma's behavior.  Cf. also Issue #1174 

[testFile.zip](https://github.com/mathics/Mathics/files/6287487/testFile.zip)

rocky:
Fixes #1245

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,8 +7,9 @@ CHANGES
 Enhancements
 ++++++++++++
 
-* ``Table`` [*expr*, *n*] is now supported.
-* ``ToString`` now excepts the *Form* form.
+* ``Table`` [*expr*, *n*] is supported.
+* ``ToString`` accepts an optional *form* parameter.
+* ``ToExpression`` handles multi-line string input
 
 Bug fixes
 +++++++++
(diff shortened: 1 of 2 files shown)


mmatera (approved):
LGTM.

PR #1248, opened 2021-04-10
Handle Compress/Uncompress checks better..

rocky:
That is, in a way that doesn't depend on OS characteristics or
specifics of how Compress compresses things.

Simplify test/helper.py and remove unused imports.

BTW: I think ToString should be the exception not the rule.
In other words the defaults to_string_expr and to_string_expected should
be opposite of what it currently is.

### DIFF
--- a/test/helper.py
+++ b/test/helper.py
@@ -1,18 +1,30 @@
+# -*- coding: utf-8 -*-
 import time
-from mathics.core.parser import parse, MathicsSingleLineFeeder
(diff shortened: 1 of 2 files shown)


rocky:
@mmatera there are a series of cleanups of the code that I would like to do in preparation of a bigger refactor of the Equal (==). 

If possible, it would be good to get these cleanups out of the way quicker.

mmatera (approved):

PR #1253, opened 2021-04-10
same -> sameQ; do_compare -> equal2

rocky:
* same -> sameQ with docstring and return value.
* do_compare -> equal2 (two argument Equal)
* sameQ properly implemented in the base class.

This is the last of the small cleanups before a more major refactor.


mmatera:
OK, makes a lot of sense..

PR #1265, opened 2021-04-11
Simplify Equal by adding class Equal  (equal2, ...) methods

rocky:
A lot of refactoring has gone on here.


rocky:
I think what would further simplify things is to store in the various classes that implement an `equal2()` method a priority for using that method. 

For some types, comparison can be simple because it is very restricted. In particular Compiled code is like that.

mmatera (commented):
So, shall we merge fixcomparison-modularize here, and then merge fixcomparison to master? Or we can just merge fixcomparison-modularize  (maybe after a rebase) directly to master?

Issue #1258, opened 2021-04-11
Routines for determining the memory usage


vsht:
Mathics currently misses some useful functions that allow to estimate the amount of RAM
available on the current machine and the memory usage. These are: [MemoryAvailable](https://reference.wolfram.com/language/ref/MemoryAvailable.html), [MemoryInUse](https://reference.wolfram.com/language/ref/MemoryInUse.html), [MaxMemoryUsed](https://reference.wolfram.com/language/ref/MaxMemoryUsed.html) and $SystemMemory (apparently not documented, here is the output for my machine with 32 GB RAM. It basically just returns the total amount of RAM in bytes)

```
In[1]:= ?$SystemMemory                                                                                                                       

Out[1]= System`$SystemMemory
        $SystemMemory = 32910770176


        Attributes[$SystemMemory]={Locked, Protected}
```

This is somewhat relevant for #1174 although not critical.

rocky:
The [psutil module](https://pypi.org/project/psutil/) seems to be what would be used in Python to get this information.

rocky:
Should be addressed via #1345. If not, please just reopen this issue with additional details.

PR #1270, opened 2021-04-12
memory functions

mmatera:
This tries to address in a very basic level #1258


mathics/builtin/system.py
```diff
@@ -455,3 +455,121 @@ def evaluate(self, evaluation) -> Real:
...
+     = ...
+    """
+
+    name = "$SystemMemory"
+    
+    def evaluate(self, evaluation) -> Integer:
+        try:
+            import psutil
+            totalmem = psutil.virtual_memory().total
+            return Integer(total)
```
rocky, line 477: @mmatera there is a typo here and below. I believe `total` should be `totalmem` which is set on line 476. 

I will be addressing this in another PR.

Issue #1231, opened 2021-04-05
Implemen `ConditionalExpression[]`


rocky:
https://reference.wolfram.com/language/ref/ConditionalExpression.html

This is mentioned in passing in Issue #62

mmatera:
This should be partially addressed by PR #1271. Still, we need a way to implement some kind of predicate reduction using assumptions.

PR #1271, opened 2021-04-12
basic implementation of ``ConditionalExpression``, ``Assuming`` and ``$Assumptions``

mmatera:
In this PR, I addressed issue #1231, together with a skeleton for handling ``Assuming`` and ``$Assumptions``. `ConditionalExpression` allows a method `to_sympy`, which translates it to a `Piecewise` expression in `sympy`.
``$Assumptions`` are not currently taken into account in the evaluations, but it can be hooked in the future in the implementation of ``Simplify``, ``Reduce``, ``Integrate``, and other symbols.


rocky:
@mmatera I am looking forward to reviewing this in detail, but it will be maybe towards the end of the week when I can look at and understand.

rocky (commented):
I spent some time trying ConditionalExpression, Assuming and $Assumptions, and aside from the words getting recognized, I couldn't make this do anything that seemed useful. 

I will put in a PR into this branch for mostly small doc-change stuff. 

My take is merge this when it does something useful, and for that probably just hooking this into Piecewise evaluation would do it.

mathics/builtin/arithmetic.py
```diff
@@ -2205,3 +2212,131 @@ def apply(self, expr, evaluation):
...
+        elif cond.is_symbol() or not cond.has_form("List", None):
+            cond = [cond]
+        else:
+            cond = cond.leaves
+        assumptions = evaluation.definitions.get_definition(
+            "System`$Assumptions", only_if_exists=True
+        )
+
+        if assumptions:
+            assumptions = assumptions.ownvalues
```
rocky, line 2261: Personally I don't like the fact that assumptions changes datatypes here. It starts out as a definition, and the switches to something that is not a definition. I'll put in a PR into this branch to separate variables with different types.  When this kind of thing is done, I find it harder to follow the code.

mathics/builtin/arithmetic.py
```diff
@@ -2205,3 +2212,131 @@ def apply(self, expr, evaluation):
...
+            if len(assumptions) > 0:
+                assumptions = assumptions[0].replace
+            else:
+                assumptions = None
+        if assumptions:
+            if assumptions.is_symbol() or not assumptions.has_form("List", None):
+                assumptions = [assumptions]
+            else:
+                assumptions = assumptions.leaves
+            cond = assumptions + tuple(cond)
```
rocky, line 2271: When I tried this, i was getting an error because aussumptions is sometimes a list such as in line 2268 and sometimes it is a tuple. 

In the PR I will put in I changed added a list cast on 2270, although there are other ways to do this.

mathics/builtin/arithmetic.py
```diff
@@ -2205,3 +2212,131 @@ def apply(self, expr, evaluation):
...
+        Expression(
+            "Set", Symbol("System`$Assumptions"), Expression("List", *cond)
```
rocky, line 2273: Here is what I am seeing with this code:

```
In[1]:= $Assumptions
Out[1]= True

In[2]:= $Assumptions = x > 2
Out[2]= x > 2

In[3]:= $Assumptions
Out[3]= x > 2

In[4]:= Assuming[-10 < n < 10, Integrate[x^n, {x, 0, 1}]]
Out[4]= Piecewise[{{1 / (1 + n) - 0 ^ (1 + n) / (1 + n), n > -Infinity && n < Infinity && n != -1}, {Infinity, True}}]

In[5]:= $Assumptions
Out[5]= {x > 2}
```
Note that the type of `$Assumptions` changes, and the default `True` quickly gets changed to `{True}`.

But I think there is a deeper problem in that, if we ever support concurrency, then saving and restoring a variable will break things.

Is there is a way to wrap this in a context, so the scope of `$Assumptions` is narrowed to the expression getting evaluated and nothing needs to be saved and restored?

mmatera: One idea I had to handle this is to set a property in the  `Evaluation` object, or in `Definitions` object, each time $Assumptions is modified, or when Assuming is called. In any case, before looking at the best implementation, a way to evaluate predicates on expressions should be implemented.

mmatera: Another option to implement the temporal switching is to use `dynamic_scoping` (defined in mathics.builtin.scoping) as in `FindRoot`. This would make the code at least more readable.

mmatera:
I think this is ready now: the function `predicate_evaluation(predicate, evaluation)` looks at $Assumptions and implement some replacements on predicate accordingly. This function is called then in the evaluation of `Piecewise`, and `ConditionalExpression`,  in a way that propagates automatically to `Integrate`.

mathics/builtin/arithmetic.py
```diff
@@ -2283,84 +2292,107 @@ class Assuming(Builtin):
...
+     = ConditionalExpression[x ^ 2 y, y > 0]
     """
 
     attributes = ("HoldRest",)
 
-    def apply_assuming(self, cond, expr, evaluation):
-        "Assuming[cond_, expr_]"
-        cond = cond.evaluate(evaluation)
-        if cond.is_true():
+    def apply_assuming(self, asumptions, expr, evaluation):
```
rocky, line 2309: A superficial thing noticed in a casual reading: asumptions -> assumptions.

mathics/builtin/calculus.py
```diff
@@ -604,11 +607,19 @@ def apply(self, f, xs, evaluation):
...
         if prec is not None and isinstance(result, sympy.Integral):
             # TODO MaxExtaPrecision -> maxn
             result = result.evalf(dps(prec))
-        result = from_sympy(result)
+        else:
+            result = from_sympy(result)
+        # If the result is defined as a Piecewise expression,
+        # use ConditionalExpression.
+        # This does not work now because the form sympy returns the values
+        if result.get_head_name() == "System`Piecewise":
```
rocky, line 618: Another small thing. I think the idiom typically used is `result.is_form()` right?

mathics/builtin/logic.py
```diff
@@ -3,7 +3,70 @@
...
+from mathics.core.rules import Rule
+
+from typing import Optional, Union, Any
+
+
+def evaluate_predicate(pred, evaluation):
+    assumptions_list = get_assumptions_list(evaluation)
+    assumption_rules = []
+    for pat in assumptions_list:
+        if pat.head == Symbol("Not"):
```
rocky, line 24: is_form again here?

mathics/builtin/arithmetic.py
```diff
@@ -2283,84 +2292,107 @@ class Assuming(Builtin):
...
     <dt>'Assuming[$cond$, $expr$]'
-      <dd>Evaluates $expr$ assuming the conditions $cond$
+      <dd>Evaluates $expr$ assuming the conditions $cond$.
     </dl>
     >> $Assumptions = { x > 0 }
      = {x > 0}
-    >> Assuming[y>0, $Assumptions]
-     = {x > 0, y > 0}
+    >> Assuming[y>0, ConditionalExpression[y x^2, y>0]]
+     = x ^ 2 y
```
rocky, line 2300: Nice!

mathics/builtin/calculus.py
```diff
@@ -509,14 +509,17 @@ class Integrate(SympyFunction):
 
     >> Integrate[f'[x], {x, a, b}]
      = f[b] - f[a]
+    >> Integrate[x/Exp[x^2/t], {x, 0, Infinity}]
+     = Piecewise[{{t / 2, Abs[Arg[t]] < Pi / 2}, {Integrate[x E ^ (-x ^ 2 / t), {x, 0, Infinity}], True}}]
+    >> Assuming[Abs[Arg[t]] < Pi / 2, Integrate[x/Exp[x^2/t], {x, 0, Infinity}]]
+     = t / 2
```
rocky, line 515: Nice!

rocky:
> I think this is ready now: the function `predicate_evaluation(predicate, evaluation)` looks at $Assumptions and implement some replacements on predicate accordingly. This function is called then in the evaluation of `Piecewise`, and `ConditionalExpression`, in a way that propagates automatically to `Integrate`.

A superficial look through and I think this is exactly the right thing to do. However I want to pull this down and try it and maybe even experiment with the code to understand what's up with sympy mentioned in one of the comments.

mathics/builtin/arithmetic.py
mmatera: Shame on me...

mathics/builtin/calculus.py
mmatera: Just to have in mind, has_form is a method of `Expression`, so to use it, we need to check first that the object is not an atom...

mathics/builtin/logic.py
mmatera: the same: we do not know a priori if pat is a Symbol or an Expression. This is the safer and simpler way to do the comparison.

mathics/builtin/calculus.py
mmatera: even checking with is_atom() the method fails for strings. For this reason, I reverted this to the previous strategy. Maybe later a better approach comes up.

rocky: Here is a suggestion: add the `is_form()`  method to Atom. The implementation is pretty simple: return False.

mmatera:
> > I think this is ready now: the function `predicate_evaluation(predicate, evaluation)` looks at $Assumptions and implement some replacements on predicate accordingly. This function is called then in the evaluation of `Piecewise`, and `ConditionalExpression`, in a way that propagates automatically to `Integrate`.
> 
> A superficial look through and I think this is exactly the right thing to do. However I want to pull this down and try it and maybe even experiment with the code to understand what's up with sympy mentioned in one of the comments. And probably the weekend is when I'll have an opportunity to do.

Good. So, I made the corrections that you mention above, and I have rebased to master. Now I leave this here until you look at this. I will continue with other issues then.

mathics/builtin/arithmetic.py
```diff
@@ -2189,15 +2195,18 @@ class Piecewise(SympyFunction):
 
     def apply(self, items, evaluation):
         "%(name)s[items__]"
-        result = self.to_sympy(Expression("Piecewise", *items.get_sequence()))
+        result = self.to_sympy(
+            Expression("Piecewise", *items.get_sequence()), evaluation=evaluation
```
rocky, line 2199: Great! This is what I was expecting to see!

rocky (approved):
Just had a chance to go over in detail and here's what I encountered. 

```
Assuming[0 < n < 1, Integrate[x^n, {x, 0, 1}]]
Out[3]= Piecewise[{{1 / (1 + n) - 0 ^ (1 + n) / (1 + n), n > -Infinity && n < Infinity && n != -1}, {Infinity, True}}]

In[4]:= Assuming[n = 1, Integrate[x^n, {x, 0, 1}]]
Out[4]= 1 / 2

In[5]:= Assuming[n = 2, Integrate[x^n, {x, 0, 1}]]
Out[5]= 1 / 3

In[6]:= Assuming[Or[n = 2, n = 1] , Integrate[x^n, {x, 0, 1}]]
Out[6]= 1 / 2

In[7]:= Assuming[Or[n = 1, n = 2] , Integrate[x^n, {x, 0, 1}]]
Out[7]= 1 / 3
```

If you want to call this a current limitation, ok. If you'd like to explain why we get what we get, I'd appreciate it. 

I leave to you whether we want to fix or merge and fix later. 

In f12c348f I removed an unused import. See also #1313 if these are tweaks you want to add.

mmatera:
> Just had a chance to go over in detail and here's what I encountered.
> 
> ```
> Assuming[0 < n < 1, Integrate[x^n, {x, 0, 1}]]
> Out[3]= Piecewise[{{1 / (1 + n) - 0 ^ (1 + n) / (1 + n), n > -Infinity && n < Infinity && n != -1}, {Infinity, True}}]
> 
> In[4]:= Assuming[n = 1, Integrate[x^n, {x, 0, 1}]]
> Out[4]= 1 / 2
> 
> In[5]:= Assuming[n = 2, Integrate[x^n, {x, 0, 1}]]
> Out[5]= 1 / 3
> 
> In[6]:= Assuming[Or[n = 2, n = 1] , Integrate[x^n, {x, 0, 1}]]
> Out[6]= 1 / 2
> 
> In[7]:= Assuming[Or[n = 1, n = 2] , Integrate[x^n, {x, 0, 1}]]
> Out[7]= 1 / 3
> ```
> 
> If you want to call this a current limitation, ok. If you'd like to explain why we get what we get, I'd appreciate it.
> 
> I leave to you whether we want to fix or merge and fix later.
> 
> In [f12c348](https://github.com/mathics/Mathics/commit/f12c348f8f361efec4afd7e278c6e50d5316cff2) I removed an unused import. See also #1313 if these are tweaks you want to add.

The first output is just wrong, but is wrong due an issue in sympy. The problem we have here is the format of the sympy return: we get something that is a piecewise definition, with the last (default) value being the unevaluated integral, but not exactly in the same form. So, there is no safe way to reduce this to a ConditionalExpression, without performing a more or les extense set of tests in the default value...

rocky:
> > Just had a chance to go over in detail and here's what I encountered.
> > ```
> > Assuming[0 < n < 1, Integrate[x^n, {x, 0, 1}]]
> > Out[3]= Piecewise[{{1 / (1 + n) - 0 ^ (1 + n) / (1 + n), n > -Infinity && n < Infinity && n != -1}, {Infinity, True}}]
> > 
> > In[4]:= Assuming[n = 1, Integrate[x^n, {x, 0, 1}]]
> > Out[4]= 1 / 2
> > 
> > In[5]:= Assuming[n = 2, Integrate[x^n, {x, 0, 1}]]
> > Out[5]= 1 / 3
> > 
> > In[6]:= Assuming[Or[n = 2, n = 1] , Integrate[x^n, {x, 0, 1}]]
> > Out[6]= 1 / 2
> > 
> > In[7]:= Assuming[Or[n = 1, n = 2] , Integrate[x^n, {x, 0, 1}]]
> > Out[7]= 1 / 3
> > ```
> > 
> > 
> > If you want to call this a current limitation, ok. If you'd like to explain why we get what we get, I'd appreciate it.
> > I leave to you whether we want to fix or merge and fix later.
> > In [f12c348](https://github.com/mathics/Mathics/commit/f12c348f8f361efec4afd7e278c6e50d5316cff2) I removed an unused import. See also #1313 if these are tweaks you want to add.
> 
> The first output is just wrong, but is wrong due an issue in sympy. The problem we have here is the format of the sympy return: we get something that is a piecewise definition, with the last (default) value being the unevaluated integral, but not exactly in the same form. So, there is no safe way to reduce this to a ConditionalExpression, without performing a more or les extense set of tests in the default value...

And what about the last two? "Or" to me doesn't mean that you can pick which ever one you want, but that both may be true. Note that saying "And" here would be wrong since `x` can't be two values at once.

mmatera:
> > > Just had a chance to go over in detail and here's what I encountered.
> > > ```
> > > Assuming[0 < n < 1, Integrate[x^n, {x, 0, 1}]]
> > > Out[3]= Piecewise[{{1 / (1 + n) - 0 ^ (1 + n) / (1 + n), n > -Infinity && n < Infinity && n != -1}, {Infinity, True}}]
> > > 
> > > In[4]:= Assuming[n = 1, Integrate[x^n, {x, 0, 1}]]
> > > Out[4]= 1 / 2
> > > 
> > > In[5]:= Assuming[n = 2, Integrate[x^n, {x, 0, 1}]]
> > > Out[5]= 1 / 3
> > > 
> > > In[6]:= Assuming[Or[n = 2, n = 1] , Integrate[x^n, {x, 0, 1}]]
> > > Out[6]= 1 / 2
> > > 
> > > In[7]:= Assuming[Or[n = 1, n = 2] , Integrate[x^n, {x, 0, 1}]]
> > > Out[7]= 1 / 3
> > > ```
> > > 
> > > 
> > > If you want to call this a current limitation, ok. If you'd like to explain why we get what we get, I'd appreciate it.
> > > I leave to you whether we want to fix or merge and fix later.
> > > In [f12c348](https://github.com/mathics/Mathics/commit/f12c348f8f361efec4afd7e278c6e50d5316cff2) I removed an unused import. See also #1313 if these are tweaks you want to add.
> > 
> > 
> > The first output is just wrong, but is wrong due an issue in sympy. The problem we have here is the format of the sympy return: we get something that is a piecewise definition, with the last (default) value being the unevaluated integral, but not exactly in the same form. So, there is no safe way to reduce this to a ConditionalExpression, without performing a more or les extense set of tests in the default value...
> 
> And what about the last two? "Or" to me doesn't mean that you can pick which ever one you want, but that both may be true. Note that saying "And" here would be wrong since `x` can't be two values at once.

You are right: to assume  `Or[a==1,  a == 2]` is `True` should be equivalent to assume that `And[a!=1, a!= 2]` is `False`, isn't it?

rocky:
> > > > Just had a chance to go over in detail and here's what I encountered.
> > > > ```
> > > > Assuming[0 < n < 1, Integrate[x^n, {x, 0, 1}]]
> > > > Out[3]= Piecewise[{{1 / (1 + n) - 0 ^ (1 + n) / (1 + n), n > -Infinity && n < Infinity && n != -1}, {Infinity, True}}]
> > > > 
> > > > In[4]:= Assuming[n = 1, Integrate[x^n, {x, 0, 1}]]
> > > > Out[4]= 1 / 2
> > > > 
> > > > In[5]:= Assuming[n = 2, Integrate[x^n, {x, 0, 1}]]
> > > > Out[5]= 1 / 3
> > > > 
> > > > In[6]:= Assuming[Or[n = 2, n = 1] , Integrate[x^n, {x, 0, 1}]]
> > > > Out[6]= 1 / 2
> > > > 
> > > > In[7]:= Assuming[Or[n = 1, n = 2] , Integrate[x^n, {x, 0, 1}]]
> > > > Out[7]= 1 / 3
> > > > ```
> > > > 
> > > > 
> > > > If you want to call this a current limitation, ok. If you'd like to explain why we get what we get, I'd appreciate it.
> > > > I leave to you whether we want to fix or merge and fix later.
> > > > In [f12c348](https://github.com/mathics/Mathics/commit/f12c348f8f361efec4afd7e278c6e50d5316cff2) I removed an unused import. See also #1313 if these are tweaks you want to add.
> > > 
> > > 
> > > The first output is just wrong, but is wrong due an issue in sympy. The problem we have here is the format of the sympy return: we get something that is a piecewise definition, with the last (default) value being the unevaluated integral, but not exactly in the same form. So, there is no safe way to reduce this to a ConditionalExpression, without performing a more or les extense set of tests in the default value...
> > 
> > 
> > And what about the last two? "Or" to me doesn't mean that you can pick which ever one you want, but that both may be true. Note that saying "And" here would be wrong since `x` can't be two values at once.
> 
> You are right: to assume `Or[a==1, a == 2]` is `True` should be equivalent to assume that `And[a!=1, a!= 2]` is `False`, isn't it?

Yes, that is DeMorgan's rule. 

Changing the single equal `=` to a double one `==` doesn't change Mathics' output.

mmatera:
> > > > > Just had a chance to go over in detail and here's what I encountered.
> > > > > ```
> > > > > Assuming[0 < n < 1, Integrate[x^n, {x, 0, 1}]]
> > > > > Out[3]= Piecewise[{{1 / (1 + n) - 0 ^ (1 + n) / (1 + n), n > -Infinity && n < Infinity && n != -1}, {Infinity, True}}]
> > > > > 
> > > > > In[4]:= Assuming[n = 1, Integrate[x^n, {x, 0, 1}]]
> > > > > Out[4]= 1 / 2
> > > > > 
> > > > > In[5]:= Assuming[n = 2, Integrate[x^n, {x, 0, 1}]]
> > > > > Out[5]= 1 / 3
> > > > > 
> > > > > In[6]:= Assuming[Or[n = 2, n = 1] , Integrate[x^n, {x, 0, 1}]]
> > > > > Out[6]= 1 / 2
> > > > > 
> > > > > In[7]:= Assuming[Or[n = 1, n = 2] , Integrate[x^n, {x, 0, 1}]]
> > > > > Out[7]= 1 / 3
> > > > > ```
> > > > > 
> > > > > 
> > > > > If you want to call this a current limitation, ok. If you'd like to explain why we get what we get, I'd appreciate it.
> > > > > I leave to you whether we want to fix or merge and fix later.
> > > > > In [f12c348](https://github.com/mathics/Mathics/commit/f12c348f8f361efec4afd7e278c6e50d5316cff2) I removed an unused import. See also #1313 if these are tweaks you want to add.
> > > > 
> > > > 
> > > > The first output is just wrong, but is wrong due an issue in sympy. The problem we have here is the format of the sympy return: we get something that is a piecewise definition, with the last (default) value being the unevaluated integral, but not exactly in the same form. So, there is no safe way to reduce this to a ConditionalExpression, without performing a more or les extense set of tests in the default value...
> > > 
> > > 
> > > And what about the last two? "Or" to me doesn't mean that you can pick which ever one you want, but that both may be true. Note that saying "And" here would be wrong since `x` can't be two values at once.
> > 
> > 
> > You are right: to assume `Or[a==1, a == 2]` is `True` should be equivalent to assume that `And[a!=1, a!= 2]` is `False`, isn't it?
> 
> Yes, that is DeMorgan's rule.
> 
> Changing the single equal `=` to a double one `==` doesn't change Mathics' output.
That is weird...

mmatera:
In any case, it is clear that And[a,b,c] -> False   is not equivalent to {a->False, b->False, c->False}`, but
`Or[a,b]->False` implies `a->False` and `b->False`. This is the rule that should be implemented, isn't it?

mmatera:
Update:

```
In[6]:= Assuming[Or[n = 2, n = 1] , Integrate[x^n, {x, 0, 1}]]
Out[6]= 1 / 2

In[7]:= Assuming[Or[n = 1, n = 2] , Integrate[x^n, {x, 0, 1}]]
Out[7]= 1 / 3
```
You are assigning here, so this is why it evaluates to the first number. If you use `==` instead, the guilty is my wrong implementation for the inference rules.

mathics/builtin/calculus.py
mmatera: That is a great idea! I implemented it in `BaseExpression` instead of `Symbol`.

mathics/builtin/arithmetic.py
mmatera: > > I think this is ready now: the function `predicate_evaluation(predicate, evaluation)` looks at $Assumptions and implement some replacements on predicate accordingly. This function is called then in the evaluation of `Piecewise`, and `ConditionalExpression`, in a way that propagates automatically to `Integrate`.
> 
> A superficial look through and I think this is exactly the right thing to do. However I want to pull this down and try it and maybe even experiment with the code to understand what's up with sympy mentioned in one of the comments. And probably the weekend is when I'll have an opportunity to do.

Update: I realized that in WMA,  logic expressions are evaluated using assumptions just if you surround them by  ``Simplify``.  On the other hand, `Integrate` uses the option  `Assumptions` (default `$Assumptions`) before return the result. Now this behavior is implemented.

mmatera: Also, I found -I guess- a safe way to reproduce the use of `ConditionalExpression` instead of ``Piecewise`` in the return of Integrate.

mmatera:
Well, this is getting bigger and bigger.  I tried to delegate the predicate evaluation to sympy, but it is simply very limited, and too complicated to use here. So, I implemented our own evaluator. There are several aspects that can be improved, but I think this is a good point to merge this.

rocky (commented):
While there is great stuff in here, I am getting concerned that the overall number of changes is getting too hard to review and understand. 

It would be great to figure out how to split this up into smaller pieces and deliver those. (Sort of like I have been trying to do with adding builtin subdirectories/subcategories.

Some of the changes are trivial and not likely to change like Integer(0) -> Integer0 and Symbol("True")->SymbolTrue. 

Something else easy might be Nand and Nor. 

Is there anything else though? 

Maybe FixedPoint with the SameTest options?

rocky (changes requested):
Let's break this up. I am happy to help here.

mmatera:
If you want, most of the non-trivial changes are included in the new module `inference.py`. We can split this into a very raw implementation of the new module, (let's say, one in which `evaluate_predicate` just returns the input). I can try to do it after I finish removing errors, warnings, and debug code...

mmatera:
This PR was splitted into  #1328 and #1330

PR #1273, opened 2021-04-13
My suggestions for the branch fixcomparison-modularize

mmatera:
In this branch, I merge changes from master and introduce some changes to the fixcomparison-modularize.
Some basic changes include splitting the equal2 method of the base class of Equal into several auxiliary methods for this class (by now, they are three: expr_cmp, sympy_cmp, and infty_cmp). Maybe this helps to make the code more readable.
Apart from this, I removed the equal2 in Numbers, and move it up to the base class Atom. Also, I removed the code doing finite precision comparisons on the equal2 methods in classes derived from BaseExpression: such kind of comparisons should be done by methods that include a parameter max_extra_prec, which are on the side of the Builtin superclasses.

Finally, I implemented a method that generates the list of comparisons that have to be done in order to evaluate the (Un)equality of a list of several expressions.


mathics/builtin/comparison.py
```diff
@@ -219,75 +219,66 @@ def numerify_args(items, evaluation):
...
-
-        lhs_sympy = lhs.to_sympy(evaluate=True, prec=COMPARE_PREC)
-        rhs_sympy = rhs.to_sympy(evaluate=True, prec=COMPARE_PREC)
+    @staticmethod
+    def get_pairs(args):
+        for i in range(len(args)):
+            for j in range(i):
+                yield (args[i], args[j])
+
+    def expr_cmp(self, lhs, rhs, max_extra_prec=None):
```
rocky, line 228: I would appreciate it if you not use the word `cmp` for `Equal` like functions. 

In Python (and other languages) `cmp`  returns some sort of integer where 0 means the items are equal and 1 (or more generally a positive number) means lhs > rhs and -1 (or more generally a negative number) means that lhs < rhs. 

And even in this code before this change, this convention is followed. 

Similarly WL took the name Equal from Lisp and used in other symbolic languages where that means about the same thing. WL uses SameQ for what in Lisp is `eq`. 

But what I am trying to get across here is that although this seems like a niggling concern, keeping things consistent is very helpful for someone steeping in programming languages and symbolic programming languages in particular and the conventions used there. 

So `expr_equal` might be a possibility because it remove `cmp` which shouldn't be used and adds `equal` which should be used. However the exact name I don't care about as long as it doesn't go against existing conventions. 

Also, it would be helpful to me if you'd add the return value `-> Optional[bool]` (which is the same thing as `Union[bool, None]` but more clear. I am sorry I didn't do this initially.

mathics/builtin/comparison.py
```diff
@@ -219,75 +219,66 @@ def numerify_args(items, evaluation):
...
+            return None
+        if len(lhs._leaves) != len(rhs._leaves):
+            return
+        for l, r in zip(lhs._leaves, rhs._leaves):
+            tst = self.equal2(l, r, max_extra_prec)
+            if not tst:
+                return tst
+        return True
+
+    def infty_cmp(self, lhs, rhs, max_extra_prec=None):
```
rocky, line 244: same `cmp` concern.

mathics/builtin/comparison.py
```diff
@@ -219,75 +219,66 @@ def numerify_args(items, evaluation):
...
+                dir2 = rhs._leaves[0]
+            if self.equal2(dir1, dir2, max_extra_prec):
+                return True
+            # Now, compare the signs:
+            dir1 = Expression("Sign", dir1)
+            dir2 = Expression("Sign", dir2)
+            return self.equal2(dir1, dir2, max_extra_prec)
+        return
+
+    def sympy_cmp(self, lhs, rhs, max_extra_prec=None):
```
rocky, line 275: See above for `cmp`

mathics/builtin/comparison.py
```diff
@@ -320,6 +311,27 @@ def equal2(self, lhs, rhs, max_extra_prec=None) -> Union[bool, None]:
         else:
             return None
 
+    def equal2(self, lhs, rhs, max_extra_prec=None) -> Union[bool, None]:
```
rocky, line 314: I apologize for not writing the equivalent `Optional[bool]`

mathics/builtin/comparison.py
```diff
@@ -333,49 +345,12 @@ def apply(self, items, evaluation):
...
-                        return SymbolTrue
-            # if isinstance(x, String) or isinstance(y, String):
-            # if not (isinstance(x, String) and isinstance(y, String)):
-            #    c = 1
-            # else:
-            #    c = cmp(x.get_string_value(), y.get_string_value())
-            else:
-                c = do_cmp(x, y)
+        for x, y in self.get_pairs(args):
+            c = do_cmp_cplx(x, y)
```
rocky, line 349: Notice that here we in fact _do_ use `cmp` in the way that it is typically used.

rocky, line 349: Even though `do_cmp_cplx` calls `do_cmp()` this is again an equal function not a `cmp` function as far as I can tell.

mathics/builtin/comparison.py
```diff
@@ -466,6 +441,39 @@ def apply(self, items, evaluation):
...
+    if isinstance(y, Complex):
+        x, y = y, x
+    if isinstance(x, Complex):
+        if isinstance(y, Complex):
+            c = do_cmp(x.real, y.real)
+            if c is None:
+                return
+            if c != 0:
+                return False
+            c = do_cmp(x.imag, y.imag)
```
rocky, line 454: Notice that _this_ `do_cmp` follows how `cmp` is traditionally used.

mathics/builtin/compilation.py
```diff
@@ -161,6 +161,9 @@ def __init__(self, cfunc, args, **kwargs):
         self.cfunc = cfunc
         self.args = args
 
+    def equal2(self, rhs):
```
rocky, line 164: Thanks! This is what I had been intending to add at some point.

mathics/core/expression.py
```diff
@@ -277,8 +272,9 @@ def equal2(self, rhs: Any) -> Optional[bool]:
 
         # If the types are the same then we'll use the classes definition of == (or __eq__).
         # Superclasses which need to specialized this behavior should redefine equal2()
-        # Think about: should we use isinstance?
-        if type(self) == type(rhs):
+        #
+        # I would use `is` instead `==` here, to compare classes.
```
rocky, line 276: You are absolutely correct here. Thanks for catching!

mathics/core/expression.py
```diff
@@ -739,15 +735,25 @@ def equal2(self, rhs: Any) -> Optional[bool]:
...
+        # From here, we can assume that both heads are the same
+        if self.get_head_name() in ("System`List", "System`Sequence"):
             if len(self._leaves) != len(rhs._leaves):
                 return False
             for item1, item2 in zip(self._leaves, rhs._leaves):
                 result = item1.equal2(item2)
                 if not result:
                     return result
             return True
+        elif self.get_head_name() in ("System`DirectedInfinity",):
```
rocky, line 755: I wonder if there isn't way to add `equal2` to `DirectedInfinity` so this code appears elsewhere? 

What do you think?

mathics/builtin/comparison.py
```diff
@@ -219,75 +219,66 @@ def numerify_args(items, evaluation):
...
+            return
+        for l, r in zip(lhs._leaves, rhs._leaves):
+            tst = self.equal2(l, r, max_extra_prec)
+            if not tst:
+                return tst
+        return True
+
+    def infty_cmp(self, lhs, rhs, max_extra_prec=None):
+        IntegerOne = Integer(1)
+        if rhs.get_head().sameQ(SymbolDirectedInfinity):
```
rocky, line 246: Down the line I have in mind to reduce this kind of thing. But that can come later. 

I wonder though if some of this couldn't be simplified by defining an equal2 in DirectedInfinity.

test/test_compare.py
```diff
@@ -2,647 +2,258 @@
...
-        #     "Compile[{x}, Sqrt[x]] == Sqrt[2]",
-        #     "Compile[{x}, Sqrt[x]] == Sqrt[2]",
-        # ),
-    ],
+    ("str_lhs", "str_rhs", "str_expected"),
+    tests1,
 )
-def test_cmp1_no_pass(str_expr, str_expected):
-    check_evaluation(str_expr, str_expected)
+def test_cmp1_no_pass(str_lhs, str_rhs, str_expected):
```
rocky, line 238: Overall I really like how this test looks. Thanks!

One small thing is that I would add a comment here that we assume that == is commutative and test that this is so. 

I similar comment probably should go in the user-visible doc too.

mathics/builtin/comparison.py
```diff
@@ -219,75 +219,65 @@ def numerify_args(items, evaluation):
...
-
-        lhs_sympy = lhs.to_sympy(evaluate=True, prec=COMPARE_PREC)
-        rhs_sympy = rhs.to_sympy(evaluate=True, prec=COMPARE_PREC)
+    @staticmethod
+    def get_pairs(args):
+        for i in range(len(args)):
+            for j in range(i):
+                yield (args[i], args[j])
+
+    def expr_eq_tst(self, lhs, rhs, max_extra_prec=None) -> Optional[bool]:
```
rocky, line 228: I am sorry I wasn't clearer here we want `expr_equal_txt` here and not `eq` .

In Lisp and symbolic languages which `eq` is `WL SameQ`. But we ahve been using sameQ inside the Python code already. 

If inside some `..eq()` function we are calling `equal2` that is definitiely a clue that we mean `equal` not `eq`. And if we are only call `sameQ` or do sameQ kinds of tests, the the word `sameQ` would be used. 

Either way, don't use `eq` which we happen not to be using. I was just pointing out that this is what other languages typically use for WL `SameQ`.

mathics/builtin/comparison.py
mmatera: I tried to put this code in different places: in a ``DirectedInfinity`` subclass of ``Expression``, or of ``Atom`` or ``Number``, and as a method in the ``Builtin``, and as a rule in the corresponding builtin,  but at the end, this seems to be the proper place. 
In the first set of options, it would require to know ``max_extra_prec`` in the ``equal2`` method. As this information is stored in the ``Definitions`` object,  ``equal2`` should receive this information as an extra parameter. But for the other equal2 methods in ``BaseExpression`` superclasses, this parameter is not needed, so to add to ``Number`` and ``DirectedInfinity`` would introduce an odd inhomogeneity that I would prefer to avoid.

The other possibility is to add this as a method in the class ``DirectedInfinity`` superclass of  ``Builtin``. This could make sense if we consider to add this kind of methods on several other subclasses of ``Builtin``. Then, we should introduce a mechanism to look at all the Builtins to see if they have this method. But actually, we have this only example of a place where we need to implement this.
Finally, the other possibility would be to introduce Builtin rules like  ``DirectedInfinity[a_]==DirectedInfinity[b_] -> Sign[a]==Sign[b]`` and leave the parser to do the job.  This is not a bad option and is a valid way to make this extensible, relying on the existing infrastructure. However, I thought that it would be computationally faster to have these  Python methods on the ``_EqualityOperator``, and provides a concise and clear, and extensible way to deal with this class of special cases.

mmatera (commented):
@rocky, regarding `_cmp`, I was not aware of this convention. In the last commit, I changed it by `_eq_tst`.
Also, I added in these methods the type of return object.  Shall I add this to all the functions? In which cases is necessary to specify the class of the returned objects?

mathics/builtin/comparison.py
mmatera: OK, so ``_equal``  instead of ``_eq_tst`` would be OK?

mathics/builtin/comparison.py
mmatera: copy that!

mathics/builtin/comparison.py
rocky: `_equal` everyplace `_eq_test` is used would be awesome!

Sorry to put you through all of this trouble. (I think though it is all for the better and worth it in the long run.)

mathics/builtin/compilation.py
mmatera: Not sure if the content of this method is what we want at the end, but this works by now.

mathics/core/expression.py
mmatera: See https://github.com/mathics/Mathics/pull/1273/files#r612879671

rocky:
> @rocky, regarding `_cmp`, I was not aware of this convention. In the last commit, I changed it by `_eq_tst`.
> Also, I added in these methods the type of return object. Shall I add this to all the functions? In which cases is necessary to specify the class of the returned objects?

Adding the return type I and others find helpful. So if you are up to doing it please do. Thanks!

mathics/builtin/comparison.py
rocky: > I tried to put this code in different places: in a `DirectedInfinity` subclass of `Expression`, or of `Atom` or `Number`, and as a method in the `Builtin`, and as a rule in the corresponding builtin, but at the end, this seems to be the proper place.
> In the first set of options, it would require to know `max_extra_prec` in the `equal2` method. As this information is stored in the `Definitions` object, `equal2` should receive this information as an extra parameter. But for the other equal2 methods in `BaseExpression` superclasses, this parameter is not needed, so to add to `Number` and `DirectedInfinity` would introduce an odd inhomogeneity that I would prefer to avoid.
> 
> The other possibility is to add this as a method in the class `DirectedInfinity` superclass of `Builtin`. This could make sense if we consider to add this kind of methods on several other subclasses of `Builtin`. Then, we should introduce a mechanism to look at all the Builtins to see if they have this method. But actually, we have this only example of a place where we need to implement this.
> Finally, the other possibility would be to introduce Builtin rules like `DirectedInfinity[a_]==DirectedInfinity[b_] -> Sign[a]==Sign[b]` and leave the parser to do the job. This is not a bad option and is a valid way to make this extensible, relying on the existing infrastructure. However, I thought that it would be computationally faster to have these Python methods on the `_EqualityOperator`, and provides a concise and clear, and extensible way to deal with this class of special cases.

Ok. Thanks for the information and trying. I guess we should just leave it and come back to it later.

As for the differences between equal2 routines, the conventional way to handle this is to add `*args, **kwargs` at the end of all of these functions. The ones that take a prec parameter would have `prec, *args, **kwargs`.  But if this is too much work. Leave it and we can come back later.

Issue #1249, opened 2021-04-10
Issues with OptionValue outside of the main function body


vsht:
Consider the following code
```
$FCMemoryAvailable=4096;
$DisableMemSet=False;
Options[MemSet] = {
	FCMemoryAvailable :> $FCMemoryAvailable
};

MemSet[x_,y_, OptionsPattern[]] :=
	If[(OptionValue[FCMemoryAvailable] - MemoryInUse[]/1000000.) <1. || $DisableMemSet,
		y,
		Set[x, y]
	]/; MatchQ[OptionValue[FCMemoryAvailable],_Integer?Positive];

MemSet[_,_, OptionsPattern[]]:=
	(
	Print["The value of FCMemoryAvailable is ", OptionValue[FCMemoryAvailable], " so this error shouldn't be triggered!"];
	Print[MatchQ[OptionValue[FCMemoryAvailable],_Integer?Positive]];
	Abort[]
	)/; !MatchQ[OptionValue[FCMemoryAvailable],_Integer?Positive];

MemSet[f[x],x^2]
```
Here we set the option FCMemoryAvailable of MemSet to 4096, hence `MatchQ[OptionValue[FCMemoryAvailable],_Integer?Positive]` should return true and the error message from the second definition of MemSet shouldn't appear. In Mma this works as expected.

```
In[1]:= Get["memset.m"]                                                                                                                                                                                                                                                                     

         2
Out[1]= x
```
However, it seems that Mathics doesn't properly parse `OptionValue[FCMemoryAvailable]` in the condition outside of the main body of MemSet. Hence, we enter the second definition of MemSet, where the parsing actually works correctly. And so

```
In[1]:= Get["memset.m"]
The value of FCMemoryAvailable is 4096 so this error shouldn't be triggered!
True
In[2]:= 
```

[memset.zip](https://github.com/mathics/Mathics/files/6289978/memset.zip)

This issue is relevant for #1174

rocky:
Should be fixed by #1274

PR #1274, opened 2021-04-14
Respect OptionValue inside a Condition (another changet to _SetOperator)

mmatera:
This is another patch to assign_elementary that fixes issue  #1249


rocky (approved):
_SetOperator has become quite a mess. #1284 added to remind us to revisit later.

But I guess for now, this is useful.

PR #1276, opened 2021-04-14
FindRoot f(Equal(a, b)) is f(Plus(a, Minus(b))

### LINKED ISSUE #1235: Arguments of FindRoot are not evaluated

rocky:
The code we had before didn't do this properly.

Fixes #1235

### DIFF
--- a/test/test_calculus.py
+++ b/test/test_calculus.py
@@ -29,5 +29,15 @@ def test_calculus():
             "{x->1.51213}",
             "Issue #1235",
         ),
+        (
+            "v1 := Exp[y] - 3y == 0; v2 = {y, 2}; FindRoot[v1, v2]",
+            "{y->1.51213}",
+            "Issue #1235",
+        ),
+        (
+            "v1 := Exp[z] - 3z == 1.0; v2 = {z, 3}; FindRoot[v1, v2]",
+            "{z->1.51213}",
+            "Issue #1235",
+        ),
     ):
         check_evaluation(str_expr, str_expected, message)
--- a/mathics/builtin/calculus.py
+++ b/mathics/builtin/calculus.py
@@ -1156,11 +1156,15 @@ class FindRoot(Builtin):
         "FindRoot[f_, xtuple_]"
         f_val = f.evaluate(evaluation)
 
-        if f_val.has_form("Equal", 2):
-            f = Expression("Minus", *f_val.leaves)
-
         xtuple_value = xtuple.evaluate(evaluation)
+        if f_val.has_form("Equal", 2):
+            f, x0_adjust = f_val.leaves
+        else:
+            x0_adjust = None
+
         if xtuple_value.has_form("List", 2):
             x, x0 = xtuple.evaluate(evaluation).leaves
+            if x0_adjust is not None and x0_adjust.to_python() != 0.0:
+                x0 = Expression("Plus", x0, Expression("Minus", x0_adjust)).evaluate(evaluation)
             return self.apply(f, x, x0, evaluation)
         return


test/test_calculus.py
```diff
@@ -29,5 +29,15 @@ def test_calculus():
             "{x->1.51213}",
             "Issue #1235",
         ),
+        (
+            "v1 := Exp[y] - 3y == 0; v2 = {y, 2}; FindRoot[v1, v2]",
+            "{y->1.51213}",
+            "Issue #1235",
+        ),
+        (
+            "v1 := Exp[z] - 3z == 1.0; v2 = {z, 3}; FindRoot[v1, v2]",
```
mmatera, line 38: why these two different equations should have the same root?

mmatera:
As far as I can understand, FindRoot[f[x]==g[x],{x,x0}]  tries to get a solution of the equation
f[x]-g[x]==0  around x==x0, which is the same than

FindRoot[f[x+x0]==g[x+x0],{x,0}]+x0

but this is not what we have here I think. Am I wrong?

test/test_calculus.py
rocky: > As far as I can understand, FindRoot[f[x]==g[x],{x,x0}] tries to get a solution of the equation
> f[x]-g[x]==0 around x==x0, which is the same than
> 
> FindRoot[f[x+x0]==g[x+x0],{x,0}]+x0
> 
> but this is not what we have here I think. Am I wrong?

Yes, you are correct and this code is incorrect (except in the trivial but often case where the right-hand side of the `==` is 0. 

So this code is wrong.

mmatera:
I ran some tests with the current master and I do not see any issue, at least with the examples that you added in the test.

```
In[43]:= lhs = Exp[z]-3 z
Out[43]= -3 z + E ^ z

In[47]:= {z,lhs} /. FindRoot[lhs==0,{z,0}]
Out[47]= {0.619061, 2.22045*^-16}

In[48]:= {z,lhs} /. FindRoot[lhs==0,{z,1}]
Out[48]= {0.619061, 2.22045*^-16}

In[49]:= {z,lhs} /. FindRoot[lhs==0,{z,2}]
Out[49]= {1.51213, 0.}

In[50]:= {z,lhs} /. FindRoot[lhs==1,{z,0}]
Out[50]= {0., 1.}

In[51]:= {z,lhs} /. FindRoot[lhs==1,{z,1}]
Out[51]= {-1.28657*^-17, 1.}

In[52]:= {z,lhs} /. FindRoot[lhs==1,{z,2}]
Out[52]= {1.90381, 1.}

In[53]:= {z,lhs} /. FindRoot[lhs==1,{z,3}]
Out[53]= {1.90381, 1.}
 ```
This is the plot of lhs:
imagen

**Comment by rocky** [comment]
The example in the ticket is `Exp[z]-3 z == 0`

We can easily special-case the `== 0` and that's done here if you remove the wrong stuff. (However expressions could be symbolically reworked along the lines suggested by the proper substitutions).

But more generaly I am given to understand we should handle nonzero numbers like: 

`Exp[z]-3 z == 4.5`

**Comment by rocky** [comment]
Just reread.  I think the fact that what you noticed worked was also noticed in the issue further down.

You should follow the exact input that is given at the beginning:

```
In[1]:= v1 := Exp[x] - 3x == 0
In[2]:= v2 := {x, 2}
In[3]:= FindRoot[v1, v2]
$RecursionLimit::reclim: Recursion depth of 200 exceeded.
Out[3]= $Aborted
```

mmatera:
The issue is related to the evaluation of derivatives in ``diff`` inside the ``apply`` method. I will have a look at this tomorrow.

mmatera:
@rocky, please check #1277

mmatera:
> The issue is related to the evaluation of derivatives in `diff` inside the `apply` method. I will have a look at this tomorrow.

PR #1277, opened 2021-04-15
fixing FindRoot

mmatera:
This seems to work.

TODO
---------

* Compile Functions and Derivatives
* Implementing other methods
* Implementing checks in AccuracyGoal  and PrecisionGoal
* FindRoot over vectors (F is a vector, or x0 is a vector)

### DIFF
--- a/mathics/builtin/calculus.py
+++ b/mathics/builtin/calculus.py
@@ -19,6 +19,7 @@ from mathics.core.expression import (
     SymbolList,
     SymbolN,
     SymbolRule,
+    from_python,
 )
 from mathics.core.convert import sympy_symbol_prefix, SympyExpression, from_sympy
 from mathics.core.rules import Pattern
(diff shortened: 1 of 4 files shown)


mathics/builtin/calculus.py
```diff
@@ -1039,6 +1040,68 @@ def apply(self, f, n, n0, evaluation, options={}):
         except:
             pass
 
+def find_root_secant(f, x0, x, opts, evaluation) -> (Number, bool):
```
rocky, line 1043: Is it not possbile under various conditions to hook into mpmath's routines for doing this? See https://mpmath.org/doc/current/calculus/optimization.html for example.

If we do so we would I suspect we would improve performance over, immediately provide more solvers, while reducing the chance of there being an error that we introduce.

rocky: And in the improtant case where the end is Equal to 0 or some constant, can't we use Sympy's solve. See https://docs.sympy.org/latest/modules/solvers/solvers.html

mmatera: Indeed, the idea is to do the same as in NIntegrate. But I wanted to have a couple of "native" routines when the optimized routines are not available.

rocky: Ok, that makes sense. But it doesn't look like we are trying to use any of the sympy or mathmp routines that core these. Is this correct?

rocky: I haven't looked at the corresponding mpmath or sympy math routines. How do they compare with what is here code-wise?

mmatera: Before delegating the work on that routines, we should check if it is possible to pre-compile the target function, and its derivatives,  in a way that we can pass those compiled versions. Also, for the one-dimensional problem, the implemented algorithms are quite efficient, having a bottleneck in the function evaluation. On the other hand, FindRoot with vector variables is not implemented yet, so I think that the use of mpmath routines should be implemented at a later stage.

rocky (approved):
LGTM 

Personally I think hooking into sympy would be really nice. And would be nice to benchmark. 

But I will add this to an issue.

rocky:
@mmatera many thanks for fixing.

Issue #1195, opened 2021-03-22
Add DateObject and Now


rocky:
* [Now](https://reference.wolfram.com/language/ref/Now.html)
* [DateObject](https://reference.wolfram.com/language/ref/DateObject.html)

mmatera:
Covered by #1279

PR #1279, opened 2021-04-15
basic implementation of Now and DateObject

mmatera:
Attends to issue #1195

### DIFF
--- a/mathics/builtin/datentime.py
+++ b/mathics/builtin/datentime.py
@@ -630,34 +630,106 @@ class DateString(_DateFormat):
         return from_python("".join(datestrs))
 
 
-class DateObject(Builtin):
+class DateObject(_DateFormat):
     """
     <dl>
       <dt>'DateObject[...]'
       <dd> Returns an object codifiyng DateList....
     </dl>
     >> DateObject[{2020, 4, 15}]
-     = DateObject[{2020, 4, 15, 0, 0, 0.}, Instant, Gregorian, ...]
+     = [...]
     """
 
+    messages = {
+        "notz": "Argument `1` in DateObject is not a recognized TimeZone specification.",
+    }
     options = {
         "TimeZone": "Automatic",
         "CalendarType": "Automatic",
         "DateFormat": "Automatic",
     }
 
+    granularities = [
+        Symbol(s)
+        for s in ["Eternity", "Year", "Month", "Day", "Hour", "Minute", "Instant"]
+    ]
+    rules = {
+        "DateObject[]": "DateObject[AbsoluteTime[]]",
+    }
+    fmt_keywords = {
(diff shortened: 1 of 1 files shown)


mathics/builtin/datentime.py
```diff
@@ -704,6 +734,21 @@ def evaluate(self, evaluation):
...
+      <dd> gives the current time on the system.
+    </dl>
+
+    >> Now
+     = ...
+    """
+
+    def evaluate(self, evaluation):
+        return Expression("DateObject").evaluate(evaluation)
+
```
rocky, line 750: I am guessing some sort of formatating `to_..` routine would be nice:

```
In[1]:= Now
Out[1]= DateObject[{2021, 4, 15, 20, 40, 3.29188}, Instant, Gregorian, -5.]
```

Otherwise, though I love this. Thanks for adding.

mmatera: There are several things that can be improved on this. For example, ``DateObject`` could be defined as an Atom, in a way that it could store a real number instead of several integers, formatting, MakeBoxes, etc. However, the question is for what we want this. For Graphics? For storing dates? depending on what is the current goal, some extra effort can be invested in this.

rocky: Since I opened the ticket here is the intended use. `Now` returns somethin like it does in  https://reference.wolfram.com/language/ref/Now.html. And simple date conversions that convert a date object into a string and vice versa. 

Nothing fancy or nothing involving graphics was contemplated or needed.

mmatera: Yes, 

> Since I opened the ticket here is the intended use.

OK, that is obvious. But the question is, which is the context of use?  
I ask you because I am not very familiar with this set of WL objects. From the examples in the documentation, they seem to be included to handle labels in graphics, and for handling different ways to describe the same time instant (Gregorian/Jewish/Islam calendars, different TimeZones, and different date and time formats). Then, there are a set of operations that can be implemented having DateObjets as operands. 

>  `Now` returns something like it does in https://reference.wolfram.com/language/ref/Now.html. And simple date conversions that convert a date object into a string and vice versa.
>

OK, so defining ``MakeBoxes`` for this kind of expression, ``ToString`` and some basic rules for conversion would be enough, isn't it?
 
> Nothing fancy and no involved graphics was contemplated or needed.

rocky: Yes, that would be Awesome! Thanks.

mmatera: OK, here goes a more complete implementation. Still, some work on handling timezones is required...

rocky: Thanks! LGTM. I am happy to put off for later the timezone for such time as when it matters.

PR #1280, opened 2021-04-16
Fix derivatives

mmatera:
superseeds #1220

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -38,6 +38,7 @@ Bug fixes
 * ``FindRoot`` now handles equations.
 
 
+
 2.1.0
 -----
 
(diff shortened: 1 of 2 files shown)


rocky:
@mmatera This had been on my radar of things to do (that require more serious concentration, and experimentation). I wanted to get to this last weekend but didn't. I will however try to look at it this weekend in depth. Thanks.

rocky (commented):
@mmatera Something got bolixed here. It looks now like a couple of things were reformatted with `black` and I don't see anyting about Bessel.  I suspect then that those changes got dropped.

mmatera:
> @mmatera Something got bolixed here. It looks now like a couple of things were reformatted with `black` and I don't see anyting about Bessel. I suspect then that those changes got dropped.

It seems that the changes here were merged. There were several extra definitions of rules in specialfunctions for the derivatives of Bessel's functions, and something like `D[BesselJ[n,x],x]` do not raise an exception.

rocky:
@mmatera given this is just a tidying of code that is already in master, probably we should merge it. But I'll let you pull the trigger.

PR #1281, opened 2021-04-17
Streams

rocky:
Some background here. At the high level I'd like to make more examples such as the gallery examples or other examples to be put in a file or package and have those when loaded run like our gallery examples work. 

In looking at this I came across InputStream and OutputStream and all the hackiness in using NSTREAMS and STREAMS.

A bit more work is needed to get what I want accomplished however this moves us in the right direction.


mmatera (approved):
In general LGTM. Something that could be included in this PR is to handle binary input and output streams using ``ByteArray`` instead of ``String``.

rocky:
> In general LGTM. Something that could be included in this PR is to handle binary input and output streams using `ByteArray` instead of `String`.

Wil probably add and then merge sometime next week. 

This is getting large, and I want to refactor a lot more to spilt out files.py. Compress/Uncompress shouldn't be there. Splitting of filesystem from files is another possibility. 

But most important is to get InputStream and OutputStream working better.

rocky:
> In general LGTM. Something that could be included in this PR is to handle binary input and output streams using `ByteArray` instead of `String`.

I looked at this and I don't understand. Reading is done by Python and that uses encoding. When the file mode is binary, there can be no encoding.

Maybe if you can point to specific lines of code that would get changed, that would help out.

mmatera:
> > In general LGTM. Something that could be included in this PR is to handle binary input and output streams using `ByteArray` instead of `String`.
> 
> I looked at this and I don't understand. Reading is done by Python and that uses encoding. When the file mode is binary, there can be no encoding.

In the current implementation, when a binary stream is read, the output is returned as a String, after decoding using utf8. It would be convenient to avoid this decoding.

> 
> Maybe if you can point to specific lines of code that would get changed, that would help out.

* `BinaryWrite` should be able to accept ByteArray atoms as an argument.
* `BinaryRead` ( and `OpenRead` with the option `BinaryFormat`) should return a ByteArray
In any case, this can be implemented later.

rocky:
> > > In general LGTM. Something that could be included in this PR is to handle binary input and output streams using `ByteArray` instead of `String`.
> > 
> > 
> > I looked at this and I don't understand. Reading is done by Python and that uses encoding. When the file mode is binary, there can be no encoding.
> 
> In the current implementation, when a binary stream is read, the output is returned as a String, after decoding using utf8. It would be convenient to avoid this decoding.
> 
> > Maybe if you can point to specific lines of code that would get changed, that would help out.
> 
> * `BinaryWrite` should be able to accept `ByteArray` atoms as an argument.
> * `BinaryRead` ( and `OpenRead` with the option `BinaryFormat`) should return a `ByteArray` atom by default
> * `_BinaryFormat` class should support ByteArray as a possible output type.
> 
> In any case, this can be implemented later.

This PR just got even bigger with the split off of filesystem from files. Bit I like this because the docs are better segregated and it is easier to work on the code. 

So my inclination right now is to merge this in just because it is large and this PR is getting unweildy. 

Please open issues to cover the remaining items.  After this is in place I think we will be in a better position do deal with each of these separately.

PR #1287, opened 2021-04-18
CharacterEncoding Rule with ToString unimplemented

rocky:
But let's at least not crash when it is used.

### DIFF
--- a/mathics/builtin/strings.py
+++ b/mathics/builtin/strings.py
@@ -1661,14 +1661,23 @@ class ToString(Builtin):
         "TotalWidth": "Infinity",
     }
 
+    messages = {"encoding": "Encoding not implemented for ToString."}
+
+
     def apply_default(self, value, evaluation, options):
         "ToString[value_, OptionsPattern[ToString]]"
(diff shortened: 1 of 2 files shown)


mathics/builtin/strings.py
```diff
@@ -1661,14 +1661,23 @@ class ToString(Builtin):
...
     def apply_default(self, value, evaluation, options):
         "ToString[value_, OptionsPattern[ToString]]"
         return self.apply_form(value, Symbol("System`OutputForm"), evaluation, options)
 
     def apply_form(self, value, form, evaluation, options):
         "ToString[value_, form_, OptionsPattern[ToString]]"
-        encoding = options["System`CharacterEncoding"]
-        text = value.format(evaluation, form.get_name(), encoding=encoding)
+        if form.has_form("Rule", 2) and form.leaves[0] == Symbol("System`CharacterEncoding"):
+            return evaluation.message("ToString", "encoding")
```
mmatera, line 1674: I think before this message, an `if` clause checking that encoding is not available is missing. Otherwise, the next line is never executed.

rocky:
Completed in Mathics3.

PR #1288, opened 2021-04-18
ColorNegate support for colors

rocky:
As part of my advocacy campaign, I have been adding to the default list of worksheets in the Omnibus Image the examples from "An Elementary Introduction to the Wolfram Language". 

I came across this from https://www.wolfram.com/language/elementary-introduction/2nd-ed/07-colors-and-styles.html and this is an easy fix.

Also as part of advocacy I added Mathics to https://en.wikipedia.org/wiki/List_of_computer_algebra_systems

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,7 @@ New builtins
 Enhancements
 ++++++++++++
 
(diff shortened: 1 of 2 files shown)


mmatera (approved):
Excellent!

PR #1289, opened 2021-04-18
Convert some non-descriptive apply1 names..

rocky:
to something that better expresses the deviation.

I use _with_xx when there is an additional xx parameter

_for is use when the argument type has changed
ColorNegate (for color, for image)


mmatera (approved):
LGTM

PR #1293, opened 2021-04-19
sort predfined Symbols

rocky:
Make it easier to spot if something is on the list and if there
are duplicate definitions as there had been before this commit.


mmatera (approved):
LGTM

PR #1294, opened 2021-04-19
Add Cylinder

rocky:
Apparently even the way old version of Three.js we have supports cylinders. 

There will be a change on the Mathics-Django side as well. 

Haven't considered asymptote or TeX yet though. 

Feel free to just make commits on top of this.

### DIFF
--- a/mathics/builtin/graphics.py
+++ b/mathics/builtin/graphics.py
@@ -29,7 +29,6 @@ from mathics.core.expression import (
     SymbolList,
     SymbolN,
     SymbolMakeBoxes,
-    strip_context,
     system_symbols,
     system_symbols_dict,
     from_python,
@@ -3716,20 +3715,21 @@ class Large(Builtin):
 
 element_heads = frozenset(
     system_symbols(
-        "Rectangle",
-        "Disk",
-        "Line",
         "Arrow",
-        "FilledCurve",
         "BezierCurve",
-        "Point",
         "Circle",
-        "Polygon",
-        "RegularPolygon",
+        "Cylinder",
+        "Disk",
+        "FilledCurve",
         "Inset",
-        "Text",
+        "Line",
+        "Point",
+        "Polygon",
+        "Rectangle",
+        "RegularPolygon",
         "Sphere",
         "Style",
+        "Text",
     )
 )
 
--- a/mathics/builtin/graphics.py
+++ b/mathics/builtin/graphics.py
@@ -1440,10 +1440,12 @@ class _ArcBox(_RoundBox):
 
 
 class DiskBox(_ArcBox):
+    """Internal Python class used when Boxing a in 2D a 'Disk'."""
     face_element = True
 
 
 class CircleBox(_ArcBox):
+    """Internal Python class used when Boxing in 2D 'Circle'."""
     face_element = False
 
 
--- a/mathics/builtin/graphics3d.py
+++ b/mathics/builtin/graphics3d.py
@@ -993,6 +993,34 @@ class Sphere(Builtin):
         "Sphere[positions_]": "Sphere[positions, 1]",
     }
 
+class Cylinder(Builtin):
+    """
+    <dl>
+      <dt>'Cylinder[{$x$, $y$, $z$}]'
+      <dd>is a cylinder of radius 0.5 with height 0.8 centered at the point {$x$, $y$, $z$}.
+
+      <dt>'Cylinder[{$x$, $y$, $z$}, $r$]'
+      <dd>is a cylinder of radius $r$ centered at the point {$x$, $y$, $z$}.
+
+      <dt>'Cylinder[{$x$, $y$, $z$}, $r$, $h$]'
+      <dd>is a cylinder of radius $r$ and height $h$ centered at the point {$x$, $y$, $z$}.
+
+      <dt>'Cylinder[{{$x1$, $y1$, $z1$}, {$x2$, $y2$, $z2$}, ... }, $r$]'
+      <dd>is a collection cylinders of radius $r$ centered at the points {$x1$, $y2$, $z2$}, {$x2$, $y2$, $z2$}, ...
+    </dl>
+
+    >> Graphics3D[Cylinder[{0, 0, 0}, 0.5, 0.8]]
+     = -Graphics3D-
+
+    ## >> Graphics3D[{Yellow, Cylinder[{{-1, 0, 0}, {1, 0, 0}, {0, 0, Sqrt[3.]}}, 1]}]
+    ##  = -Graphics3D-
+    """
+
+    rules = {
+        "Cylinder[]": "Cylinder[{0, 0, 0}, 0.5, 0.5]",
+        "Cylinder[positions_]": "Cylinder[positions, 0.5, 0.5]",
+    }
+
 
 class Cuboid(Builtin):
     """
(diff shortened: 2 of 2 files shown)


mmatera:
I wonder if it would be possible to use mathics_scanner to define parsers for Graphics/Graphics3D like expressions. Then, it would be helpful to update or define new backends...

rocky:
> I wonder if it would be possible to use mathics_scanner to define parsers for Graphics/Graphics3D like expressions. Then, it would be helpful to update or define new backends...

I don't understand what this means but would like to. Pretend this exists, what you then write?

rocky:
By the way, I do think it would be nice to have a Three.js backend.

rocky:
I am seeing a problem with Asymptote formatting of cyclinders where the bounding box starts below the base of the cylinder. However I can't figure out how to fix this. @mmatera if you can, let's fix. Otherwise let's skip.

rocky:
I just checked the WL specification and at least to me it is weird in that from the two-argument point form you extract the length.

mmatera:
> I just checked the WL specification and at least to me it is weird in that from the two-argument point form you extract the length.

I do not get this. Could you expand it with an example?

rocky:
> > I just checked the WL specification and at least to me it is weird in that from the two-argument point form you extract the length.
> 
> I do not get this. Could you expand it with an example?

From https://reference.wolfram.com/language/ref/Cylinder.html : 

> Cylinder[] is equivalent to Cylinder[{{0,0,-1},{0,0,1}}].
> ... 
> Graphics3D[Cylinder[{{0, 0, 0}, {1, 1, 1}}, 1/2]]

I don't know how one orients something inside the grid box - For a sphere orientation is nonexistent. 

We need to be concerned with not just how WL takes its Cyclinder input, but pay attention to the most common definition for asymptote, Three.js and other graphics libraries that provide a cylinder. The consensus is not to do it this way, but instead have origin, radius and length. Sometimes "radius" is diameter and called "width", but the conversion from radius to width is trivial. So I think the conversion from two points to a length would be done in the Boxing routine might add minor adjustments to this.

mmatera:
> > > I just checked the WL specification and at least to me it is weird in that from the two-argument point form you extract the length.
> > 
> > 
> > I do not get this. Could you expand it with an example?
> 
> From https://reference.wolfram.com/language/ref/Cylinder.html :
> 
> > Cylinder[] is equivalent to Cylinder[{{0,0,-1},{0,0,1}}].
> > ...
> > Graphics3D[Cylinder[{{0, 0, 0}, {1, 1, 1}}, 1/2]]
> 
> I don't know how one orients something inside the grid box - For a sphere orientation is nonexistent.
> 
> We need to be concerned with not just how WL takes its Cyclinder input, but pay attention to the most common definition for asymptote, Three.js and other graphics libraries that provide a cylinder. The consensus is not to do it this way, but instead have origin, radius and length. Sometimes "radius" is diameter and called "width", but the conversion from radius to width is trivial. So I think the conversion from two points to a length would be done in the Boxing routine might add minor adjustments to this.
 

In a 3DGraphics, you have a canonical set of coordinate orthogonal axes and a viewpoint. De default cylinder is just a cylinder of radius 1, height 2, center in the origin, and with a symmetry axis aligned with the z-axis.
Generic cylinders are defined in terms of the radius and the two central points of the corresponding tapes.  
In other graphic languages, a cylinder is first defined in terms of its radius and height, and then the object is rotated and translated. It is not hard to define a rotation that brings the z axis onto a vector v: if v is parallel to the z axis, do nothing. Otherwise, normalize de vectors   v'=v/|v|,   n=v x z/|vxz|  and m=v' x n, and build the rotation matrix R = [n m v'].
The other way is to read the spherical coordinates of v by
phi = Arg(v[[1]]+I v[[2]])
theta = ArcCos(v[[3]]/|v|)

rocky:
> > > > I just checked the WL specification and at least to me it is weird in that from the two-argument point form you extract the length.
> > > 
> > > 
> > > I do not get this. Could you expand it with an example?
> > 
> > 
> > From https://reference.wolfram.com/language/ref/Cylinder.html :
> > > Cylinder[] is equivalent to Cylinder[{{0,0,-1},{0,0,1}}].
> > > ...
> > > Graphics3D[Cylinder[{{0, 0, 0}, {1, 1, 1}}, 1/2]]
> > 
> > 
> > I don't know how one orients something inside the grid box - For a sphere orientation is nonexistent.
> > We need to be concerned with not just how WL takes its Cyclinder input, but pay attention to the most common definition for asymptote, Three.js and other graphics libraries that provide a cylinder. The consensus is not to do it this way, but instead have origin, radius and length. Sometimes "radius" is diameter and called "width", but the conversion from radius to width is trivial. So I think the conversion from two points to a length would be done in the Boxing routine might add minor adjustments to this.
> 
> In a 3DGraphics, you have a canonical set of coordinate orthogonal axes and a viewpoint. De default cylinder is just a cylinder of radius 1, height 2, center in the origin, and with a symmetry axis aligned with the z-axis.
> Generic cylinders are defined in terms of the radius and the two central points of the corresponding tapes.
> In other graphic languages, a cylinder is first defined in terms of its radius and height, and then the object is rotated and translated. It is not hard to define a rotation that brings the z axis onto a vector v: if v is parallel to the z axis, do nothing. Otherwise, normalize de vectors v'=v/|v|, n=v x z/|vxz| and m=v' x n, and build the rotation matrix R = [n m v'].
> The other way is to read the spherical coordinates of v by
> phi = Arg(v[[1]]+I v[[2]])
> theta = ArcCos(v[[3]]/|v|)

What you say, makes sense but this is a bit more involved that I want to undertake right now. So for myself I am going to leave this be.

mmatera:
> > > > > I just checked the WL specification and at least to me it is weird in that from the two-argument point form you extract the length.
> > > > 
> > > > 
> > > > I do not get this. Could you expand it with an example?
> > > 
> > > 
> > > From https://reference.wolfram.com/language/ref/Cylinder.html :
> > > > Cylinder[] is equivalent to Cylinder[{{0,0,-1},{0,0,1}}].
> > > > ...
> > > > Graphics3D[Cylinder[{{0, 0, 0}, {1, 1, 1}}, 1/2]]
> > > 
> > > 
> > > I don't know how one orients something inside the grid box - For a sphere orientation is nonexistent.
> > > We need to be concerned with not just how WL takes its Cyclinder input, but pay attention to the most common definition for asymptote, Three.js and other graphics libraries that provide a cylinder. The consensus is not to do it this way, but instead have origin, radius and length. Sometimes "radius" is diameter and called "width", but the conversion from radius to width is trivial. So I think the conversion from two points to a length would be done in the Boxing routine might add minor adjustments to this.
> > 
> > 
> > In a 3DGraphics, you have a canonical set of coordinate orthogonal axes and a viewpoint. De default cylinder is just a cylinder of radius 1, height 2, center in the origin, and with a symmetry axis aligned with the z-axis.
> > Generic cylinders are defined in terms of the radius and the two central points of the corresponding tapes.
> > In other graphic languages, a cylinder is first defined in terms of its radius and height, and then the object is rotated and translated. It is not hard to define a rotation that brings the z axis onto a vector v: if v is parallel to the z axis, do nothing. Otherwise, normalize de vectors v'=v/|v|, n=v x z/|vxz| and m=v' x n, and build the rotation matrix R = [n m v'].
> > The other way is to read the spherical coordinates of v by
> > phi = Arg(v[[1]]+I v[[2]])
> > theta = ArcCos(v[[3]]/|v|)
> 
> What you say, makes sense but this is a bit more involved that I want to undertake right now. So for myself I am going to leave this be.

I could take this if we have in clear how to pass a rotation to Tree.js.

rocky:
It looks like in r49 the graphics object has a Quaternion object which has methods setFrom... methods like setFromRotationMatrix, setFromEuler, etc. See https://github.com/mrdoob/three.js/blob/r49/src/core/Quaternion.js and https://github.com/mrdoob/three.js/blob/r49/docs/api/core/Quaternion.html

mmatera:
I was looking at this, and I see that to do this properly, we need to implement the `Transform` Builtin proposed by @poke1024.   When I have some time, I will continue with the rebase of that PR, and at least take the parts that fit here. Then, the implementation should become easier...

PR #1295, opened 2021-04-20
Conditional expression plus evpredicates

mmatera:
@rocky, this branch contains a starting point for using `$Assumptions` in things like `ConditionalExpression`.


mmatera:
Merged into PR #1271

PR #1296, opened 2021-04-20
Adding Arg 

mmatera:
This PR adds ``Arg`` based on Sympy. Some extra rules in Mathics add added in order to handle infinity quantities.


rocky:
> The only problem is that is not working :( I have to check why...

Ok. I've run into that kind of situation too.

PR #1298, opened 2021-04-22
Letter number

rocky:
@mmatera Again not sure what error messages/tests we should have here.  Please advise.

Also, handling `Alphabet`s is probably  a bigger project that we probably need come back to.

### DIFF
--- a/mathics/builtin/strings.py
+++ b/mathics/builtin/strings.py
@@ -668,6 +668,9 @@ class LetterNumber(Builtin):
     >> LetterNumber[Characters["Peccary"]]
     = {16, 5, 3, 3, 1, 18, 25}
 
+    >> LetterNumber[{"P", "Pe", "P1", "eck"}]
+    = {16, {16, 5}, {16, 0}, {5, 3, 11}}
+
     #> LetterNumber[4]
      : The argument 4 is not a string.
      = LetterNumber[4]
@@ -694,19 +697,10 @@ class LetterNumber(Builtin):
                 r = [letter_number(c, start_ord)[0] if c.isalpha() else 0 for c in py_chars]
                 return Expression("List", *r)
         elif chars.has_form("List", 1, None):
-            leaves = chars.leaves
-
-            # if len(leaves) == 0:
-            #     return None
-            # if len(leaves) > 2:
-            #     return None
-            # else:
-            #     if len(leaves) == 2:
-            #         # FIXME: validate 2nd arg and handle it.
-            #         pass
-            strings = [leaf.get_string_value() for leaf in leaves]
-            r = [letter_number(c, start_ord)[0] if c.isalpha() else 0 for chars in strings for c in chars]
-            return Expression("List", *r)
+            result = []
+            for leaf in chars.leaves:
+                result.append(self.apply(leaf, evaluation))
+            return Expression("List", *result)
         else:
             return evaluation.message(self.__class__.__name__, "nas", chars)
         return None
--- a/mathics/builtin/strings.py
+++ b/mathics/builtin/strings.py
@@ -661,12 +661,25 @@ class LetterNumber(Builtin):
     >> LetterNumber["B"]
      = 2
 
+    >> LetterNumber["ss2!"]
+     = {19, 19, 0, 0}
+
     Get positions of each of the letters in a string:
     >> LetterNumber[Characters["Peccary"]]
     = {16, 5, 3, 3, 1, 18, 25}
+
+    #> LetterNumber[4]
+     : The argument 4 is not a string.
+     = LetterNumber[4]
     """
 
-    # We proobably need to implement Alphabet to handle other
+    messages = {
+        "nas": (
+            "The argument `1` is not a string."
+        ),
+    }
+
+    # We probably need to implement Alphabet to handle other
     # alphabets like Greek and Russian.
     def apply(self, chars: List[Any], evaluation):
         "LetterNumber[chars_]"
@@ -678,7 +691,8 @@ class LetterNumber(Builtin):
                 # FIXME generalize ord("a")
                 return letter_number([py_chars[0]], start_ord)[0]
             else:
-                return None
+                r = [letter_number(c, start_ord)[0] if c.isalpha() else 0 for c in py_chars]
+                return Expression("List", *r)
         elif chars.has_form("List", 1, None):
             leaves = chars.leaves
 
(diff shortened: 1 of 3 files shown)


mmatera:
Does this help?

```
In[1]:= LetterNumber["ss2!"] 
Out[1]= {19, 19, 0, 0}

In[2]:= LetterNumber[4]

LetterNumber::nas: The argument 4 is not a string.
Out[2]= LetterNumber[4]

In[3]:= LetterNumber[Graphics[{}]]

LetterNumber::nas: The argument -Graphics- is not a string.
Out[3]= LetterNumber[Graphics[{}]
```

mathics/builtin/strings.py
```diff
@@ -638,6 +638,69 @@ class LetterCharacter(Builtin):
...
+    # each char has length 1.
+    return [Integer(ord(char.lower()) - start_ord) for char in chars]
+
+class LetterNumber(Builtin):
+    """
+    <dl>
+      <dt>'LetterNumber'[$c$]
+      <dd>returns the position of the character $c$ in the English alphabet.
+
+      <dt>'LetterNumber["string"]'
```
mmatera, line 653: I would implement `Alphabet[]` at least for English, (Latin?) and the second parameter, in a way that anything that is not the default raises a message. At the end, the basic implementation of Alphabet is  just a dictionary with the form "alphabetname": "abcd..."
isn't it?

rocky: You have a clearer idea of what to do at this point than I do. So if you'd just do it, I'd appreciate. 

Otherwise I am happy to have this hang out as a draft for a while. 

For larger context, I have been going over https://www.wolfram.com/language/elementary-introduction/2nd-ed/11-strings-and-text.html and more generally https://www.wolfram.com/language/elementary-introduction/2nd-ed/ which has the attribution Copyright 2021 (and nothing more). The 1st Edition, which I have in book form, has a "non-commercial share-alike" copyright and that is similar. 

And the bigger context even here is that these give examples that can be used in worksheets. (See the docker image or the sqlite file in mathics-omnibus and please also see https://github.com/Mathics3/mathics-django/pull/32).  But in addition this is a much more gentle way to guide us in filling out the code in a way where users can see immediate results. 

The problem I have with FeynCalc, Rubi, KnotTheory and similar packages like this is that they are hundreds if not thousands of lines long and use sometimes sophisticated constructs in intertwined ways.  I think if we have a more solid base to start out with, things will go easier there. 

I tried pulling a 2006 version of KnotTheory and tried building it. In terms of the number of things we need to fix, there are far fewer. I counted maybe 3 or 4 things. Still, without those I am not able to get this to run. And at least for me this is kind of disappointing.

mmatera:
In that case, OK, but if you want to merge this, I think it does not hurt. In can improve this later.

rocky:
I'll make a pass at some point to sync with the error messages. 

During the week,  basically I have only small chunks of time to do things. Whatever can be done in this time, I do, but things that don't fit have to wait.

mathics/builtin/strings.py
axkr: >  At the end, the basic implementation of Alphabet is just a dictionary with the form "alphabetname": "abcd..."
> isn't it?
I think there exists python bindings for the ICU project to create alphabets:
- http://site.icu-project.org/

Unfortunately the alphabets doesn't seem to be exactly the same as in MMA.

rocky: I looked at the ICU project and it looks both awesome and bureaucratic. 

The desire is extremely simple and basic: for each language give me an ordered list of the alphabet with a way to convert from one case to the other. And don't even need bidirectional conversion so you can choose which case to start out with.

However, although I can see how to do case conversion using unicode properties (and you'd think then that a library would use that provide such a function equivalent to "lower" or "upper"), there isn't anything that says give me the first letter in the alphabet and iterate to to the last letter, as far as I can tell. 

But I could be wrong here. The documentation while extensive isn't all that useful. Sigh.

mmatera: I think that an ICU-based `Alphabet` is something to implement in an external module. Here, I would limit to define a basic set of alphabets (let's say,  "English", "Spanish", "German", and  "Greek"  alphabets, that is what I could handle). A Pymathics module then can overload this basic implementation.

axkr: With ICU more things like `IntegerName` and `Transliterate` can be created:
- https://github.com/axkr/symja_android_library/blob/master/symja_android_library/doc/functions/IntegerName.md
- https://github.com/axkr/symja_android_library/blob/master/symja_android_library/doc/functions/Transliterate.md

rocky: Yes, I was thinking along those lines too. There seems a lot of functionality somewhere in there, which extends beyond just Alphabet.

At some point I will post a query on StackOverflow. But if you look at past queries on this topic they generally are met with derision. I just upvoted https://stackoverflow.com/questions/32375797/what-unicode-ranges-are-considered-letters and it appears I am the _only_ one to have done so just now after 5 1/2 years with no great answers.

axkr: NO problemif you want to port it.

rocky:
@mmatera I think this is ready for this level of detail. However one code path that we don't test is found in this example: 

```
>> LetterNumber[{"P", "Pe", "P1", "eck"}]
 = {16, 16, 5, 16, 0, 5, 3, 11}
```

and that's because I don't know if the above is what is expected.

mmatera:
> @mmatera I think this is ready for this level of detail. However one code path that we don't test is found in this example:
> 
> ```
> >> LetterNumber[{"P", "Pe", "P1", "eck"}]
>  = {16, 16, 5, 16, 0, 5, 3, 11}
> ```
> 
> and that's because I don't know if the above is what is expected.

In WMA, the output of that sentence is
`{16, {16, 5}, {16, 0}, {5, 3, 11}}`

rocky:
> > @mmatera I think this is ready for this level of detail. However one code path that we don't test is found in this example:
> > ```
> > >> LetterNumber[{"P", "Pe", "P1", "eck"}]
> >  = {16, 16, 5, 16, 0, 5, 3, 11}
> > ```
> > 
> > 
> > and that's because I don't know if the above is what is expected.
> 
> In WMA, the output of that sentence is
> `{16, {16, 5}, {16, 0}, {5, 3, 11}}`

Should be addressed in 71b755df

rocky:
@mmatera I was thinking about this a little more. For a small subset of cases not requiring a `LoadModule["pymathics`ICU`"]` I suppose the small set we have here is fine. 

But please, let us not extend this in core this way.  Instead let us delegate this out to a Pymathics module which is based on something that purports to handle in more general language support.

mathics/builtin/strings.py
```diff
@@ -42,6 +43,36 @@
 }
 
 
+alphabet_descriptions = {
+    "English": {
+        "Lowercase": "abcdefghijklmnopqrstuvwxyz",
+        "Uppercase": "ABCDEFGHIJKLMNOPQRSTUVWXYZ",
+    },
+    "Spanish": {
+        "Lowercase": "abcdefghijklmnñopqrstuvwxyz",
```
rocky, line 52: WL Aphabets omit any accented characters such as those in Spanish above an "e", but leave those tilde for "n"?  Similarly for an umlaut for Germain?

mmatera: No! This is the "Eñe" and is a letter very important for Spanish speakers! https://en.m.wikipedia.org/wiki/%C3%91
:)

mmatera:
Actually, my initial idea was to use `Alphabet` to something which allows to hook custom definitions inside LetterNumber and other similar builtins. Then, the actual definition of `Alphabet` could be implemented as a pymathics module, or as a .m WL module. The problem is how to implement the "lowercase" for generic alphabets.

PR #1304, opened 2021-04-22
Set $InputFileName on mathics_open()

rocky:
TODO write a test for this.


mmatera:
I think the test would be the one that motivated this PR. If that test passes, then we can implement it later in the test engine.

mmatera (approved):
LGTM

mmatera:
> LGTM

I included a simple test for $InputFileName. If passes in all the platforms, I think that this is ready for merging.

Issue #1240, opened 2021-04-07
``\(something\)-> RowBox[{"something"}]`` and ``\!\(something\)-> DisplayForm[RowBox[{"something"}]]``


mmatera:
In KnotTheory, these "operator" are used. So, we would need to implement this in the scanner.
See
[https://reference.wolfram.com/language/tutorial/InputSyntax.html](https://reference.wolfram.com/language/tutorial/InputSyntax.html) ("String‐Based Input" section)

and 
[https://reference.wolfram.com/language/tutorial/TextualInputAndOutput.html#28564](https://reference.wolfram.com/language/tutorial/TextualInputAndOutput.html#28564)

mmatera:
PR #1306 should fix this.

PR #1306, opened 2021-04-23
fix support for \!

mmatera:
This handles #1240

### DIFF
--- a/mathics/builtin/strings.py
+++ b/mathics/builtin/strings.py
@@ -14,7 +14,13 @@ from heapq import heappush, heappop
 from typing import Callable
 
 from mathics.version import __version__  # noqa used in loading to check consistency.
-from mathics.builtin.base import BinaryOperator, Builtin, Test, Predefined
+from mathics.builtin.base import (
+    BinaryOperator,
+    Builtin,
+    Test,
+    Predefined,
+    PrefixOperator,
+)
 from mathics.core.expression import (
     Expression,
     Symbol,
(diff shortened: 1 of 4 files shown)


mathics/builtin/strings.py
```diff
@@ -1688,6 +1695,31 @@ def apply_form(self, value, form, evaluation, options):
...
+      <dd>is the ad hoc fullform for \! $box$. just
+          for internal use...
+
+    >> \! \(2+2\)
+     = 4
+    </dl>
+    """
+
+    operator = "\\!"
+    precedence = 670
```
rocky, line 1711: @mmatera Just curious - how did you arrive at the value 670?

rocky:
@mmatera I don't have any thoughts or comments other than how the precidence was determined and how we can check that this is correct. 

If this helps, let's go with it.

mathics/builtin/strings.py
mmatera: This seems to be the default in WL...
```
In[1]:= Precedence[anycustomsymbol]
Out[1]= 670
```

rocky: Thanks - for the information.

mmatera: Just a comment:  `InterpretedBox` is not a symbol in WL. However, someone else introduced it in the MathicsScanner (`\!`,  `\(`, etc) were already implemented in the scanner. I just hooked it to a new class that provides the basic elements needed for the parser, and a more or less simple way to reproduce the basic behavior.

rocky: Ok. As with too many other things, I guess we will address in a more complete fashion latter. 

For now this doesn't seem too disruptive or complex.

PR #1308, opened 2021-04-23
Revert "Feyncalcfixes"

mmatera:
I merge this here instead in mmatera/Mathics:master. I will try to revert this:    Reverts mathics/Mathics#1307


rocky:
@mmatera I think we reverted this previously by just reseting the SHA1.

The way this was done was:

```
$ git reset --hard 7fb691aae9bf21d5437acb153ecc # 7b... is the SHA1 of the point we wanted to go back to . 
$ git push --force origin HEAD  # it is useful and careful to add "origin HEAD" to limit this to the one branch
```

PR #1309, opened 2021-04-24
DRY pytest tests

rocky:
In preparation for working on $InputFilename tests I realized that we have a _lot_ of duplicate boilerplate code in our py.tests.  While this isn't a thorough cleaning, it is a start.

### DIFF
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -30,4 +30,4 @@ jobs:
     - name: Test Mathics
       run: |
         pip install pytest pexpect
-        make -j3 check
+        make check
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -30,4 +30,4 @@ jobs:
     - name: Test Mathics
       run: |
         pip install pytest pexpect
-        make check
+        make -j3 check
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -29,4 +29,4 @@ jobs:
     - name: Test Mathics
       run: |
         pip install pytest pexpect
-        make -j3 check
+        make check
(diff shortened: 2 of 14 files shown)


mmatera (approved):
I think this is ready for merging.

.github/workflows/osx.yml
```diff
@@ -30,4 +30,4 @@ jobs:
     - name: Test Mathics
       run: |
         pip install pytest pexpect
-        make -j3 check
```
rocky, line 33: I understand the pros and cons of `-j3`: in the normal case this saves time which I appreciate. But when there is an error, it make things harder to track down. 

There might be a way to set this in an individual run for debugging. I don't know.

.github/workflows/ubuntu.yml
```diff
@@ -29,4 +29,4 @@ jobs:
     - name: Test Mathics
       run: |
         pip install pytest pexpect
-        make -j3 check
```
rocky, line 32: And one other thought here is that it is possible to do this differently for different OS's.

rocky:
> I think this is ready for merging.

Thanks for doing this. Much cleaner. I think we should squash the commits before or as part of the merge.

.github/workflows/ubuntu.yml
mmatera: Sorry for this. I disabled -j3 just for debugging, and I planed to put it back after finishing, but I forgot to do that...

mmatera:
> > I think this is ready for merging.
> 
> Thanks for doing this. Much cleaner. I think we should squash the commits before or as part of the merge.

OK, I leave this part to you if you agree.

rocky:
@mmatera I removed one of changes in `test_files.py` which wasn't respecting using the Windows path separator, but I believe there is another PR that has this fixed, so that should go in when that is merged. (This PR is big enough already.)

I squashed things a little bit, but not totally.  This PR is still a bit rambling, but so be it.

PR #1310, opened 2021-04-24
support for Alphabets

mmatera:
@rocky, this is how I think the support for different alphabets could work. A maybe still better approach is to use also `dicts` (instead of strings) to store the alphabets. Thoughts?


rocky:
The changes outside adding Alphabets are great! Thanks. 

As for Alphabets, the problem is that this isn't scalable to the vast variety that WL supports. To do this properly and more fully the approach of hooking into an existing library like ICU in a Pymathics modue is the better long-term solution. 

But in the short term, sure let's go with this.

PR #1313, opened 2021-04-25
Misc Arg tweaks:

rocky:
* `Arg[0]` needs to be defined 0 (otherwise it can be indeterminate)
* More complete doc for Arg
* Add Method option


mmatera:
LGTM

PR #1314, opened 2021-04-25
Use has_form() more effectively in a couple of places

rocky:
A couple of `has_form()` uses noticed in going over #1271


mmatera:
LGTM

PR #1317, opened 2021-04-25
Accept Prime[] with a list.

rocky:


mmatera (approved):
LGTM

mmatera:
Just one thing: I have to check, but I think that the attribute `Listable` just convert 
`F[{a,b,c,d}]` into ${F[a],F[b],F[c],F[d]}$. This is OK, but implies the independent evaluation of `F[a]`, `F[b]`, `F[c]` and `F[d]`. 
However, often is more efficient to evaluate  `F[a]`, `F[b]`, `F[c]` and `F[d]` in the same routine. This is the case for example when you want to find prime numbers, to evaluate terms in Fibonacci's succession. This would be nice to have a mechanism to combine both possibilities (tagging a function as `Listable`, while the evaluation is performed over the list, and not over the individual elements.

rocky:
Accept Prime[] with a list.

> Just one thing: I have to check, but I think that the attribute `Listable` just convert
> `F[{a,b,c,d}]` into ${F[a],F[b],F[c],F[d]}$. This is OK, but implies the independent evaluation of `F[a]`, `F[b]`, `F[c]` and `F[d]`.
> However, often is more efficient to evaluate `F[a]`, `F[b]`, `F[c]` and `F[d]` in the same routine. This is the case for example when you want to find prime numbers, to evaluate terms in Fibonacci's succession. This would be nice to have a mechanism to combine both possibilities (tagging a function as `Listable`, while the evaluation is performed over the list, and not over the individual elements.

Ok. We should keep that in mind. But not having the Listable attribute set is plain wong, and it is better to have something that is inefficient than wrong. 

In going over the items. there were routines that accepted lists but didn't have the Listable attribute set. I suspect that the Listable attribute doesn't override the code, but we should check.

rocky:
One last thing to note. In functions that are listable, the default behavior when one of the items has an error in it is to fill in the None value. I suspect WL may handle this differently on certain occasions. 

Therefore either we will need a way to specify what to do when there is a prolbme (when it not the current behavior), or may need to implement listable explicitly to handle this.

PR #1318, opened 2021-04-25
Split and Sort specialfunctions

rocky:
This is in preparation for other work I wanted to do and bigger refactoring.

Initially, I was just trying to go over functions that have `NumericFunction` and `ListableFunction` properties.

(BTW in `Select[Names["System`*"], MemberQ[Attributes[#], Listable] &]` has a bug somehwere in it.)

It was helpful to have the names sorted for this purpose so I started doing that in specialfunctions. This was unweildy so I have split out  `legandre.py` and `bessel.py` from `specialfunctions.py`. 

Down the line though I'd like to have a subdirectory of builtins called specialfunctions and then we can also adjust our documentation to track this hierachy.  But even as is, I like the fact that the list in SpecialFunctions is smaller while Bessel functions and Legendre polynomials have their own small list.

### DIFF
--- a/mathics/builtin/__init__.py
+++ b/mathics/builtin/__init__.py
@@ -7,7 +7,7 @@ import os.path as osp
 from mathics.settings import ENABLE_FILES_MODULE
 from mathics.version import __version__  # noqa used in loading to check consistency.
 
-# Get a list of file in this directory. We'll exclude from the start
+# Get a list of files in this directory. We'll exclude from the start
 # files with leading characters we don't want like __init__ with its leading underscore.
 __py_files__ = [
     osp.basename(f[0:-3])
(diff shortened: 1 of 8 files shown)


mmatera:
OK, sounds good, but before splitting, check the branch fixbessel, where I had added rules for the derivatives of many of the special functions.
For some reason, the branch fixderivatives does not contains these changes.

mmatera:
> OK, sounds good, but before splitting, check the branch fixbessel, where I had added rules for the derivatives of many of the special functions.
> For some reason, the branch fixderivatives does not contains these changes.

False alarm : changes in fixderivatives seems merged, so all these changes are already in this branch

PR #1319, opened 2021-04-26
Go over functions with Listable attribute..

rocky:
up to about Range. Start from about there...

Also expand Attributes to accept a string parameter as it should.


mmatera (approved):
LGTM

PR #1331, opened 2021-04-30
timing individual tests

mmatera:
Just another small took for checking performance along the doctests.


rocky:
I just tried this. If you want to merge this in to master, I don't see any problem.

PR #1332, opened 2021-05-01
Add logical operations: Nand and Nor.

rocky:
Note: there will be futher changes to And when we merge ConditionalExpressions.

### DIFF
--- a/mathics/builtin/logic.py
+++ b/mathics/builtin/logic.py
@@ -3,7 +3,14 @@
 from mathics.version import __version__  # noqa used in loading to check consistency.
 from mathics.builtin.base import BinaryOperator, Predefined, PrefixOperator, Builtin
 from mathics.builtin.lists import InvalidLevelspecError, python_levelspec, walk_levels
-from mathics.core.expression import Expression, Symbol
+from mathics.core.expression import (
+    Expression,
+    Symbol,
+    SymbolTrue,
+    SymbolFalse,
+)
+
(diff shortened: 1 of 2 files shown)


rocky:
@mmatera when this clears and gets into master, I will rebase https://github.com/mathics/Mathics/pull/1330 for the further changes that are needed there. I can either put this in a new branch, or overwrite the existing one - your choice. 

But the idea here, is that `Nand` and `Nor` is a separate useful thing independent of Condtional Expressions. By having these separated it helps better understand what's going on not just in that PR but in the future if we try to understand what's up or decouple the two.

Thanks for your patience and understanding.

mmatera:
@rocky, thanks for doing this. Go ahead with the merging and rebasing.

mathics/builtin/logic.py
```diff
@@ -118,6 +125,38 @@ class Not(PrefixOperator):
     }
 
 
+class Nand(Builtin):
+    """
+    <dl>
+    <dt>'Nand[$expr1$, $expr2$, ...]'
+    <dt>'$expr1$ && $expr2$ && ...'
+        <dd> Implements the logical `NOR` function.  The same that Not[And[$expr1$, $expr2$, ...]]
```
axkr, line 133: NOR seems to be a typo.

rocky: Thanks - should be fixed now in 369f6d18

PR #1336, opened 2021-05-01
Reinstate tests from test $InputFileName in recursive get

rocky:
Reinstate some $InputFileName tests from test/files.py


rocky:
@mmatera it looks like there is still a bug in `$InputFilename` from this test as best as I can tell.

rocky:
Is completed in Mathics3.

PR #1338, opened 2021-05-01
Segregate I/O, Files, and Filesystem builtins

rocky:


mathics/builtin/__init__.py
```diff
@@ -116,22 +116,25 @@ def is_builtin(var):
...
-    module_names += ["files", "importexport"]
-
 modules = []
 import_builtins(module_names)
 
 _builtins = []
 builtins_by_module = {}
 
-for subdir in ("drawing", "numbers", "specialfns",):
+disable_file_module_names = ["files_io.files", "files_io.importexport"] if ENABLE_FILES_MODULE else []
```
mmatera, line 130: I do not get this line: why you put things in a list of disabled, if that modules are enabled?

mmatera:
Apart from that, it seems to be OK

mathics/builtin/__init__.py
rocky: You are absolutely correct - this is backwards! Fixing now...

(It was an artifact of not properly transforming more complicated code...)

rocky:
> Apart from that, it seems to be OK

Thanks again for checking over. Should be fixed in b05c648f and I've moved more of the Main Loop objects from expression here and started documenting this better. 

So please doublecheck that I haven't botched things further in doing this. Thanks.

Issue #1329, opened 2021-04-30
General::noopen even if $Path is set


tueda:
Thanks to #1229, `$Path` is now modifiable. But I was wondering if changing `$Path` is reflected in the `Get` command.

Suppose I have `lib/Lib.m`, whose file content is
```mma
42
```

Then, with `wolframscript`,
```mma
AppendTo[$Path, FileNameJoin[{Directory[], "lib"}]];
Print[<< Lib`];
```
or simply
```mma
AppendTo[$Path, "lib"];
Print[<< Lib`];
```
prints the number in `lib/Lib.m`:
```
42
```

But with `mathicsscript`, both give
```
General::noopen: Cannot open Lib`.
```

Version:
```
Mathicscript: 2.1.2, Mathics 2.1.0
on CPython 3.9.4 (default, Apr  7 2021, 05:45:11)
using SymPy 1.7.1, mpmath 1.2.1, numpy 1.20.2
```

rocky:
There are probably a couple of things here. The first bug I'd see is that `Get[]` isn't looking at `$Path` for relative files. In particular:

```mathematica
Print[<< "Lib.m"];
```
isn't resolving the full path for `Lib.m`. 

And the second problem (which I'd look at after fixing the first one), is that "Lib`" is supposed to be the same thing as "Lib.m" in this context. 

As for fixing, I will defer fixing the first problem which seems pretty simple until after https://github.com/mathics/Mathics/pull/1338 gets merged into master, because the file where this bug appears is in that and this will move its location.

The other thing that I would suggest in issues is to give why it might be useful to fix this bug over the swarm of other bugs that are around. If this is part of some open source package that is of interest to general users, then the priority of addressing this is increased as far as I my time goes. If on the other hand this is just something you casually noticed and thought you'd report, we'll thanks for reporting it and it may help others  to be aware that there is a problem. But as far as addressing this, at least as far as my time goes, this is much less important. 

On the other hand if you want to fix, then describing why it might be important to fix is mute. We welcome contributors and would like to see others contribute more.

tueda:
Thank you for looking into this and giving the explanation.

Here I describe some background of when I hit this issue.

[...]

PR #1343, opened 2021-05-02
Fix a number of small bugs

rocky:
* Synchronize `PATH_VAR` to $Path at least before `mathics_open()` is called.
*  Add to_python option string_quotes: boolean. If `False`don't add extra surrounding quotes.
* Remove extra space in list in Main Loop doc

Partially addresses https://github.com/mathics/Mathics/issues/1329

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -2617,7 +2617,8 @@ class Complex(Number):
         return self.real.to_sympy() + sympy.I * self.imag.to_sympy()
 
     def to_python(self, *args, **kwargs):
-        return complex(self.real.to_python(), self.imag.to_python())
+        return complex(self.real.to_python(*args, **kwargs),
+                       self.imag.to_python(*args, **kwargs))
 
     def to_mpmath(self):
         return mpmath.mpc(self.real.to_mpmath(), self.imag.to_mpmath())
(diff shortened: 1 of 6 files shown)


mathics/core/expression.py
```diff
@@ -2962,7 +2963,11 @@ def to_sympy(self, **kwargs):
         return None
 
     def to_python(self, *args, **kwargs) -> str:
-        return '"%s"' % self.value  # add quotes to distinguish from Symbols
+        if kwargs.get("string_quotes", True):
```
mmatera, line 2966: Just in case, I shouldn't be useful to have `string_quotes` with `True` as default value, for backward compatibility?

mmatera (approved):
Apart from the small comment, LGTM

mathics/core/expression.py
rocky: I don't understand. If you haven't set "string_quotes",  it defaults to True.  Almost all of the code currently is this way: no parameter is set.

If you pass this parameter, then you set it accordingly. 

I had thought about a better name, and at some in the past called it something else, but I can't remember the name.

mmatera: You are right! Discard my comment, please.

PR #1256, opened 2021-04-11
Fixcomparison rebase

rocky:
This is for testing mergabiity with master separately from mergability with the fixcomparison branch. 

It will not be merged directly, but the fixcomparison branch with this and further changes is intended to be merged.

PR #1345, opened 2021-05-02
Go over $SystemMemory and MemoryAvailable[]

rocky:
* we were not using the psutil return values in reporting values
* make psutil mandatory rather than optional: it appears to work on all
  the platforms we support.
* Add new variables/functions to `CHANGES.rst`

Fixes #1256


mmatera (approved):
LGTM

PR #1347, opened 2021-05-03
Dispatch2

mmatera:
In this PR, `Dispach` implements a basic "dispatch table" for rules. By now, it just convert `Rule` expressions to `Rule` objects, and store them in objects of a subclass of  `Atom`.

### DIFF
--- a/mathics/builtin/patterns.py
+++ b/mathics/builtin/patterns.py
@@ -38,11 +38,13 @@ The attributes 'Flat', 'Orderless', and 'OneIdentity' affect pattern matching.
 
 
 from mathics.version import __version__  # noqa used in loading to check consistency.
-from mathics.builtin.base import Builtin, BinaryOperator, PostfixOperator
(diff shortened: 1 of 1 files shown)


rocky:
@mmatera  sorry for letting this slip through the cracks. I just rebased off of master so we can benchmark improvments.

rocky:
@mmatera  Looked this over and see that this adds a new feature. So is this ready to be merge in as is? Or are there are there also performance improvements? If the latter, what tests would see use for that?

The general plan is to try to copy mathics/Mathics to Mathics3/mathics-core next weekend. I am trying to address PRs that are viable and in the backlog. 

Also, it would be good not to put in more PR's in this repository if it can instead wait to get to the new repository.

mmatera:
@rocky,

this in principle just defines de functionality. By now, this is just an alias for a list of rules, which was needed for compatibility with certain packages. I guess we can merge this.

PR #1351, opened 2021-05-03
Add a test for $SystemMemory, MemoryAvailable[]...

rocky:
and MemoryInUse[]

P.S. I think MemoryInUse is not quite right.

### DIFF
--- a/mathics/builtin/system.py
+++ b/mathics/builtin/system.py
@@ -485,6 +485,10 @@ class MemoryAvailable(Builtin):
 
     >> MemoryAvailable[]
      = ...
+
+    The relationship between $SystemMemory, MemoryAvailable, and MemoryInUse:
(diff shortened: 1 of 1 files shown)


mathics/builtin/system.py
```diff
@@ -485,6 +485,10 @@ class MemoryAvailable(Builtin):
 
     >> MemoryAvailable[]
      = ...
+
+    The relationship between $SystemMemory, MemoryAvailable, and MemoryInUse:
+    >> $SystemMemory > MemoryAvailable[] > MemoryInUse[]
+     = True
```
rocky, line 491: I suppose they could be equal if all are 0, but that is unthinkable.

mmatera: I like this test!

rocky: Great - then let's ship.

PR #1360, opened 2021-05-07
specify testing pips and have a way to run pytest in parallel

rocky:
Add requirements-dev.txt


mmatera (approved):
LGTM

rocky:
Merging in. But it appears that in order to take advantage of this in CI , I think we need to do more work on tests to ensure that they can work independently and not leave crud around that other parallel test will trip over.

PR #1366, opened 2021-05-09
fix nand and nor latex symbols

mmatera:


mmatera:
Update: the problem is the following. In the title of a section associated with an operator, the character of the operator is put inside a ``\code{\lstinline 'char' }```.  As `\lstinline` expect verbatim code, it fails to understand math mode expressions.
In this PR I just removed `\lstinline` in a way that LaTeX accepts LaTeX code.

rocky:
@mmatera in #1399 I used your observation that the problem was that the text is inside "inline" text and I revised the doc string so that the operators are now outside. In fact there are many places as far as formatting goes that when you have an operator expression like: 

```
$expr1$ \u22BD $expr2$ \u22BD ...
```
I am not seeing any benefit in using the quotes at all. The benefit is only when there is an name like `Nand` (and maybe including the bracket. As a convenience to minimize the use of the quote I think the quote was extended to the closing bracket. 

In that PR I also moved forward in using unicode in more of the operators that have no ASCII equivalent. 

I am (slowly) moving towards the idea of replacing the ASCII operators like `==` too, but really, better would be to hook in all of that with mathics-scanner and have some way for it to fallback to the ASCII operators when the front end doesn't support that.

Should this PR be closed?

PR #1368, opened 2021-05-11
Import HTML - rebasing from PR #644

mmatera:
This PR is just a rebase of #644


mmatera:
@rocky could you help me look at why in the windows CI it fails loading the `fileformats` module?

rocky:
> @rocky could you help me look at why in the windows CI it fails loading the `fileformats` module?

Sure - but in about 8 hours from now.

rocky:
> > @rocky could you help me look at why in the windows CI it fails loading the `fileformats` module?
> 
> Sure - but in about 8 hours from now.

@mmatera The problem is in [setup.py](https://github.com/mathics/Mathics/blob/master/setup.py#L128) 

Although not ideal, this new module needs to be added to this list. (I think there is a way we can use [`findpackages`](https://setuptools.readthedocs.io/en/latest/userguide/package_discovery.html) to better organize this, but I haven't had the opportunity to investigate. 

In sum, this is _not_ a Windows problem per se, but an artifact of the way MS Windows works in that it _first_ installs the code and then runs it. The other OS's I think run out of the source tree. 

So in essence we are catching a problem that folks would notice if they were to install this from PyPI.

One other thing - I looked at the new docstring for `fileformats/__init__.py`. Thanks for adding this, I'd like to make a pass over the wording to tweak it a little before merge.   Thanks!

mmatera:
> > > @rocky could you help me look at why in the windows CI it fails loading the `fileformats` module?
> > 
> > 
> > Sure - but in about 8 hours from now.
> 
> @mmatera The problem is in [setup.py](https://github.com/mathics/Mathics/blob/master/setup.py#L128)
> 
> Although not ideal, this new module needs to be added to this list. I think there is a way we can use [`findpackages`](https://setuptools.readthedocs.io/en/latest/userguide/package_discovery.html) to better organize this, but I haven't had the opportunity to investigate.
> 
> In sum, this is _not_ a Windows problem per se, but an artifact of the way MS Windows works in that it _first_ installs the code and then runs it. The other OS's I think run out of the source tree.
> 
> So in essence we are catching a problem that folks would notice if they were to install this from PyPI.
> 
> One other thing - I looked at the new docstring for `fileformats/__init__.py`. Thanks for adding this, I'd like to make a pass over the wording to tweak it a little before merge. Thanks!

Excellent, you saved for me a couple of hours of research! So, if you can, go ahead over the changes.

PR #1369, opened 2021-05-12
General autoload

rocky:
Needs to be coordinated with https://github.com/Mathics3/mathics-django/pull/41/files


rocky:
> Maybe you need to check if this is defined...

Ok - should be addressed now -- good catch.

PR #1372, opened 2021-05-13
Adjust Degree test

rocky:
Is this consistent with wolframscript?

### DIFF
--- a/mathics/builtin/numbers/constants.py
+++ b/mathics/builtin/numbers/constants.py
@@ -258,12 +258,12 @@ class Degree(MPMathConstant, NumpyConstant, SympyConstant):
     >> Degree == Pi / 180
      = True
 
+    >> N[\\[Degree]] == N[Degree]
+     = True
+
     #> Cos[Degree[x]]
      = Cos[Degree[x]]
 
-    ## Issue 274
-    #> \\[Degree] == ° == Degree
-     = True
 
     #> N[Degree]
      = 0.0174533


mmatera:
It is not, and this is the reason: `º` from the keyboard is a different character for WL than `\[Degree]`, which is `°`. In the terminal, however, both characters looks the same (as the second):

```
 In[1]:= ToCharacterCode["º"]  (*from the keyboard*)
 Out[1]= {186}

 In[2]:= ToCharacterCode["\[Degree]"]
 Out[2]= {176}
```
Both characters have the same graphical representation, but for example, if you copy the output of
`Print["\[Degree]"]` you get `°`

 and paste inside of `ToCharacterCode["°"]` you get `{176}`.

rocky:
> It is not, and this is the reason: `º` from the keyboard is a different character for WL than `\[Degree]`, which is `°`. In the terminal, however, both characters looks the same (as the second):
> 
> ```
>  In[1]:= ToCharacterCode["º"]  (*from the keyboard*)
>  Out[1]= {186}
> 
>  In[2]:= ToCharacterCode["\[Degree]"]
>  Out[2]= {176}
> ```
> 
> Both characters have the same graphical representation, but for example, if you copy the output of
> `Print["\[Degree]"]` you get `°`
> 
> and paste inside of `ToCharacterCode["°"]` you get `{176}`.

Given this, I've adjusted the test to avoid using any sort of character. And I have added this information in mathics-scanner to note that sometimes 186 can appear in the display, while 176 is the correct value both for unicode and wl's unicode. 

On my terminals I consistently see 176, not 186. However, given the fact that it can vary is reason enough not to rely on this in a doctest.

PR #1374, opened 2021-05-13
Easy access to system information

rocky:
This will be used for example in the Django front-end's "about"
page.


rocky:
Merging to make it easier to use on the Django side. Feel free to comment and we can change things down the line.

PR #1377, opened 2021-05-16
use ToString for formatting in Evaluation

mmatera:
In the same direction as PR #1376, this PR starts to decouple formatting routines from the mathics core. 
* `evaluation.format` implements formatting by applying `ToString[HoldForm[], format]` to the result.
* doctest now parses strings in a way that we can use "\[charname]" inside the expected results.

The idea is that, at the end, every conversion be consistent with the output of ToString.

### DIFF
--- a/mathics/builtin/fileformats/htmlformat.py
+++ b/mathics/builtin/fileformats/htmlformat.py
@@ -23,7 +23,6 @@ import sys
 try:
     import lxml.html as lhtml
 except ImportError:
-    print("lxml.html is not available...")
     pass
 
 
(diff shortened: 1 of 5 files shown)


TiagoCavalcante:
@mmatera what is the status of this?

mmatera:
This was also a try to reorganize the way in which formatting is done. The whole question was If the code for formatting  (spread into `Evaluation.evaluate`, `BaseExpression.format`, `BaseExpresion.do_format`, and `BaseExpression.boxes_to_...` )  and the code in `ToString` should be shared, or if they are different things. I think now we have taken a different direction, so this is outdated

PR #1379, opened 2021-05-17
Make Python package lxml optional

### LINKED ISSUE #1378: lxml should be optional
The print there is messing up sage testing. See https://trac.sagemath.org/ticket/31778#comment:13

rocky:
Fixes #1378


rocky:
Will revise after #1380 lands.

PR #1380, opened 2021-05-17
remove warning lxml

mmatera:
This should fix 1379.

### DIFF
--- a/mathics/builtin/fileformats/htmlformat.py
+++ b/mathics/builtin/fileformats/htmlformat.py
@@ -20,11 +20,15 @@ from io import BytesIO
 import re
 import sys
 
+
 try:
     import lxml.html as lhtml
+
+    lhtml_available = True
 except ImportError:
-    print("lxml.html is not available...")
-    pass
+    from mathics.builtin.fileformats.xmlformat import  parse_xml_stream
+
+    lhtml_available = False
 
 
 def node_to_xml_element(node, strip_whitespace=True):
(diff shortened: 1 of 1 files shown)


mmatera:
This should fix the issue

mathics/builtin/fileformats/htmlformat.py
```diff
@@ -89,14 +93,19 @@ class ParseError(Exception):
...
 
     def parse_html_stream(f):
-        parser = lhtml.HTMLParser(encoding="utf8")
-        return lhtml.parse(f, parser)
-
+        if lhtml_available:
+            parser = lhtml.HTMLParser(encoding="utf8")
+            return lhtml.parse(f, parser)
+        else:
+            return parse_xml_stream(f)
```
rocky, line 100: If we can use `parse_xml_stream`, then why do we need `lhtml` at all?

mmatera: because I am not too sure that this really works with any HTML. But in case lhtml is not available, it worth the shot, isn't it?

mmatera: https://stackoverflow.com/questions/47229309/what-are-the-differences-between-lxml-and-elementtree

rocky: Thanks for the information. I followed the link and the link in there to lxml and read all of that. One thing that strikes me is that the information in https://lxml.de/1.3/compatibility.html is from 2008. I am not sure that information and the information from StackOverflow (from 3 years ago) is still relevant since Python continually improves.

As best as I can tell, for Pythoin 3.6 and up, https://docs.python.org/3/library/html.parser.html seems to be the thing to use.

The other thing in the back my mind is that there are some things in #1379 that are not here. But I think the best way to deal with that is finish this, and then I'll go back and revise that for what remains.

mathics/builtin/fileformats/htmlformat.py
```diff
@@ -20,11 +20,15 @@
...
 
+
 try:
     import lxml.html as lhtml
+
+    lhtml_available = True
 except ImportError:
-    print("lxml.html is not available...")
-    pass
+    from mathics.builtin.fileformats.xmlformat import  parse_xml_stream
```
rocky, line 29: I tried using this on the HTML import example and parse_xml_stream wasn't suitable replacement as far as I could tell.

mmatera:
Superseed by #1384

PR #1387, opened 2021-05-22
Make pyston 2.2 compatible

rocky:


mathics/builtin/datentime.py
```diff
@@ -126,7 +126,7 @@ def apply(self, evaluation):
             return SymbolInfinity
 
 
-if sys.platform != "win32":
+if sys.platform != "win32" and ("Pyston" not in sys.version):
```
mmatera, line 129: Pyston does not support threads?

rocky: Pyston does support threads. It SEGV's here and I haven't investigated why. 

But I suspect since this also doesn't work on Microsoft Windows the problem is in TimedConstraint rather than pyston.

PR #1391, opened 2021-05-25
Remove HTML-oriented code.

rocky:
That is now just in the Django copy of this.

This needs to be coordinated with https://github.com/Mathics3/mathics-django/pull/48

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,7 +26,7 @@ _Database_/
 build/
 dist/
 mathics/Makefile
-mathics/doc/tex/data
+mathics/doc/tex/doc_tex_data.pcl
 mathics/doc/tex/documentation.tex
 mathics/doc/tex/images/
 mathics/doc/tex/logo-heptatom.pdf
--- a/mathics/doc/doc.py
+++ b/mathics/doc/doc.py
@@ -1,16 +1,16 @@
 # -*- coding: utf-8 -*-
-"""
-A module which extracts LaTeX from documentation/*.mdoc and from Mathics modules,
+"""A module which extracts LaTeX from documentation/*.mdoc and from Mathics modules,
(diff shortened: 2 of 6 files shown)


mmatera:
@rocky, my guess is that we want to have both kind of doc classes inside Mathics because it could be useful to build the doc pages in other projects. What would be actually good is to disentangle the tex and the HTML generators into two classes/modules, but both inside Mathics.  Then, we can use them for testing, for build the documentation in PDF and HTML, and also for calling them as a backend for the online help in IWolfram and MathicsServer and other possible frontends.

rocky:
> @rocky, my guess is that we want to have both kind of doc classes inside Mathics because it could be useful to build the doc pages in other projects. What would be actually good is to disentangle the tex and the HTML generators into two classes/modules, but both inside Mathics. Then, we can use them for testing, for build the documentation in PDF and HTML, and also for calling them as a backend for the online help in IWolfram and MathicsServer and other possible frontends.

I strongly disagree. Taking docstrings (or however Builtins and Functions are documented) and getting documentation out of them formatted in a way that is desirable for the front end is not a Mathics core function.

Just because X and Y want to use A (HTML) and Z wants to use B (LaTeX), doesn't mean both A and B need to be in Mathics core creating that is the union of everything wanted. We ran into that situation with supporting  Django, asymptote, scikitimag, GNU Readline and so on. 

Instead, A (HTML) is its own package and B (LaTeX) is its own package. And both of those can pull from Mathics core but live outside. 

And by the way we don't have any unit tests for any of the doc transformations. So instead we run a very heavyweight Travis CI and looks for build errors which doesn't mean that there aren't some transformations that are just plain wrong.

The A and B packages would do more extensive (and generally lighter weight) testing.

PR #1393, opened 2021-05-30
Make psutil optional - msys doesn't have

rocky:
.gitignore: Add cygwin DLL

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.c
 *.cpp
+*.dll
 *.egg
 *.gz
 *.nix
--- a/mathics/builtin/system.py
+++ b/mathics/builtin/system.py
@@ -8,7 +8,6 @@ Global System Information
 
 import os
 import platform
-import psutil
 import sys
 import re
 import subprocess
@@ -28,6 +27,13 @@ from mathics.core.expression import (
 from mathics.builtin.base import Builtin, Predefined
 from mathics import version_string
 from mathics.builtin.strings import to_regex
+
+try:
+    import psutil
+except:
+    have_psutil = False
+else:
+    have_psutil = True
 
 
 class Aborted(Predefined):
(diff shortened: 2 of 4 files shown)


mmatera:
@rocky, why do you want to remove this dependency?

rocky:
Not remove it per se but make it optional. 

msys doesn't support it.  Although there are other problems with msys, as I test with more OS's I think it better for Mathics to build more places moving mathics core to be more like a library which can be used more kinds of places. 

I am also seeing a situation where the prompt-toolkit in one of the windows environments where prompt-toolkit isn't available but GNU Readline is.  (And down the line then I guess we also need to allow OSX's lineedit  as an alternative. 

I know this is kind of a reversal of my view from before, but as Mathics has grown and we can add projects like mathics-omnibus on the outside to manage packaging pieces, my thoughts here have changed.

In sum at this point I think it better to allow the various pieces to be installed in more minimal environments.

Your thoughts?

mmatera:
@rocky, thanks for the explanation. So, the idea is that if the `psutil` library is not available (like in windows), the Builtins that depend on it do not load, and then we also avoid running the doctests on them, isn't it? The other way we used to handle these optional libraries was to check into the ``apply`` method if they are available, and return `$Failed` if the library is not available, which of course is a problem in doctests.

mmatera (approved):
LGTM

rocky:
> @rocky, thanks for the explanation. So, the idea is that if the `psutil` library is not available (like in windows), the Builtins that depend on it do not load, and then we also avoid running the doctests on them, isn't it? 

Yes, they don't run doctests. 

> The other way we used to handle these optional libraries was to check into the `apply` method if they are available, and return `$Failed` if the library is not available, which of course is a problem in doctests.

I will keep that in mind. The situation with doctest for me is a little out of control. 

BTW I will be checking to see if the RealDigits tests that are not part of examples can be moved to unit tests to speed this up.

PR #1394, opened 2021-05-30
Move many RealDigits tests to unit tests

rocky:


mmatera (approved):
LGTM. In any case, it would be interesting to see why the same evaluation that runs fast in the front-end, takes so much time in the doctest engine...

rocky:
> LGTM. In any case, it would be interesting to see why the same evaluation that runs fast in the front-end, takes so much time in the doctest engine...

The doctest  clearly does more and can test error output. But I suspect we have memory problems as well.  

 Just one more problem to add to the long list of things to look at and address.

mmatera:
> > LGTM. In any case, it would be interesting to see why the same evaluation that runs fast in the front-end, takes so much time in the doctest engine...
> 
> The doctest clearly does more and can test error output. But I suspect we have memory problems as well.

I checked that, and the time is consumed by the evaluation, not by the subsequent checks. It seems to be a memory issue.

> 
> Just one more problem to add to the long list of things to look at and address.

This is just something to have in mind for a future revision of the global performance...

PR #1401, opened 2021-06-02
mathics/test.py -> docbuild.py and ...

rocky:
* Add --latex-only to write just the LaTeX file
* Go over to make more clear what this big mess is doing.


Makefile
```diff
@@ -73,11 +73,11 @@ gstest:
 
 #: Create data that is used to in Django docs and to build TeX PDF
 doc-data mathics/doc/tex/doc_tex_data.pcl: mathics/builtin/*.py mathics/doc/documentation/*.mdoc mathics/doc/documentation/images/*
-	$(PYTHON) mathics/test.py -ot -k
+	$(PYTHON) mathics/docbuild.py -ot -k
```
rocky, line 76: Ideally this would be called `doc_test_and_build.py` but this is too long. And we already have a `doctest.py` I can't think of a better short name without using more jargon.

PR #1409, opened 2021-06-06
Asymptote and other tweaks

rocky:


rocky:
We can't do the thing we really want to do for the asymptote formatter - use "circle" instead of dot so that the points are scalable, but this gets us closer. 

So adding this to master, and then will come back at it some other time.

PR #1420, opened 2021-06-12
Change double quote to single quote

TiagoCavalcanteTrindade:
Solve [the problem mentioned here](https://github.com/Mathics3/mathics-django/pull/74#issuecomment-860109856).

### DIFF
--- a/mathics/builtin/drawing/graphics3d.py
+++ b/mathics/builtin/drawing/graphics3d.py
@@ -462,7 +462,7 @@ class Graphics3DBox(GraphicsBox):
         We include enclosing script tagging.
         """
         json_repr = self.boxes_to_json(leaves, **options)
-        js = f'<graphics3d data="{json_repr}"/>'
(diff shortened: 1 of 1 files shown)


mathics/builtin/drawing/graphics3d.py
```diff
@@ -462,7 +462,7 @@ def boxes_to_js(self, leaves=None, **options):
         We include enclosing script tagging.
         """
         json_repr = self.boxes_to_json(leaves, **options)
-        js = f'<graphics3d data="{json_repr}"/>'
```
rocky, line 465: Better is to use some sort of triple quoting which almost certainly can't appear inside json_repr

TiagoCavalcanteTrindade: @rocky HTML just accept single or double quote. The `json_repr` don't contain double quotes.

rocky: Not a big deal. I'll adjust it down the line.

Issue #1009, opened 2020-11-15
Catch and Throw not implemented


mmatera:
These functions are necesary for PR #995.

mmatera:
Implemented in #1025

TiagoCavalcanteTrindade:
@mmatera what still need to be done?

Issue #904, opened 2020-09-18
Update Three.js from version r52 to something closer to r120


rocky:
There are probably lots of improvements that have occurred in the last 7 years. Staring in r50 the name changes. See https://github.com/mrdoob/three.js/releases/r50 for example.

mmatera:
We could try to just change the version and see if it does work. If not, I would keep the current version for a while...

rocky:
I tried something like that, but I could have done it wron.. It doesn't work. 

You probably need to do read the release notes, and do it step by step or several step by  steps.

mmatera:
What I can tell you is that once I tried to hack the code for the Iwolfram, and it was a non trivial task, even for the basic functionality. The interface has a lot of work on the top of tree.js, and I didn't expect that update it be an easy task...

rocky:
Thanks for the information. This could be a very long-term project. I am hoping someone with an interest with this sort of stuff will come a long and help out.

slel:
In SageMath we are lucky to have @paulmasson and recently @jcamp0x2a
contribute many Three.js upgrades and related improvements.

If someone works on Three.js in Mathics, studying what is done
in Sage may provide inspiration. Here are some links.

- [Sage 9.2 release tour](https://wiki.sagemath.org/ReleaseTours/sage-9.2): section [Graphics](https://wiki.sagemath.org/ReleaseTours/sage-9.2#Graphics)
- [Sage 9.1 release tour](https://wiki.sagemath.org/ReleaseTours/sage-9.1): section [Improvements in the Three.js 3D viewer](https://wiki.sagemath.org/ReleaseTours/sage-9.1#Improvements_in_the_three.js_3D_viewer)
- [Sage 9.0 release tour](https://wiki.sagemath.org/ReleaseTours/sage-9.0): section [Three.js is the new default 3D viewer](https://wiki.sagemath.org/ReleaseTours/sage-9.0#Three.js_is_the_new_default_3D_viewer)
- [Sage tickets about Three.js](https://trac.sagemath.org/query?order=id&desc=1&summary=~threejs&or&summary=~three.js&col=ticket&col=summary&col=reporter&col=author&col=reviewer&col=type&col=status&col=priority&col=milestone)
- [three.js in sage-devel](https://groups.google.com/forum/#!searchin/sage-devel/three.js%7Csort:date)
- [three.js in sage-release](https://groups.google.com/forum/#!searchin/sage-release/three.js%7Csort:date)
[...]

PR #1432, opened 2021-06-20
Fix Throw and Catch tests

TiagoCavalcanteTrindade:
The tests of `Throw` and `Catch` were written like: `<< test` not `>> test`.

Also shouldn't #1009 be closed?

Maybe #904 too?

### DIFF
--- a/mathics/builtin/control.py
+++ b/mathics/builtin/control.py
@@ -70,6 +70,8 @@ class CompoundExpression(BinaryOperator):
     ## Issue 531
     #> z = Max[1, 1 + x]; x = 2; z
      = 3
+
+    #> Clear[x]; Clear[z]
     """
 
     operator = ";"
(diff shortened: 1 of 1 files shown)


rocky:
> The tests of `Throw` and `Catch` were written like: `<< test` not `>> test`.
> 
> Also shouldn't #1009 be closed?

I recall @mmatera had a reason for keeping this open, but I could be wrong.

rocky:
Ok - let's try these doc changes.

rocky:
@TiagoCavalcanteTrindade  See 7acedec6

Note that you can't just cut and paste WL documentation.  It is quirky and I don't like it, but Single quote is for `<code>` which can include some (but not all) math expressions. `$` encloses math expressions. We don't use backtick. 

Some of this is described in https://mathics-development-guide.readthedocs.io/en/latest/extending/developing-code/extending/documentation-markup.html

TiagoCavalcanteTrindade:
@rocky I'll change it, I just modified the `<<` to `>>` and changed the tests until the pass.

rocky:
The commit mentioned has already been committed.

PR #1433, opened 2021-06-20
Improve Range performance a bit

TiagoCavalcanteTrindade:


rocky:
How does one see  what improvement is gained by this?

TiagoCavalcanteTrindade:
@rocky I removed a function call.

rocky:
> @rocky I removed a function call.

Whey do we think that the conversion from sympy call isn't needed?

TiagoCavalcanteTrindade:
@rocky ok, this makes sense.

Issue #1064, opened 2020-12-24
Implement WL Show


rocky:
See https://reference.wolfram.com/language/ref/Show.html?q=Show

This is used in Combinatorica.

In the Django app when you do this: 

```
Needs["DiscreteMath`CombinatoricaV0.9`"]
ShowGraph[Star[6]]
```

![Mathics-Combinatorica](https://user-images.githubusercontent.com/8851/103070559-be489600-458f-11eb-969d-e1f9000dbe8c.png)

rocky:
Not sure when this got fixed, but it is fixed.

PR #1438, opened 2021-06-22
Fix String.to_sympy()

### LINKED ISSUE #1129: to_sympy(String) returns None; use repr for to_python()
```python
class String(Atom):
    ...
    def get_string_value(self) -> str:
        return self.value

    def to_sympy(self, **kwargs):
        return None

    def to_python(self, *args, **kwargs) -> str:
        return '"%s"' % self.value  # add quotes to distinguish from Symbols

```

_to_sympy()_ should probably return `self.value`. Also I think there shoudl be a parameter to `to_python()` that indicates whether you want quotes or not. And _to_python()_ is wrong since repr should be used to handle the fact that the string might have embedded quotes.

TiagoCavalcanteTrindade:
Fixes #1129.

Also, @rocky shouldn't #1064 be closed?

### DIFF
--- a/mathics/core/expression.py
+++ b/mathics/core/expression.py
@@ -2958,8 +2958,8 @@ class String(Atom):
     def get_string_value(self) -> str:
         return self.value
 
-    def to_sympy(self, **kwargs):
-        return None
+    def to_sympy(self, **kwargs) -> str:
+        return self.value
 
     def to_python(self, *args, **kwargs) -> str:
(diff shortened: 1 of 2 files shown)


TiagoCavalcanteTrindade:
@rocky I looked at the tests and the FAILs are happening because of the string isn't the same than its respective SymPy object, but is there any SymPy's string that isn't an expression?

PR #1439, opened 2021-06-22
Remove System`$Notebooks from settings.m

TiagoCavalcanteTrindade:
Remove System`$Notebooks from settings.m as all front ends already implement it

### DIFF
--- a/mathics/__init__.py
+++ b/mathics/__init__.py
@@ -50,6 +50,18 @@ using SymPy {sympy}, mpmath {mpmath}, numpy {numpy}""".format(
 if "cython" in version_info:
     version_string += f", cython {version_info['cython']}"
 
+
+def load_default_settings_files(definitions, load_cli_settings: bool = True):
+    import os.path as osp
+    from mathics.core.definitions import autoload_files
+
+    root_dir = osp.realpath(osp.dirname(__file__))
+
+    autoload_files(definitions, root_dir, "autoload", False)
+    if load_cli_settings:
+        autoload_files(definitions, root_dir, "autoload-cli", False)
+
+
 license_string = """\
 Copyright (C) 2011-2021 The Mathics Team.
 This program comes with ABSOLUTELY NO WARRANTY.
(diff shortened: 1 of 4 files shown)


rocky (changes requested):
This was no doubt a TODO item until we got all the setting in place in the other repositories and pushed. 

We should add a test however for the mathics CLI for it settings. You can find a corresponding one over in the mathicsscript project. 

Would you please add the test (adjusting accordingly)?  I suppose we could test that Notebooks is universally false too.

TiagoCavalcanteTrindade:
@rocky I added the test, there's a TODO saying that we should do more, but what?

rocky:
> @rocky I added the test, there's a TODO saying that we should do more, but what?

I don't know what you are talking about. Everything now looks fine. Thanks!

rocky (approved):
LGTM  after CI clears.

TiagoCavalcanteTrindade:
@rocky [here](https://github.com/Mathics3/mathicsscript/blob/master/test/test_settings.py#L6) is the FIXME.

rocky:
> @rocky [here](https://github.com/Mathics3/mathicsscript/blob/master/test/test_settings.py#L6) is the FIXME.

@TiagoCavalcanteTrindade  Ah - ok. thanks. Here I just tested that settings were set to _some_ value.  In those cases where we think values will change a bit over time (e.g. if a version number were set) then this is the best thing.

For things like Notebook, we _know_ this will never be the case so that should probably test false explicitly.

For some things, were a value might change like a version but we know the datatype, then we should test on the datatype.

That's the kind of thing I meant.

TiagoCavalcanteTrindade:
@rocky ok, I'll do this test.

rocky (changes requested):
The whole test/test_settings.py test I now learn is a mess. (And yes, I was the one who created this out of ignorance).

It should be revised to autoload and at least test the values are defined which means the a value different from the setting name comes back. Possibly with a particular value or a particular type.

rocky (approved):
It's been a long time coming, but finally! Great work!

TiagoCavalcanteTrindade:
Yes, after a lot of force pushes it's done!

PR #1451, opened 2021-06-25
Add Cylinder function

TiagoCavalcanteTrindade:
Rebase of #1450.

Closes #1294.

### DIFF
--- a/mathics/format/asy.py
+++ b/mathics/format/asy.py
@@ -19,6 +19,7 @@ from mathics.builtin.box.graphics import (
 
 from mathics.builtin.box.graphics3d import (
     Graphics3DElements,
+    Cylinder3DBox,
     Line3DBox,
     Point3DBox,
     Polygon3DBox,
@@ -149,6 +150,38 @@ def bezier_curve_box(self, **options) -> str:
 
 
 add_conversion_fn(BezierCurveBox, bezier_curve_box)
+
+
+def cylinder3dbox(self, **options) -> str:
+    if self.face_color is None:
+        face_color = (1, 1, 1)
+    else:
+        face_color = self.face_color.to_js()
+
+    asy = ""
+    i = 0
+    while i < len(self.points) / 2:
+        asy += "draw(surface(cylinder({0}, {1}, {2}, {3})), rgb({2},{3},{4}));".format(
+            tuple(self.points[i * 2].pos()[0]),
+            self.radius,
+
+            # distance between start and end
+            (
+                (self.points[i * 2][0][0] - self.points[i * 2 + 1][0][0])**2 +
+                (self.points[i * 2][0][1] - self.points[i * 2 + 1][0][1])**2 +
+                (self.points[i * 2][0][2] - self.points[i * 2 + 1][0][2])**2
+            ) ** 0.5,
+
+            (1, 1, 0), # FIXME: currently always drawing around the axis X+Y
+            *face_color[:3]
+        )
+
+        i += 1
+
+    return asy
+
+
+add_conversion_fn(Cylinder3DBox)
 
 
 def filled_curve_box(self, **options) -> str:
@@ -310,7 +343,7 @@ def polygon3dbox(self, **options) -> str:
 add_conversion_fn(Polygon3DBox)
 
 
-def polygonbox(self, **options):
+def polygonbox(self, **options) -> str:
     line_width = self.style.get_line_width(face_element=True)
     if self.vertex_colors is None:
         face_color = self.face_color
--- a/mathics/builtin/box/graphics3d.py
+++ b/mathics/builtin/box/graphics3d.py
@@ -694,12 +694,85 @@ currentlight=light(rgb(0.5,0.5,1), specular=red, (2,0,2), (2,2,2), (0,2,2));
         ]
 
 
-class Point3DBox(PointBox):
+class Cylinder3DBox(_Graphics3DElement):
+    """
+    Internal Python class used when Boxing a 'Cylinder' object.
+    """
+
+    def init(self, graphics, style, item):
+        super(Cylinder3DBox, self).init(graphics, item, style)
+
+        self.edge_color, self.face_color = style.get_style(_Color, face_element=True)
+
+        if len(item.leaves) != 2:
+            raise BoxConstructError
+
+        points = item.leaves[0].to_python()
+        if not all(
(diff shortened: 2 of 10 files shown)


TiagoCavalcanteTrindade:
@rocky could you review my changes? I put a FIXME [here](https://github.com/TiagoCavalcanteTrindade/Mathics/blob/add-cylinder-function/mathics/format/asy.py#L453) I don't know to transform the current data in a *draw around this axis*.

mathics/format/asy.py
```diff
@@ -310,7 +311,7 @@ def polygon3dbox(self, **options) -> str:
 add_conversion_fn(Polygon3DBox)
 
 
-def polygonbox(self, **options):
+def polygonbox(self, **options) -> str:
```
rocky, line 314: Routines to draw SVG are no longer here but are in `format/svg.py`

TiagoCavalcanteTrindade: Oh, ok I gonna rebase it.

mathics/format/asy.py
```diff
@@ -427,3 +428,34 @@ def sphere3dbox(self, **options) -> str:
 
 
 add_conversion_fn(Sphere3DBox)
+
+
+def cylinder3dbox(self, **options) -> str:
```
rocky, line 433: Also move  - here into `format/asy.py`

mathics/format/asy.py
rocky: rebase doesn't track refactoring like this.  You will probably have to move it by hand.

rocky:
> @rocky could you review my changes? I put a FIXME [here](https://github.com/TiagoCavalcanteTrindade/Mathics/blob/add-cylinder-function/mathics/format/asy.py#L453) I don't know to transform the current data in a _draw around this axis_.

Right now there isn't a transform function, but there should be one. In #1373 there is code that needs to be adapted [here](https://github.com/mathics/Mathics/blob/0a283aa975225d2c6ed04b09feab3d219ec4774c/mathics/builtin/tensors.py#L498-L583)

TiagoCavalcanteTrindade:
@rocky you think it's better we do that before we merge?

Also, what do you think about the [odd number of coordinates handling](https://github.com/Mathics3/mathics-django/pull/31#issuecomment-868875962)?

rocky:
Here is a different idea for how to approach this and at the same time go in the direction we want to go in....

Write a `to_json()` routine for this. But actually, I'd start out with sphere, since that might be easier and here we want to focus on the JSON aspect. 

Then we should start a graphics-backend module. (In the past, it was erroneously called a pymathics module, but I realize this was wrong. This is a pure module  _of any kind_ that takes JSON with our simple graphics API and converts it into threejs.   You could start out with `graphics3d.js` and/or `mathics.js` from the Django repository and turn that into a nodejs module (or however you want to package it).

rocky:
> @rocky you think it's better we do that before we merge?
> 
> Also, what do you think about the [odd number of coordinates handling](https://github.com/Mathics3/mathics-django/pull/31#issuecomment-868875962)?

Release 3.0.0 of Mathics core and Django are currently _out_. Look on PyPI or in "releases" of this project.

rocky:
> Release 3.0.0 of Mathics core and Django are currently _out_. Look on PyPI or in "releases" of this project.

I am still working on mathics omnibus though.

TiagoCavalcanteTrindade:
> > @rocky you think it's better we do that before we merge?
> > Also, what do you think about the [odd number of coordinates handling](https://github.com/Mathics3/mathics-django/pull/31#issuecomment-868875962)?
> 
> Release 3.0.0 of Mathics core and Django are currently _out_. Look on PyPI or in "releases" of this project.

@rocky I didn't find the information I need here.

> Write a to_json() routine for this. But actually, I'd start out with sphere, since that might be easier and here we want to focus on the JSON aspect.

Ok, I guess it should be in another PR, isn't it?

> Then we should start a graphics-backend module. (In the past, it was erroneously called a pymathics module, but I realize this was wrong. This is a pure module of any kind that takes JSON with our simple graphics API and converts it into threejs. You could start out with graphics3d.js and/or mathics.js from the Django repository and turn that into a nodejs module (or however you want to package it).

@rocky if you are talking about keeping the transformations in 2 places I think there's no problem, but otherwise  I'm afraid it has too many layers in the code.

rocky:
> > Release 3.0.0 of Mathics core and Django are currently _out_. Look on PyPI or in "releases" of this project.
> 
> @rocky I didn't find the information I need here.

I don't understand what information you need. 

All I was saying is that there are new releases: https://pypi.org/project/Mathics3/ (pypi) and https://github.com/mathics/Mathics/releases/tag/3.0.0

What 
> 
> > Write a to_json() routine for this. But actually, I'd start out with sphere, since that might be easier and here we want to focus on the JSON aspect.
> 
> Ok, I guess it should be in another PR, isn't it?
> 

Yes, all of this is a new PR. New approach, new PR. And let's try to start simple to work up to cylinders which add this additional wrinkle.

TiagoCavalcanteTrindade:
> I don't understand what information you need.

Sorry for the confusion.

The information I was needing was what we should do when the user does something like:
```wl
Cylinder[{{0,0,0}}] (* it should raise an error because of the user didn't provided an end point, just the start point *)
```

But you already answered [here](https://github.com/Mathics3/mathics-django/pull/31#issuecomment-869073001).

TiagoCavalcanteTrindade:
I left a FIXME saying we need to do the following `transformation_vector * (radius, radius, radius)` instead of `(1, 1, 1) * (radius, radius, radius)`. The current version does the plot range be bigger than it should.

As we are going to do the transformation part in other PR I guess this should be merged before that.

Also, the following is the behavior of an odd number of points: 
image

rocky (approved):
LGTM - I wonder why we get the spurious black style wars. 

Did you run `black .` ?

TiagoCavalcanteTrindade:
@rocky no, I'm doing that (I've forgotten).

rocky:
I suspect then it is the black version that is used. What does `black --version` say?

TiagoCavalcanteTrindade:
@rocky I don't have git pre-commit installed, I guess it's that.

My black version is 21.6b0.

rocky:
Ok - Let's set everything to 21.6b0. 

I will add a check to the workflows too.

Issue #717, opened 2017-10-03
Subsets missing


fpe-pedraza:
The Mathematica Subsets[] command, for combinations generation, is missing.

rocky:
@fpe-pedraza What is an example of such an invocation that doesn't work?

Issue #715, opened 2017-08-28
Simplify changes parameter


wcgbg:
`Simplify[t1-t2]` changes the value of `t2`. See how `t2` changes in `Out[4]`, `Out[6]`, and `Out[8]`.

```
$ mathics

Mathics 1.0
on CPython 2.7.12 (default, Nov 19 2016, 06:48:10) 
using SymPy 1.0, mpmath 0.19

Copyright (C) 2011-2016 The Mathics Team.
This program comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under certain conditions.
See the documentation for the full license.

Quit by pressing CONTROL-D

In[1]:= abc=Solve[{a*s0*s0+b*s0+c==p0,a*s1*s1+b*s1+c==p1,a*s2*s2+b*s2+c==p2},{a,b,c}][[1]]
Out[1]= {a -> (p0 (s1 - s2) - p1 (s0 - s2) + p2 (s0 - s1)) / (s0 ^ 2 s1 - s0 ^ 2 s2 - s0 s1 ^ 2 + s0 s2 ^ 2 + s1 ^ 2 s2 - s1 s2 ^ 2), b -> (-p0 (s1 ^ 2 - s2 ^ 2) + p1 (s0 ^ 2 - s2 ^ 2) - p2 (s0 ^ 2 - s1 ^ 2)) / (s0 ^ 2 s1 - s0 ^ 2 s2 - s0 s1 ^ 2 + s0 s2 ^ 2 + s1 ^ 2 s2 - s1 s2 ^ 2), c -> (p0 s1 s2 (s1 - s2) - p1 s0 s2 (s0 - s2) + p2 s0 s1 (s0 - s1)) / (s0 ^ 2 s1 - s0 ^ 2 s2 - s0 s1 ^ 2 + s0 s2 ^ 2 + s1 ^ 2 s2 - s1 s2 ^ 2)}

In[2]:= t1=Simplify[(s1-s0)*(2*a*s0+b)/.abc]
Out[2]= (-2 p0 s0 s1 + 2 p0 s0 s2 + p0 s1 ^ 2 - p0 s2 ^ 2 + p1 s0 ^ 2 - 2 p1 s0 s2 + p1 s2 ^ 2 - p2 s0 ^ 2 + 2 p2 s0 s1 - p2 s1 ^ 2) / (s0 s1 - s0 s2 - s1 s2 + s2 ^ 2)

[...]

Issue #712, opened 2017-08-05
--execute does not work as expected


Yurlungur:
When I run 

```shell
mathics -e "Print[1+1];" -script myfile.mathics
```

where `myfile.mathics` simply contains the line `Print["Hello"];` I get

```shell
In[1]:= Print[1+1];
6
```

And `myfile.mathics` does *not* execute. This seems counter to the information printed with `mathics -h`. What *should* happen and what I would like to be able to do is that `Print[1+1];` should be executed and then `myfile.mathics` should be run as a script.

I'm happy to help fix this if I can figure out where this functionality is being implemented. Can you point me to the right part of the code?

GarkGarcia:
Hey @Yurlungur. If you're still interested, the bug is probably somewhere in [main.py](https://github.com/mathics/Mathics/blob/master/mathics/main.py).

I'm not sure the bug is still present thought. I'll investigate it.

Yurlungur:
Wow. I'd completely forgotten about this. Sure, I'm still interested, if the bug is still present. I'll take a look.

rocky:
@Yurlungur 

Here is what I am currently getting: 

```
$ mathics -e "Print[1+1];" -script /myfile.mathics
2
None
```
Personally I find mxing `-e` and `-script` and `Print` weird. What was it you were really trying to do?

Yurlungur:
Honestly, it's been so long that I don't remember why I wanted to do this. An easy fix might be to just put that one can't mix these flags into the help message and move on.

TiagoCavalcanteTrindade:
@Yurlungur I'm fixing it.

rocky:
Personally, I'd like to understand the use case and then document this assuming it is useful. Or if it matches the behavior of the WL command line options that is useful to know too. 

Also a unit test would be nice.

Issue #662, opened 2016-12-10
`Repeated[a]`  causes infinite recursion


jlapeyre:
```
In[1]:= Repeated[a]
$RecursionLimit::reclim: Recursion depth of 200 exceeded.
Out[1]= $Aborted
```
Mathics 1.0
on CPython 2.7.12 |Continuum Analytics, Inc.| (default, Jul  2 2016, 17:42:40) 
using SymPy 1.0, mpmath 0.19

rocky:
I can't remember when, but I believe @mmatera fixed this a while back. We now get: 

```
In[2]:= Repeated[a]
Out[2]= a..

```

PR #1453, opened 2021-06-27
Improve 'Show' documentation

TiagoCavalcanteTrindade:
@rocky #36 is done, but the behavior of `Show` looks strange to me (it shows the list of graphics and the options too).

Also, shouldn't #301, #662, #712, #715, and #717 be closed?


rocky (approved):
LGTM

PR #1454, opened 2021-06-27
Implement Factorial2

### LINKED ISSUE #653: Factorial2 does not simplify
For integral numerical arguments ≥ -1, at least, this should be easy: reusing existing implementation a Factorial2[n] is Factorial[n/2] * 2^(n/2) for n even and Factorial[n+1] / Factorial2[n+1] for n odd (reducing to previous case). Yet inputs 1!!, 2!!, 3!!, 4!!, ... stay unevaluated.

TiagoCavalcanteTrindade:
Closes #653.

I set the precedence to the same as the `Factorial` precedence, but I don't know if it is right.

### DIFF
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -1861,6 +1861,67 @@ class Factorial(PostfixOperator, _MPMathFunction):
     mpmath_name = "factorial"
 
 
+class Factorial2(PostfixOperator, _MPMathFunction):
+    """
+    <dl>
+      <dt>'Factorial2[$n$]'
+      <dt>'$n$!!'
+      <dd>computes the double factorial of $n$.
+    </dl>
+
+    >> 5!!
+     = 15
+
+    >> Factorial2[-3]
+     = -1
+
+    'Factorial2' also accepts irrational numbers:
+    >> N[Pi!!, 6]
+     = 3.35237
+
+    'Factorial2' also accepts complex numbers:
+    >> N[I!!, 6]
+     = 2.71713 + 0.279527 I
+
+    'Factorial2' is not defined for even negative numbers.
+    #> (-4)!!
+     : Factorial2 is not defined for even negative numbers.
+     = -4!!
+    """
+
+    operator = "!!"
+    precedence = 610
+    mpmath_name = "fac2"
+    sympy_name = "factorial2"
+
+    messages = {
+        "ndf": "Factorial2 is not defined for even negative numbers.",
+        "unknownp": "'`1`' not in ('Automatic'', 'sympy', 'numpy', 'mpmath')",
+    }
+
+    def apply(self, number, evaluation, options={}):
+        "Factorial2[number_?NumberQ, OptionsPattern[%(name)s]]"
+
+        preference = self.get_option(options, "Method", evaluation).get_string_value()
+
+        if preference not in ("Automatic", "sympy", "numpy", "mpmath"):
+            return evaluation.message("Factorial2", "unknownp", preference)
+
+        try:
+            if isinstance(number, Integer):
+                return from_sympy(getattr(sympy, self.sympy_name)(number.to_sympy()))
+            else:
+                return from_mpmath(
+                    getattr(mpmath, self.mpmath_name)(number.to_mpmath())
+                )
+        except:
+            if number.to_python() < 0 and number.to_python() % 2 == 0:
+                # even negative number
+                return evaluation.message("Factorial2", "ndf")
+
+            raise
+
+
 class HarmonicNumber(_MPMathFunction):
     """
     <dl>
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,13 @@
 CHANGES
 =======
 
+3.0.1
+-----
+
+New functions:
+- Cylinder
+- Factorial2 (!!)
+
 3.0.0
 -----
 


rocky (commented):
We should probably have at least one doc test here. 

If you feel the need to test edge cases, then this would go in `test/test_arithmetic.py`.

Note that down the line we'll be picking up the operator and precedence from a JSON table produced by the data in the mathics-scanner project.

TiagoCavalcanteTrindade:
> We should probably have at least one doc test here.

Does the following don't count as a test?https://github.com/TiagoCavalcanteTrindade/Mathics/blob/factorial2/mathics/builtin/arithmetic.py#L1872-L1873

> If you feel the need to test edge cases, then this would go in test/test_arithmetic.py.

@rocky I guess that if it has more tests they should be visible to the user.

> Note that down the line we'll be picking up the operator and precedence from a JSON table produced by the data in the mathics-scanner project.

@rocky should I add it in `tokeniser.py`'s `tokens`?

mathics/builtin/arithmetic.py
```diff
@@ -1861,6 +1861,23 @@ class Factorial(PostfixOperator, _MPMathFunction):
     mpmath_name = "factorial"
 
 
+class Factorial2(PostfixOperator, _MPMathFunction):
+    """
+    <dl>
+    <dt>'Factorial2[$n$]'
+    <dt>'$n$!!'
```
rocky, line 1868: BTW I prefer indenting the `<dt>`s inside the `<dl>` like the way other XML tags work.   Yeah, I don't know why we don't add the ends to some things like `</dt>` or `</dd>` or `</li>`. 

It would be nice also to say _what_ Factorial2 is.

mathics/builtin/arithmetic.py
```diff
@@ -1861,6 +1861,23 @@ class Factorial(PostfixOperator, _MPMathFunction):
...
+        <dd>computes the double factorial of $n$.
+    </dl>
+
+    >> 5!!
+     = 15
+    """
+
+    operator = "!!"
+    precedence = 610
```
rocky, line 1877: I can't find mention of the precedence either. 

This hasn't been added to mathics-scanner, so it would be nice to add this into the table there. 

The sympy equivalent is also `factorial2` so that should be listed as well.

mathics/builtin/arithmetic.py
TiagoCavalcanteTrindade: > BTW I prefer indenting the <dt>s inside the <dl> like the way other XML tags work. 

I too, I just made it because I saw this indentation elsewhere.

> Yeah, I don't know why we don't add the ends to some things like </dt> or </dd> or </li>.

Adding them work too.

> It would be nice also to say what Factorial2 is.

Maybe `<dd>computes the double factorial of $n$.`?

mathics/builtin/arithmetic.py
TiagoCavalcanteTrindade: Ok.

rocky:
> 
> 
> > We should probably have at least one doc test here.
> 
> Does the following don't count as a test?https://github.com/TiagoCavalcanteTrindade/Mathics/blob/factorial2/mathics/builtin/arithmetic.py#L1872-L1873

Yes, you are correct - this is sufficient. My mistake. 

> 
> > If you feel the need to test edge cases, then this would go in test/test_arithmetic.py.
> 
> @rocky I guess that if it has more tests they should be visible to the user.
> 
> > Note that down the line we'll be picking up the operator and precedence from a JSON table produced by the data in the mathics-scanner project.
> 
> Should I add it in `tokeniser.py`'s `tokens`?

It is already there on line 135. It is not in the character table `mathics_scanner/data/named-characters.yml`.

TiagoCavalcanteTrindade:
@rocky adding the SymPy equivalent doesn't work (SymPy's factorial2 doesn't accept negative either complex numbers).

I still didn't understand what the precedence should be.

I don't know what put in `unicode-equivalent-name`, my current code is the following: (named-characters.yml)

```yml
Factorial2:
  ascii: "!!"
  has-unicode-inverse: false
  is-letter-like: false
  operator-name: Factorial2
  precedence: 610
  unicode-equivalent: "!!"
  unicode-equivalent-name: EXCLAMATION MARK
  wl-unicode: "!!"
  wl-unicode-name: EXCLAMATION MARK
```

rocky:
> 
> 
> @rocky adding the SymPy equivalent doesn't work (SymPy's factorial2 doesn't accept negative either complex numbers).
> 

Hmm.... there are a couple of options here.  Implement an `apply()`  and filter out the kinds of things sympy doesn't handle. I see this is also in scipy. Scipy is an optional dependency. I haven't looked to see how we handle scipy. 

> I still didn't understand what the precedence should be.
> 
> I don't know what put in `unicode-equivalent-name`, my current code is the following: (named-characters.yml)

In the comments at the top of the YAML:


> #   unicode-equivalent: A unicode equivalent for the named-character, if it
> #                       exists.
> #
> #   unicode-equivalent-name: The name of the unicode equivalent, if it exist

So if there is none don't add keys for these. 

> 
> ```yaml
> Factorial2:
>   ascii: "!!"
>   has-unicode-inverse: false
>   is-letter-like: false
>   operator-name: Factorial2
>   precedence: 610
>   unicode-equivalent: "!!"
>   unicode-equivalent-name: EXCLAMATION MARK
>   wl-unicode: "!!"
>   wl-unicode-name: EXCLAMATION MARK
> ```

rocky:
I should also mention that I have started (but nowhere near finished) adding AMLaTeX equivalents into the table so following https://tex.stackexchange.com/questions/308540/unicode-math-and-double-factorial

It looks like there should be 

```
amslatex: "{!}{!}"
```

TiagoCavalcanteTrindade:
> So if there is none don't add keys for these. Similarly for the wolfram unicode equivalents

Ok, I didn't see it.


> It looks like there should be
> ```
> amslatex: "{!}{!}"
> ```

I didn't understand if I should add that.

rocky:
@TiagoCavalcanteTrindade one other small thing.  When new builtins get added, it would be helpful to mention that in `CHANGES.rst`  (This makes it easier on whoever is doing the release.)

So I guess not only one here but also one in Cylinder as well. (And if you want to change the indention on `dt` there, I wouldn't mind. I noticed, but wasn't going to be too Fascist about it.)

TiagoCavalcanteTrindade:
@rocky I saw that in some other functions there isn't a SymPy version. Should we add it here?

rocky:
> @rocky I saw that in some other functions there isn't a SymPy version. Should we add it here?

I think so, but probably put that in a branch because I am not totally sure of the ramifications, (but want to find out).

TiagoCavalcanteTrindade:
@rocky sorry, I wasn't clear. I don't understand the necessity of a SymPy version, it would increase the code complexity.

Also, where should I add the changes in `CHANGES.rst`? 3.1.0?

rocky:
> @rocky sorry, I wasn't clear. I don't understand the necessity of a SymPy version, it would increase the code complexity.

Sympy can handle symbolic expressions generally and MPMath can't.  So in theory Sympy either now or in the future might be able to handle `Pi !!`  MPMath can only give a numeric approximation. 

> 
> Also, where should I add the changes in `CHANGES.rst`? 3.1.0?

Please start with `3.0.1` the version determination if made more close to the time of release, or of necessity. Adding a new function doesn't change the middle number.

rocky:
> > @rocky sorry, I wasn't clear. I don't understand the necessity of a SymPy version, it would increase the code complexity.
> 
> Sympy can handle symbolic expressions generally and MPMath can't. So in theory Sympy either now or in the future might be able to handle `Pi !!` MPMath can only give a numeric approximation.
> 
> > Also, where should I add the changes in `CHANGES.rst`? 3.1.0?
> 
> Please start with `3.0.1.` The version determination is made more close to the time of release, or at the time when it is necessary. Adding a new function doesn't change the middle number.

Another aspect that has been in the back of my mind is in programs that want to autotranslate from Mathics to Sympy might be able to use the mapping information. Ideally again this goes in a table like something found in mathics-scanner right now. However at least recording this here helps when the information is aggregated elsewhere.

TiagoCavalcanteTrindade:
@rocky ok. But one thing: even with SymPy disabled `Pi!!` returns `Pi!!`, I must pass it to `N` to show a numeric value.

And how should we do it? I was thinking something like this:
```py
def apply:
    if is positive:
        try:
            import sympy
            return sympy.factorial2(number)
        catch:
             mpmath.fac2(number)
```

What do you think?

Also, maybe close https://github.com/Mathics3/mathics-scanner/issues/15?

TiagoCavalcanteTrindade:
@rocky I'm thinking that adding SymPy support for `Factorial2` won't help because SymPy's `factorial2` don't support non-integers (as `Pi`, `E` or `I`).

rocky:
> @rocky I'm thinking that adding SymPy support for `Factorial2` won't help because SymPy's `factorial2` don't support non-integers (as `Pi`, `E`, or `I`). Anyway, if you think is better to add SymPy support here I can do it.

Have you timed the two routines for speed? Have you analyzed the precision and accuracy of both methods? 

I don't the @TiagoCavalcanteTrindade or any of us should be in the business of deciding what is best when it comes to numerical evaluation.  (If you have a Ph.d in this stuff or can show experience and research in this, then let's talk).

The folks that wrote this code probably know more than we do. We have ways (or should have) ways to allow the user to indicate what she wants. I view my job is just to do the plumbing to make what's available available. 

And as I said just noting correspondences between the systems will help us down the line.

rocky:
> > So if there is none don't add keys for these. Similarly for the wolfram unicode equivalents
> 
> Ok, I didn't see it.
> 
> > It looks like there should be
> > ```
> > amslatex: "{!}{!}"
> > ```
> 
> I didn't understand if I should add that.

Yes, please add this.

TiagoCavalcanteTrindade:
> I don't think that the @TiagoCavalcanteTrindade or any of us should be in the business of deciding what is best when it comes to numerical evaluation.

Yes, I guess you're right.

> The folks that wrote this code probably know more than we do.

I guess you're right here too, so I gonna [try to] implement the SymPy support tomorrow.

Just a note about the precision: the double factorial of a positive integer is an integer, so the precision should be 100%, and as SymPy's `factorial2` just allow integer numbers I guess it doesn't any method with square roots or something with high precision lose. If it uses the most common formula (https://en.wikipedia.org/wiki/Double_factorial#Negative_arguments) precision won't be a problem as the unique operation that loses precision here is division.  
About the performance, SymPy is probably faster, but the manner the things are done in Mathics Core right now the difference between SymPy and MPMath isn't perceptible.

rocky:
> > I don't think that the @TiagoCavalcanteTrindade or any of us should be in the business of deciding what is best when it comes to numerical evaluation.
> 
> Yes, I guess you're right.

In my opinion it takes a lot of courage and integrity that I admire to be able to admit you are wrong. 

In terms of the facts, as best as I understand you _are_ correct. _Right now_ sympy seems to indicate it uses the same method described in Wikipedia. 

However there are other things at play here which drive this in a different direction. mpmath recently has been pretty much on hold. sympy is not in that state. So who knows what will happen in the future here? In the past we've with calculations that were iterated we've come across the issue of being able to guarantee the right number of digits in precision. 

As you say that "shouldn't" be a problem here. But there have been too many things that "shouldn't" have been a problem that were problems.. And this is not something I really want to spend time thinking about when there is a blanket, simple process that seems to cover everything: when one of our libraries implements something, note that an allow it to go through. 

There is code in number/constants.py that handles calculating the same thing several ways, and maybe that can be consulted here. 

> 
> > The folks that wrote this code probably know more than we do.
> 
> I guess you're right here too, so I gonna [try to] implement the SymPy support tomorrow.
> 
> Just a note about the precision: the double factorial of a positive integer is an integer, so the precision should be 100%, and as SymPy's `factorial2` just allow integer numbers I guess it doesn't any method with square roots or something with high precision lose. If it uses the most common formula (https://en.wikipedia.org/wiki/Double_factorial#Negative_arguments) precision won't be a problem as the unique operation that loses precision here is division.
> About the performance, SymPy is probably faster, but the manner the things are done in Mathics Core right now the difference between SymPy and MPMath isn't perceptible.

I

TiagoCavalcanteTrindade:
@rocky should I write `factorial2` and `fac2` twice (in the `_name` attributes and in the call) or use `getattr`?

Also, is my `ndf` message good?

mathics/builtin/arithmetic.py
```diff
@@ -1861,6 +1861,54 @@ class Factorial(PostfixOperator, _MPMathFunction):
...
+        "Factorial2[number_?NumberQ]"
+
+        if isinstance(number, Integer):
+            if number.to_python() < 0 and number.to_python() % 2 == 0:
+                # even negative number
+                return evaluation.message("Factorial2", "ndf")
+
+            return from_sympy(sympy.factorial2(number.to_sympy()))
+        else:
+            return from_mpmath(mpmath.fac2(number.to_mpmath()))
```
rocky, line 1909: Look at numbers/constants.py for how we handle which routine to use when there are several backends  like mpmath and sympy for handling.

TiagoCavalcanteTrindade: Ok.

mathics/builtin/arithmetic.py
```diff
@@ -1861,6 +1861,60 @@ class Factorial(PostfixOperator, _MPMathFunction):
...
+      <dd>computes the double factorial of $n$.
+    </dl>
+
+    >> 5!!
+     = 15
+
+    >> Factorial2[-3]
+     = -1
+
+    'Factorial2' also accept irrational numbers:
```
rocky, line 1878: Small grammar typo here and below "also accept" -> "also accepts"

mathics/builtin/arithmetic.py
```diff
@@ -1861,6 +1861,60 @@ class Factorial(PostfixOperator, _MPMathFunction):
...
+    }
+
+    def apply(self, number, evaluation, preference="sympy"):
+        "Factorial2[number_?NumberQ]"
+
+        if not (preference in ("sympy", "numpy", "mpmath")):
+            return evaluation.message("Factorial2", "unknownp", preference)
+
+        if isinstance(number, Integer) and (preference != "mpmath"):
+            if number.to_python() < 0 and number.to_python() % 2 == 0:
```
rocky, line 1909: Something to think about.  Another possibility is to pass to sympy and report back errors it gives.

That way if/when sympy handles negative even numbers, we will pick that up and not have to code here as we would do by being proactive about this. 

Not a big deal, but something to keep in mind.

mathics/builtin/arithmetic.py
TiagoCavalcanteTrindade: Again, thanks!

mathics/builtin/arithmetic.py
TiagoCavalcanteTrindade: @rocky the error is not so user-friendly. It says it's in a pole of the gamma function.

rocky: Ok but you could still trap the error and _then_ check for negativity. And if it's not that, use whatever sympy has to say.

Again that way, if sympy adapts, you do. And having code in place to trap for the error might be used if there are _other_ kinds of messages you want to redo.

TiagoCavalcanteTrindade: @rock I don't understand why do it.

My current approach is the following:
```py
def apply(self, number, evaluation):
    "Factorial2[number_?NumberQ]"

    if number.to_python() < 0 and number.to_python() % 2 == 0:
        # even negative number
        return evaluation.message("Factorial2", "ndf")

    if isinstance(number, Integer):
        return from_sympy(getattr(sympy, self.sympy_name)(number.to_sympy()))
    else:
        return from_mpmath(getattr(mpmath, self.mpmath_name)(number.to_mpmath()))
```

For either SymPy or MPMath the double factorial of a negative odd leads to an error.

rocky: In https://en.wikipedia.org/wiki/Double_factorial negative even is defined. It is just not _currently_ implemented. 

Do you know it will _never_  be implemented in _either_ of these? 

Again, I'd like to think of this code as not offering opinions so much as just acting as a pass through to an unbounded open set of other libraries. We let the other libraries be the experts on the thing they do. (But there is more than one so, we don't really have to rely on that totally). 

The attitude that this code is going to implement all of graphics routines, or its own documentation system has just got us into trouble time and time again.

TiagoCavalcanteTrindade: @rocky I'm saying that if we write this and then one of them implements it we would remove the `ndf` error, so we're going to change the code anyway.

rocky: But users get the benefit right away. Removing what will be dead code has little impact here. 

But whatever - beggars can't be choosers. I suppose this is fine and it can be redone in the future if desired.

TiagoCavalcanteTrindade: @rocky I'm afraid that first trying and then checking for the negativity would be too complex (and it'd be done twice), and the difficulty of removing the `ndf` is going to be smaller than removing the check.

Anyway, you're working in Mathics for much more time than me, so I guess your decision is going to be the best. What do you say?

rocky: > @rocky I'm afraid that first trying and then checking for the negativity would be too complex (and it'd be done twice), 

It's exactly the same code test: `if number.to_python() < 0 and number.to_python() % 2 == 0:`

It would just be moved whereever failure is detected. 

When there is an error very little is being done so I don't see it as a big deal. And the second test is there just to provide a more sensible error message. 

> and the difficulty of removing the `ndf` is going to be smaller than removing the check.
> 
> Anyway, you're working in Mathics for much more time than me, so I guess your decision is going to be the best. What do you say?

Since seems to be a big deal and it should be a big deal either way, just do it however you want. We can adjust later if it becomes important, which I don't think it will.

rocky: But one thing please keep in mind because it has come up time and time again. Don't try to second guess the libraries. If you want to improve them, go to the respective repositories and suggest patches and PRs. And in that process you may come to learn more about why things are the way they are and subtleties that you may have missed. 

I go about changing or "fixing" the sympy/numpy/mpmath library code very humbly.

TiagoCavalcanteTrindade: @rocky the try didn't increase the complexity of the code too much (to say the truth it didn't at all, the pseudocode in my mind was much more complex).

About fixing MPMath, this sounds like a great idea! I'll give it a try.

TiagoCavalcanteTrindade:
@rocky we both understand it wrong, (-2n)!! is really undefined:
image

rocky:
https://en.wikipedia.org/wiki/Double_factorial#Negative_arguments indicates
that there is a way to define this.

mathics/builtin/arithmetic.py
rocky: Ok - I'll look at this when I get a chance.

TiagoCavalcanteTrindade:
@rocky yes, I deleted the message because I saw this latter, in my first plot I thought that a ±∞ appeared, but it was just the plot range 😆.

image

mathics/builtin/arithmetic.py
rocky: > Ok - I'll look at this when I get a chance.

https://github.com/mathics/Mathics/pull/1459 is what i have. It is a bit messier than I had hoped because there are bugs outside of just the Factorial2 part.

But this brings up another point when you just jump off into an areas that you've never looked at before. Be prepared to find bugs and uncover a new can of worms. 

As you are aware, we're trying to work through all of these. But the misfeature are still vast.

TiagoCavalcanteTrindade:
Superpassed by #1459.

PR #1455, opened 2021-06-28
Fix weird results when combining -script and -e

### LINKED ISSUE #712: --execute does not work as expected

TiagoCavalcanteTrindade:
Fixes #712.

### DIFF
--- a/test/data/script.m
+++ b/test/data/script.m
@@ -1,0 +1,1 @@
+Print["Hello"];
--- a/test/test_cli.py
+++ b/test/test_cli.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+import subprocess
+
+import os.path as osp
+import re
+import pytest
+import sys
+
+
+def get_testdir():
+    filename = osp.normcase(osp.dirname(osp.abspath(__file__)))
+    return osp.realpath(filename)
+
+
+@pytest.mark.skipif(sys.version_info < (3, 7), reason="requires Python 3.7 or higher")
+def test_cli():
+    script_file = osp.join(get_testdir(), "data", "script.m")
+
+    # asserts output contains 'Hello' and '2'
+    assert re.match(
+        r"Hello\s+2",
+        subprocess.run(
+            ["mathics", "-e", "Print[1+1];", "-script", script_file],
+            capture_output=True,
+        ).stdout.decode("utf-8"),
+    )
+
+
+if __name__ == "__main__":
+    test_cli()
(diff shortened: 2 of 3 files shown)


test/data/script.m
```diff
@@ -0,0 +1 @@
+Print["Hello"];
```
rocky, line 1: It would be awesome if you could get your editor to add newlines at the end of a file.

rocky (approved):
LGTM

test/data/script.m
TiagoCavalcanteTrindade: I forgot to format the file.

TiagoCavalcanteTrindade:
Looks like that `capture_output` was added in Python 3.7.

rocky:
> 
> 
> Looks like that `capture_output` was added in Python 3.7.

I guess we can skip the test when Python < 3.7.

TiagoCavalcanteTrindade:
Ok, how I do it?

I thought in jump the assert if the version (which I was going to get by `os` module) is smaller than 7, but I guess there's a better way.

rocky:
> Ok, how I do it?
> 
> I thought in jump the assert if the version (which I was going to get by `os` module) is smaller than 7, but I guess there's a better way.

See 

https://stackoverflow.com/questions/38442897/how-do-i-disable-a-test-using-pytest#:~:text=Skipping%20a%20test%20based,or%20higher%22)%0Adef%20test_func_one()%3A%0A%20%20%20%20...

TiagoCavalcanteTrindade:
@rocky maybe skep if is Windows?

test/test_cli.py
```diff
@@ -0,0 +1,30 @@
...
+    return osp.realpath(filename)
+
+
+@pytest.mark.skipif(sys.version_info < (3, 7), reason="requires Python 3.7 or higher")
+def test_cli():
+    script_file = osp.join(get_testdir(), "data", "script.m")
+
+    # asserts output contains 'Hello' and '2'
+    assert re.match(
+        r"Hello[ \n\t]*2",
```
rocky, line 21: On WIndows `\r` is needed, but probably better to use `\s`

TiagoCavalcanteTrindade: Yes, makes sense.

PR #1456, opened 2021-06-28
Fix missing elements of `StringSplit`

### LINKED ISSUE #969: Wrong behavior of  StringSplit with WhitespaceCharacter
```StringSplit["13  a22    bbb", WhitespaceCharacter] // InputForm```
should be: 
```
{"13", "", "a22", "", "", "", "", "bbb"}
```

TiagoCavalcante:
Fixes #969.

The side effect is that the whitespaces are always shown:
image

Maybe change `MathMLForm`?

### DIFF
--- a/mathics/builtin/strings.py
+++ b/mathics/builtin/strings.py
@@ -532,7 +532,7 @@ class WhitespaceCharacter(Builtin):
      = True
 
     >> StringSplit["a\nb\r\nc\rd", WhitespaceCharacter]
-     = {a, b, c, d}
+     = {a, b, , c, d}
 
     For sequences of whitespace characters use 'Whitespace':
     >> StringMatchQ[" \n", WhitespaceCharacter]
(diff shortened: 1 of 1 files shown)


rocky:
My take is that the display on Dango is a separate issue. I am okay with this going in like this.

@mmatera - you thoughts?

TiagoCavalcante:
@rocky I have been thinking about this and the commas in Mathics Django shouldn't be ommited.

rocky:
> @rocky I have been thinking about this and the commas in Mathics Django shouldn't be omitted.

All I care about is that we are consistent with whatever Mathematica does. Does this PR change that?

TiagoCavalcante:
> All I care about is that we are consistent with whatever Mathematica does. Does this PR change that?

Well, according to the [documentation](https://reference.wolfram.com/language/ref/WhitespaceCharacter.html?q=WhitespaceCharacter) yes.

rocky:
@TiagoCavalcante Thanks for the information. Here's my take.  If this can wait until after we do the move to Mathics3/mathics-core (which will happen next weekend) then fix up the conflict and you'll be able to push to a branch on that.

TiagoCavalcante:
> @TiagoCavalcante Thanks for the information. Here's my take. If this can wait until after we do the move to Mathics3/mathics-core (which will happen next weekend) then fix up the conflict and you'll be able to push to a branch on that.

Ok, until that I'll fix it so it works in the Python 3.9 (I remember this was working before...)

TiagoCavalcante:
Oh, my mistake. This still is working, I wrote the query wrongly.

TiagoCavalcante:
Superseded by [mathics-core#5](https://github.com/Mathics3/mathics-core/pull/5).

PR #1462, opened 2021-07-01
Asy exporter

mmatera:
@rocky : This would be a very basic implementation of the exporter. This works both to produce a string (what you want here->https://github.com/Mathics3/mathicsscript/pull/44) and to export to a file.
Examples:

```
In[9]:= ExportString[Graphics,"asy"]
Out[9]:= "
        
        usepackage(\"amsmath\");
        size(6.6667cm, 2.9762cm);
        label(\"$\text{Graphics}$\", (200.0,89.28571428571429), (0,0), rgb(0, 0, 0));
        clip(box((172,76.786), (228,101.79)));
        
        "

In[10]:= ExportString[Graphics[{Disk[]}],"asy"]
Out[10]:= "
         
         usepackage(\"amsmath\");
         size(5.8333cm, 5.8333cm);
         filldraw(ellipse((175,175),175,175), rgb(0, 0, 0), nullpen);
         clip(box((0,0), (350,350)));
         
         "
```

### DIFF
--- a/mathics/autoload/formats/Asy/Export.m
+++ b/mathics/autoload/formats/Asy/Export.m
@@ -1,0 +1,31 @@
+(* Text Exporter *)
+
+Begin["System`Convert`Asy`"]
+
+Options[AsyExport] = {
+		      "CharacterEncoding" :> $CharacterEncoding, (*Not used by now...*)
+};
+
+AsyExport[strm_OutputStream, expr_, OptionsPattern[]]:=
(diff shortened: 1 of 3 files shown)


rocky:
@mmatera  Great! Let's do it after CI clears. Later on we should fix up Mathics Core so that we don't produce the `\begin{asy}` and `\end{asy}`. But that can come later.

Issue #1452, opened 2021-06-27
requirements-full.txt not included in released source code


vicp74:
There's a [similar issue as in mathics-scanner](https://github.com/Mathics3/mathics-scanner/issues/29): requirements-full.txt is not present in the released [Mathics3-3.0.0.tar.gz](https://github.com/mathics/Mathics/releases/download/3.0.0/Mathics3-3.0.0.tar.gz), but in this case it is present in [Source code (zip or tar.gz)](https://github.com/mathics/Mathics/archive/refs/tags/3.0.0.tar.gz). Can it also be included in the first one?

rocky:
Should be there now  in mathics-scanner 1.2.3.

vicp74:
I mean the problem is **also** present in Mathics. Please reread my previous message without taking into account mathics-scanner.

rocky:
Ok - thanks for the clarification. Next week I'll try to release another Mathics version.

rocky:
@vicp74 @thierry-FreeBSD --- I think in the last release this is now fixed.

vicp74:
Yes, it's fixed. Thanks!
This issue can be closed now.

PR #1465, opened 2021-07-02
Rename `faceColor`

TiagoCavalcanteTrindade:
Rename `faceColor` to `color` (in the JSONfy part of the graphics).  
And uncapitalize the lights' names.

### DIFF
--- a/mathics/builtin/box/graphics3d.py
+++ b/mathics/builtin/box/graphics3d.py
@@ -373,7 +373,7 @@ class Graphics3DBox(GraphicsBox):
         json_repr = format_fn(elements, **options)
 
         # TODO: Cubeoid (like this)
-        # json_repr = [{'faceColor': (1, 1, 1, 1), 'position': [(0,0,0), None],
+        # json_repr = [{'color': (1, 1, 1, 1), 'position': [(0,0,0), None],
         # 'size':[(1,1,1), None], 'type': 'cube'}]
 
         json_repr = json.dumps(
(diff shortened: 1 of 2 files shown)


rocky:
A couple of things here. First this needs to be coordinated with changes on all of the front-end side. (A release is coming up over the weekend so now is not a good time for this.)

As suggested by https://github.com/JerryI/Mathematica-ThreeJS-graphics-engine perhaps we are already doing too much work on the Core side and should offload that the renderer part of the frontend. 

And if that is done, we keep the WL/Mathics names which means keeping the capitalization.

There may be a need or desire to have a lower-level JSON representation which makes the job of the JSON to graphics library job easier.  But still it feels to me that we should still follow WL/Mathics style names where possible.

TiagoCavalcanteTrindade:
@rocky so we should capitalize the primitives.

rocky:
> @rocky so we should capitalize the primitives.

Not just captialize them but use the corresponding names as well. If there are things that need to get filled in that obviously will never be part of the graphics description (these are low-level things like a translation function name of a WL concept to a three-js concept), then you get to choose whatever you want and use whatever style you want.

rocky:
Note: Release as promised is underway. The main motivation for this was to address #1452. (And I happen to have time this weekend). This change is API breaking and while useful, needs to be delayed a little.

TiagoCavalcanteTrindade:
Strange, the tests pass locally.

PR #1484, opened 2021-07-09
Reinstate Density plot for SVG - rough cut

rocky:
Back in a less-hacky way and where we *describe* what's up.

Looks are a little ugly but can probably be fixed using rectangular gradients described in comments in density_plot_box().

Also, we now produce more human-readable SVG. Since this is produced by hand, it is useful to facilitate humans debugging this.

### DIFF
--- a/mathics/format/svg.py
+++ b/mathics/format/svg.py
@@ -1,6 +1,4 @@
 # -*- coding: utf-8 -*-
-import json
-
 """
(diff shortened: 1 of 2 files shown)


rocky:
This works as is and is already a little large (if not ideal). So I am merging in, but I will also continue to improve this in further PRs.

PR #1492, opened 2021-07-21
Graphics3d breaking changes

TiagoCavalcanteTrindade:
image

### DIFF
--- a/mathics/builtin/box/graphics3d.py
+++ b/mathics/builtin/box/graphics3d.py
@@ -373,7 +373,7 @@ class Graphics3DBox(GraphicsBox):
         json_repr = format_fn(elements, **options)
 
         # TODO: Cubeoid (like this)
-        # json_repr = [{'faceColor': (1, 1, 1, 1), 'position': [(0,0,0), None],
+        # json_repr = [{'color': (1, 1, 1, 1), 'position': [(0,0,0), None],
         # 'size':[(1,1,1), None], 'type': 'cube'}]
 
         json_repr = json.dumps(
--- a/mathics/builtin/box/graphics3d.py
+++ b/mathics/builtin/box/graphics3d.py
@@ -303,12 +303,6 @@ class Graphics3DBox(GraphicsBox):
                             light["target"][j] * boxscale[j] for j in range(3)
                         ]
 
-                # Rescale viewpoint
-                self.viewpoint = [
-                    vp * max([xmax - xmin, ymax - ymin, zmax - zmin])
-                    for vp in self.viewpoint
-                ]
-
             w = 0 if (xmin is None or xmax is None) else xmax - xmin
             h = 0 if (ymin is None or ymax is None) else ymax - ymin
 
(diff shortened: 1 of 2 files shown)


mathics/builtin/box/graphics3d.py
```diff
@@ -384,14 +375,6 @@ def boxes_to_json(self, leaves=None, **options):
                     "ticks": ticks,
                     "ticks_style": js_ticks_style,
                 },
-                "extent": {
```
rocky, line 387: Can't you just ignore it in the other side?

mathics/builtin/box/graphics3d.py
```diff
@@ -63,40 +63,40 @@ def _prepare_elements(self, leaves, options, max_width=None):
 
         if lighting == "System`Automatic":
             self.lighting = [
-                {"type": "Ambient", "color": [0.3, 0.2, 0.4]},
+                {"type": "ambient", "color": [0.3, 0.2, 0.4]},
```
rocky, line 66: Isn't this handled by the transformation layer in mathics-graphics3d-backend ?

mathics/builtin/box/graphics3d.py
TiagoCavalcanteTrindade: Yes, but why keep it there?

mathics/builtin/box/graphics3d.py
TiagoCavalcanteTrindade: Yes, but is weird to have primitive names lower-case and light names capital-case.

rocky: I wasn't saying this is wrong - just that mathics-threejs-backend _also_ needs to be able to handle things like this. 

The current plan is to release Mathics Django as is. And then _after_ that settles we will a slew of API breaking changes.

In general things, have breaking way too much recently. We need more control everywhere on limiting  breakage.

TiagoCavalcanteTrindade: Ok, so just remove the viewpoint scaling by now as it isn't a breaking change?

mathics/builtin/box/graphics3d.py
rocky: Same as above. It will change and should. But not before the Mathics-Django release.

rocky: Probably a good way to approach this is to start going over the API doc in mathics-threejs-backend review that and then come back here to make it match that specification. 

But all of this is major API upheaval. I'd like to have at least one more stable release before that.

mathics/builtin/box/graphics3d.py
rocky: With all of the changes it is hard to follow this PR.  It feels like after things settle down we should plain on a major Graphics3D in compact JSON design.  

Personally, I think that all of the design work and implementation in mathics-threejs-backend could go on now if you are so disposed. Nothing prevents creating whatever API  is best and converting what we have now to that.

In the PRs here in the short term we should separate breaking API from non-breaking. 

Adding fields to existing JSON is fine that isn't going to break older Mathics-Django that happen use Mathics inore dev. changing existing values would.  And having mathics-threejs-backend ignore existing fields is fine too.

PR #1500, opened 2021-07-25
Remove viewpoint rescale

TiagoCavalcanteTrindade:
A change in mathics-threejs-backend needs to be done in parallel to this.

### DIFF
--- a/mathics/builtin/box/graphics3d.py
+++ b/mathics/builtin/box/graphics3d.py
@@ -303,12 +303,6 @@ class Graphics3DBox(GraphicsBox):
                             light["target"][j] * boxscale[j] for j in range(3)
                         ]
 
-                # Rescale viewpoint
-                self.viewpoint = [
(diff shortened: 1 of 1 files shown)


mathics/builtin/box/graphics3d.py
```diff
@@ -303,12 +303,6 @@ def calc_dimensions(final_pass=True):
                             light["target"][j] * boxscale[j] for j in range(3)
                         ]
 
-                # Rescale viewpoint
```
rocky, line 306: Recall Asymptote is another backend. If boxing is changed, that too needs to be adjusted. 

Better would be some way for the backend to specify its preferences and do this conditionally.

TiagoCavalcanteTrindade: > Recall Asymptote is another backend. If boxing is changed, that too needs to be adjusted.

I already fixed it.

> Better would be some way for the backend to specify its preferences and do this conditionally.

For now, I moved this code to another place.

rocky:
LGTM

PR #1510, opened 2021-07-31
Read list refactor

rocky:
This is a rebase of some old code. I got maybe 80% of the way a while ago. But  this is still needed. 

Start at `py.test test/test_importexport.py`

and `fix read_in_stream()`


rocky:
This PR has been long and painful. And the code still suffers from the mess that was there previously. 

However rather than try to fix this in one go, I am merging this in and will come back in another PR to clean this up. 

Having a basis for proper stream where we manage stream numbers and close streams properly is benefit enough in of itselfs.

PR #1512, opened 2021-08-01
Fix bug in session.get_settings()..

rocky:
When we don't have a setting defined, session.get_settings should return
None.


mmatera (commented):
LGTM

PR #1527, opened 2021-08-06
Add Coordinates class

TiagoCavalcante:
Not working, but a start point.

### DIFF
--- a/mathics/builtin/box/graphics3d.py
+++ b/mathics/builtin/box/graphics3d.py
@@ -727,14 +727,7 @@ class Cuboid3DBox(_Graphics3DElement):
         if len(item.leaves) != 1:
             raise BoxConstructError
 
-        points = item.leaves[0].to_python()
-        if not all(
-            len(point) == 3 and all(isinstance(p, numbers.Real) for p in point)
-            for point in points
-        ):
-            raise BoxConstructError
-
-        self.points = [Coords3D(graphics, pos=point) for point in points]
+        self.points = [Coords3D(graphics, pos=point) for point in item.leaves[0]]
 
     def extent(self):
         return [coords.pos()[0] for coords in self.points]
--- a/mathics/builtin/drawing/graphics3d.py
+++ b/mathics/builtin/drawing/graphics3d.py
@@ -8,6 +8,7 @@ Functions for working with 3D graphics.
 from mathics.version import __version__  # noqa used in loading to check consistency.
 
 from mathics.core.expression import Expression, Real
+from mathics.core.coordinates import Coordinates
 
 from mathics.builtin.base import BoxConstructError, Builtin, InstanceableBuiltin
 from mathics.builtin.colors.color_directives import RGBColor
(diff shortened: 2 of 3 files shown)


mathics/builtin/box/graphics3d.py
```diff
@@ -727,14 +727,7 @@ def init(self, graphics, style, item):
...
 
-        points = item.leaves[0].to_python()
-        if not all(
-            len(point) == 3 and all(isinstance(p, numbers.Real) for p in point)
-            for point in points
-        ):
-            raise BoxConstructError
-
-        self.points = [Coords3D(graphics, pos=point) for point in points]
+        self.points = [Coords3D(graphics, pos=point) for point in item.leaves[0]]
```
rocky, line 730: It feels like something is wrong here in having to propagate `graphics` down to each point rather than storing it once in say self.

rocky: I am wondering if we we shouldn't rename the `Coords3D` and `Coord` to something like `GraphicsCoords3D` and `GraphicsCoord2D`  or maybe just one `GraphicsCoord` with a dimension. to distinguish the fact that this kind of coordinate belongs in a some sort of graphics and therefore has a `graphics` field and it is meaningful for this kind of point to be scaled and for example. 

The new coordinate class in contrast, does not care whether its is placed in a graphic image or not.

mathics/builtin/drawing/graphics3d.py
```diff
@@ -243,29 +250,47 @@ class Cuboid(Builtin):
...
+                        Real(coordinates.coordinates[0][1] + 1),
+                    ),
+                )
+
+            return Expression(
+                "Cuboid",
+                List(
+                    List(*(coordinates.coordinates[0])),
+                    List(
+                        Real(coordinates.coordinates[0][0] + 1),
```
rocky, line 282: Personally I would prefer `.x`, `.y` and `.z` properties on the coordinates. Using the properties disallows setting them, and it is clearer. 

I see this kind of thing happen a lot in mathics-threejs-backend and thought the same thing.

mathics/builtin/drawing/graphics3d.py
```diff
@@ -243,29 +250,47 @@ class Cuboid(Builtin):
...
+                    ),
+                ),
+            )
+
+        if len(coordinates) % 2 == 1:
             # The number of points is odd, so abort.
             evaluation.error("Cuboid", "oddn", positions)
 
-        return Expression("Cuboid", positions)
+        return Expression("Cuboid", coordinates)
```
rocky, line 293: I think here we will also want to convert everything back to mathics, rather than use coordinates.

mathics/core/coordinates.py
```diff
@@ -0,0 +1,16 @@
...
+class Coordinates:
+    def __init__(self, coordinates):
+        self.coordinates = coordinates.to_python()
+
+        if not all(2 <= len(coordinate) <= 3 for coordinate in self.coordinates):
+            raise Exception
+
+    @property
+    def dimension(self):
+        return len(self.coordinates[0])
```
rocky, line 13: I think there should be a class for `Coordinate` as well.

rocky (commented):
Overall a great start. Thanks!

rocky:
https://github.com/mathics/Mathics/pull/1373/files points out a change we should make. 

Whenever we see `value.leaves[x].round_to_float()`, that should be instead  `value.leaves[x].to_mpmath()`,

TiagoCavalcante:
@rocky seems that it would just slow Mathics down. I'm doing tests, and removing `Coords3D` is giving positive results.

rocky:
> @rocky seems that it would just slow Mathics down. I'm doing tests, and removing `Coords3D` is giving positive results.

OK - thanks for testing this!

TiagoCavalcante:
@rocky I had tested that without Cythonizing. That slow us down if Cython is enabled.

rocky:
@TiagoCavalcante Feel free to reinstate this on the other side.

PR #1547, opened 2021-08-14
UniformPolyhedra

rocky:
Work remains to sets center point and "radius".
    
Also, UniformPolyhedron doesn't take the above parameters directly while  the named versions, e.g. Octahedron do allow these parameters - work out how we want to represent this then.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -31,7 +31,7 @@ Gamma functions:
 
 Uniform Polyhedron
 * ``Dodecahedron``
-* ``Icosohedron``
+* ``Icosahedron``
 * ``Octahedron``
 * ``TetraHedron``
 * ``UniformPolyedron``
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,12 @@ Gamma functions:
 * ``PolyGamma``
 * ``Stieltjes``
 
+Uniform Polyhedron
+* ``UniformPolyedron``
+* ``Dodecahedron``
+* ``Octahedron``
+* ``TetraHedron``
+
 
 Bugs
 ++++
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,10 +30,11 @@ Gamma functions:
 * ``Stieltjes``
 
 Uniform Polyhedron
-* ``UniformPolyedron``
 * ``Dodecahedron``
+* ``Isohedron``
 * ``Octahedron``
 * ``TetraHedron``
+* ``UniformPolyedron``
 
 
 Bugs
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -31,7 +31,7 @@ Gamma functions:
 
 Uniform Polyhedron
 * ``Dodecahedron``
-* ``Isohedron``
+* ``Icosohedron``
 * ``Octahedron``
 * ``TetraHedron``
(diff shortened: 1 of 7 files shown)


mathics/format/json.py
```diff
@@ -208,3 +210,23 @@ def sphere_3d_box(self) -> list:
...
+
+def uniform_polyhedron_3d_box(self) -> list:
+    face_color = self.face_color
+    if face_color is not None:
+        face_color = face_color.to_js()
+    data = [
+        {
+            "type": "uniformPolyhedron",
+            "color": face_color[:3],
+            "faceColor": face_color,
```
TiagoCavalcanteTrindade, line 223: This won't be used in mathics-threejs-backend.

CHANGES.rst
```diff
@@ -29,6 +29,12 @@ Gamma functions:
 * ``PolyGamma``
 * ``Stieltjes``
 
+Uniform Polyhedron
+* ``UniformPolyedron``
+* ``Dodecahedron``
+* ``Octahedron``
+* ``TetraHedron``
+
```
TiagoCavalcanteTrindade, line 37: Maybe add icosahedron?

mathics/format/json.py
rocky: I am not sure which field exactly you mean, but all of these are used. I tried setting facecolor and that works. 

BTW, since faceColor is there,  wiki entry for UniformPolyhedra in mathics-threejs-backend probably should be updated to include this field.

CHANGES.rst
rocky: Sorry, I missed that. Much of that is in the next commit ae3c7ec  but I missed something here for this to work.

Please just add PRs on top of this if you find the mistake.

mathics/format/json.py
TiagoCavalcanteTrindade: `faceColor` should be replaced in all places by `color`, it works because of the translation layer.

mathics/format/json.py
```diff
@@ -116,7 +116,7 @@ def cylinder_3d_box(self):
         [self.points],
         "cylinder",
         face_color,
-        {"faceColor": face_color, "radius": self.radius},
```
TiagoCavalcanteTrindade, line 117: Great! One less thing for the translation layer.

rocky: But please check the other places in this file,  json.py, to make sure they all should be changed too.

Again, just add a PR on top of this.

mathics/builtin/drawing/uniform_polyhedra.py
```diff
@@ -96,11 +111,15 @@ class Tetrahedron(Builtin):
       <dd>a regular tetrahedron centered at the origin with unit edge length.
     </dl>
 
-    >> Graphics3D[Tetrahedron[]]
+    >> Graphics3D[Tetrahedron[{{0,0,0}, {1,1,1}}, 2]]
```
TiagoCavalcanteTrindade, line 114: The edge length difference is only seen when you show the axes or other elements in the scene.

rocky: Ok - example extended. It also shows a bug in how axes are shown though.

rocky:
Merging in now. Of course more work will and should be done in future PRs.

PR #1549, opened 2021-08-14
More cleaning

TiagoCavalcanteTrindade:
The next is work in uniform polyhedra!

About `to_json`: now it is in its own file instead of the 3d box class, it happens for all primitives, so there is no why to check if the convert function is inexistent.

### DIFF
--- a/mathics/format/json.py
+++ b/mathics/format/json.py
@@ -44,7 +44,8 @@ def convert_coord_collection(
                 "opacity": opacity,
                 "color": color[:3],
             },
-        } for items in collection
+        }
+        for items in collection
     ]
 
     # print(data)
@@ -125,7 +126,6 @@ def line_3d_box(self):
     Compact (lower-level) JSON formatting of a Line3DBox.
     """
     # TODO: account for line widths and style
-    data = []
     color = self.edge_color.to_rgba()
     data = convert_coord_collection(self.lines, "line", color)
     # print("### json Line3DBox", data)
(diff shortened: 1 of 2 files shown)


mathics/format/json.py
```diff
@@ -59,10 +59,7 @@ def graphics_3D_elements(self, **options) -> list:
     result = []
     for element in self.elements:
         format_fn = lookup_method(element, "json")
-        if format_fn is None:
-            result += element.to_json()
-        else:
-            result += format_fn(element)
+        result += format_fn(element)
```
rocky, line 62: this assumes all of the to_json routines have been converted.

rocky (approved):
Personally, I find this kind stuff turning loops into longer more complicated comprehensions not all that interesting.  Hey, if there were a _comment_ describing the intent that might balance the complication added. 

And when you go to debug code like this it is more difficult to work with. It is precisely for this kind of prevalent stuff that I spent years writing more sophisticated debuggers than pdb. 

But I realize I am bucking the trend here.  So be it.

rocky:
I don't view this as necessarily cleaning so much as changing from one style to another.

mathics/format/json.py
TiagoCavalcanteTrindade: They do, I made a Ctrl+Shift+F (search in all files) and nothing of `to_json`.

TiagoCavalcanteTrindade:
> Personally, I find this kind stuff turning loops into longer more complicated comprehensions not all that interesting. Hey, if there were a comment describing the intent that might balance the complication added.
>
> And when you go to debug code like this it is more difficult to work with. It is precisely for this kind of prevalent stuff that I spent years writing more sophisticated debuggers than pdb.
> 
> But I realize I am bucking the trend here. So be it.

I know the difference isn't that big, but things like `Point[{{0, 0, 0}, a thousand of elements, {1000, 1000, 1000}}]` are going to benefit from this.

PR #1561, opened 2021-08-24
Replace `_?TypeQ` by `_Type`

TiagoCavalcanteTrindade:
Revert the bug in `Complex`.

### DIFF
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -316,7 +316,7 @@ class Re(SympyFunction):
         return number.real
 
     def apply_number(self, number, evaluation):
-        "Re[number_?NumberQ]"
+        "Re[number_Number]"
 
         return number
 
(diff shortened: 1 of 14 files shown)


mmatera:
The problem I see with this is that `_Type` actually means `_Head`, i.e. it matches with something of the form `Head[...]`
`Real` and `Complex` are "heads" for numbers, but `Number` I think it is not the head of any expression. This is why in many places we have `_?NumberQ` 
Did you guys change the way in which patterns are interpreted for numbers?

rocky:
> Did you guys change the way in which patterns are interpreted for numbers?

No - we just don't understand WL that well (and I haven't studied the WL docs like I should)

TiagoCavalcanteTrindade:
@mmatera no. I'll revert it too. Strangely the tests passed. Doing tests `_Number` isn't working. `_List` is. I'll look at it more deeply tomorrow.

There is a lot of `_?NumericQ`, and I believe that a lot of it can be replaced by `_?NumberQ`. The 2nd is more than 3 times faster.

PR #1569, opened 2021-08-31
trying to improve performance for NumericQ/NumberQ tests

mmatera:
Here we go again... another try to make NumericQ faster...

### DIFF
--- a/mathics/builtin/comparison.py
+++ b/mathics/builtin/comparison.py
@@ -24,11 +24,14 @@ from mathics.core.expression import (
     Integer1,
     Number,
     Symbol,
-    SymbolFalse,
-    SymbolTrue,
-    SymbolDirectedInfinity,
-    SymbolInfinity,
     SymbolComplexInfinity,
+    SymbolDirectedInfinity,
+    SymbolFalse,
+    SymbolGreater,
+    SymbolInfinity,
+    SymbolLess,
+    SymbolSequence,
+    SymbolTrue,
 )
 from mathics.core.numbers import dps
 
@@ -198,7 +201,7 @@ class _InequalityOperator(BinaryOperator):
     def numerify_args(items, evaluation):
         items_sequence = items.get_sequence()
         all_numeric = all(
-            item.is_numeric() and item.get_precision() is None
+            item.is_numeric(evaluation) and item.get_precision() is None
             for item in items_sequence
         )
 
@@ -374,13 +377,10 @@ class _EqualityOperator(_InequalityOperator):
 class _ComparisonOperator(_InequalityOperator):
     "Compares arguments in a chain e.g. a < b < c compares a < b and b < c."
 
-    def apply(self, items, evaluation):
-        "%(name)s[items___]"
-        items_sequence = items.get_sequence()
-        if len(items_sequence) <= 1:
-            return SymbolTrue
-        items = self.numerify_args(items, evaluation)
-        wanted = operators[self.get_name()]
+    @staticmethod
+    def sapply(wanted, items, evaluation):
+        items = _ComparisonOperator.numerify_args(items, evaluation)
+
         for i in range(len(items) - 1):
             x = items[i]
             y = items[i + 1]
@@ -391,6 +391,13 @@ class _ComparisonOperator(_InequalityOperator):
                 return SymbolFalse
             assert c in wanted
         return SymbolTrue
+
+    def apply(self, items, evaluation):
+        "%(name)s[items___]"
+        items_sequence = items.get_sequence()
+        if len(items_sequence) <= 1:
+            return SymbolTrue
+        return self.sapply(operators[self.get_name()], items, evaluation)
 
 
 class Inequality(Builtin):
@@ -824,9 +831,25 @@ class Positive(Builtin):
 
     attributes = ("Listable",)
 
-    rules = {
-        "Positive[x_?NumericQ]": "If[x > 0, True, False, False]",
-    }
+    def apply(self, x, evaluation):
+        "%(name)s[x_]"
+        if not x.is_numeric(evaluation):
+            return
+
+        if isinstance(x, Complex):
+            return
+        elif isinstance(x, Number):
+            res = x.to_python() > 0
+        else:
+            res = Less.sapply(
+                    (-1,),  # operators["System`Less"],
+                    Expression(SymbolSequence, Integer0, x),
+                    evaluation,
+                )
+            if res is None:
+                return
+            res = res.is_true()
+        return SymbolTrue if res else SymbolFalse
 
 
 class Negative(Builtin):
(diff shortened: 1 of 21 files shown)


rocky (commented):
Overall this looks good and is probably much needed. 

However, please,  let's benchmark this to understand what the benefit is. Many thanks.

mathics/builtin/comparison.py
```diff
@@ -376,13 +377,10 @@ def apply_other(self, args, evaluation):
...
     "Compares arguments in a chain e.g. a < b < c compares a < b and b < c."
 
-    def apply(self, items, evaluation):
-        "%(name)s[items___]"
-        items_sequence = items.get_sequence()
-        if len(items_sequence) <= 1:
-            return SymbolTrue
-        items = self.numerify_args(items, evaluation)
-        wanted = operators[self.get_name()]
+    @staticmethod
```
rocky, line 380: This is basically taking what I wrote about `evaluation_fast()` and doing this directly in Python code. 

Sure, this is fine too, and probably we should do more of this as well. 

But as before I really want to more benchmarks such as for inequality. 

And the `benchmarks.py` routine should be moved over to the benchmarks repository. 

It has some nice features, but I note that like everything else it is a monolith that isn't going to scale as is. You run all the tests all the time as opposed to having tests isolated and then combining them in a single report if that's what you want to do.

mathics/builtin/numbers/calculus.py
```diff
@@ -1377,7 +1376,7 @@ class FindRoot(Builtin):
     def apply(self, f, x, x0, evaluation, options):
         "FindRoot[f_, {x_, x0_}, OptionsPattern[]]"
         # First, determine x0 and x
-        x0 = Expression(SymbolN, x0).evaluate(evaluation)
+        x0 = _numeric_evaluation_with_prec(x0, evaluation)
```
rocky, line 1379: @mmatera I would appreciate it if we could try to regularize the names - whether or not this had been done previously. 

Functions that are basically concrete `apply()` method  of Builtin Class X should follow a distinct pattern. In the future we may autogenerate this.

mathics/builtin/numeric.py
```diff
@@ -67,8 +67,13 @@ def _numeric_evaluation_with_prec(expr, evaluation, prec=SymbolMachinePrecision)
         d = get_precision(prec, evaluation)
     except PrecisionValueError:
         return
+    print("expr:",expr)
```
rocky: Was this intended to get added?

mathics/builtin/numeric.py
```diff
@@ -80,19 +85,22 @@ def _numeric_evaluation_with_prec(expr, evaluation, prec=SymbolMachinePrecision)
...
+    if expr.has_form("N", 2):
+        return _numeric_evaluation_with_prec(expr._leaves[0], evaluation, expr._leaves[1])
 
     name = expr.get_lookup_name()
     if name != "":
         nexpr = Expression(SymbolN, expr, prec)
         result = evaluation.definitions.get_value(
             name, "System`NValues", nexpr, evaluation
         )
+        print(nexpr, "->", result)
```
rocky, line 105: Also this print?

mathics/builtin/numbers/calculus.py
mmatera: The point here is that there are certain functions that I think shouldn't be implemented as member functions, to avoid using the Mathics mechanism of calling functions recursively. This is one case:
`_numeric_evaluation_with_prec(x0, evaluation, prec)` 
should do the same that `Expression("N",x0, evaluation, prec)`, but internally, without all the recursive pattern matching, which is very slow.

On the other hand, if someone has a better name for this function, let's change it.

mathics/builtin/comparison.py
mmatera: > This is basically taking what I wrote about `evaluation_fast()` and doing this directly in Python code.
> 
yep. The problem is that I do not know how that function could be implemented. I think that the best is to have Python functions (or maybe C functions?) for functions that are called for often, avoiding the Mathics-parser calling mechanism.



> Sure, this is fine too, and probably we should do more of this as well.
> 
> But as before I really want to more benchmarks such as for inequality.
> 
> And the `benchmarks.py` routine should be moved over to the benchmarks repository.
> 
> It has some nice features, but I note that like everything else it is a monolith that isn't going to scale as is. You run all the tests all the time as opposed to having tests isolated and then combining them in a single report if that's what you want to do.

OK with the benchmarks. When we have a more complete interface, I can try to move the benchmark tests there.

mathics/builtin/numbers/calculus.py
rocky: > The point here is that there are certain functions that I think shouldn't be implemented as member functions, to avoid using the Mathics mechanism of calling functions recursively.

There wasn't any recursive pattern matching using `evaluate_fast()` and the `sapply..`

 This is one case:
> `_numeric_evaluation_with_prec(x0, evaluation, prec)`
> should do the same that `Expression("N",x0, evaluation, prec)`, but internally, without all the recursive pattern matching, which is very slow.

The difference it seems between this and `sapply` is simply that one is a class method and the other isn't in a class at all. In terms of the difference in time there is some, but it is very small. And that's not at all where the time is going.

mmatera: OK, but what do you think is the best? staticmember methods or plane Python functions? I was not sure and for this reason I use one of then on each case...

mathics/builtin/comparison.py
TiagoCavalcante: I like the idea of implementing something in C. The Python C API is hard to use (I think) but is very complete. It can be done in a way that the overhead is almost 0 (if not 0). That can be reached because we can create Python objects from C, not just convert C object to Python.

If it be done, what would be written in C? (at least to start)

Do you and @rocky think that worths translating mathics-scanner to C?

mmatera: I think that mathics-scanner is OK as it is. What is very slow is the pattern matching over S-Expressions. This is what should be improved, maybe reimplementing it in C, or just by looking at a better way of representing Expressions.

mmatera: For example, look at https://github.com/corywalker/expreduce
This is another implementation of WL, but written in Go. The implementation is much more limited, but I think the performance of the pattern matcher is much better.

mathics/builtin/numbers/calculus.py
rocky: I have to think about it, and try some experiments and maybe benchmark. So lemme get back to you on this. In some cases though when a function uses class members like constant

mathics/builtin/comparison.py
TiagoCavalcante: Knowing Go, that should be instantaneous. I'll look at that, but probably the code is too different from Mathics to be useful. I don't know how to split pattern matching. What I already thought was to write the `Buitin` class in C, but that was very hard. Anyway, I'll try that again another day.

The problem of using Go is that it has a garbage collector, therefore, the overhead would be bigger.

TiagoCavalcante: To say the truth, expreduce seems to be very close to Mathics in pattern matching. Dunno if rewriting that in C would be much faster, but with sure that would be faster. Now I have more ideas to what I should do.

mathics/builtin/numbers/calculus.py
TiagoCavalcante: @mmatera I guess it can be used [here](https://github.com/mathics/Mathics/blob/faster_quick_pattern_test/mathics/builtin/arithmetic.py#L137) too.

mathics/builtin/comparison.py
mmatera: The idea was not to try to implement anything in Go, but just to have another open implementation that we can use to compare. 
On the other hand, I do not think that the problem is that `Builtin` classes are written in Python, but the way in which the methods of these classes are accessed during the evaluation.

TiagoCavalcante: By what I understand all apply methods are stored in a big dict.

PR #1570, opened 2021-09-01
Improve Rational performance

TiagoCavalcante:
The benchmark:
```
Mathics git repo /workspaces/mathics-benchmark/Mathics at 58dc69
10000 iterations of Rational...
  1.312635 secs for: Rational[10, 1]                         
  1.235519 secs for: Rational[2, 3]                          

Mathics git repo /workspaces/mathics-benchmark/Mathics at b9b11b
10000 iterations of Rational...
  1.199198 secs for: Rational[10, 1]                         
  1.159303 secs for: Rational[2, 3]
```

Way more important than that tenth of second is that there is a lot of places where instead of `.to_sympy()` we could use the faster `.value`.

### DIFF
--- a/mathics/builtin/arithmetic.py
+++ b/mathics/builtin/arithmetic.py
@@ -876,10 +876,10 @@ class Rational_(Builtin):
     def apply(self, n, m, evaluation):
         "%(name)s[n_Integer, m_Integer]"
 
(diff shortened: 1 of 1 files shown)


rocky (approved):
LGTM

PR #1571, opened 2021-09-04
Small tweaks

TiagoCavalcante:
I don't know why the difference is so big, but...

image

### DIFF
--- a/mathics/builtin/arithfns/basic.py
+++ b/mathics/builtin/arithfns/basic.py
@@ -308,7 +308,7 @@ class Plus(BinaryOperator, SympyFunction):
                 real, imag = value.to_sympy().as_real_imag()
                 if real <= 0 and imag <= 0:
                     return True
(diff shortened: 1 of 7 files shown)


rocky (commented):
The benchmark shows  great `Im` and `Re` improvements. 

Would you please split this PR into just that part and  in put another PR for the other changes with   benchmarks for those? 
 
To better isolate performance improvements, it is preferable  to have smaller and more PRs than larger ones.

Thanks.

rocky:
> I don't know why the difference is so big, but...

Well, we should start understanding why the differences is big.  My hypothesis is that it is more about removing the "apply" methods and replacing them with "isinstance()" calls, than it is about replacing to_python() with .value. 

That doesn't mean we shouldn't replace to_python().  It is more about understanding and focusing on where the bigger gains are. 


> 
> image

TiagoCavalcante:
`Cuboid` was using `to_python()`.
image

The other change keeps the same performance.

PR #1573, opened 2021-09-04
Fix 2d cuboid

TiagoCavalcante:
I was searching for methods with too much applies and I remebered of this.


rocky (approved):
Will be testing and will put in another PR after this merges.

PR #1574, opened 2021-09-04
Improve MakeBoxes performance by rewriting some rules

TiagoCavalcante:
image


rocky:
Will be adding a new PR after merge.

rocky:
@TiagoCavalcante I am seeing pretty small changes and in contrast to what you have maybe the same amount of loss elsewhere. 


report


Will commit, but again I think we should look for better and indisputable wins.

PR #1576, opened 2021-09-05
Small performance improvements though rule rewriting

TiagoCavalcante:
Last small tweaks at least for the next week.

image

I'll update mathics-benchmark so it breaks the query into multiple lines every x characters.

### DIFF
--- a/mathics/builtin/arithfns/basic.py
+++ b/mathics/builtin/arithfns/basic.py
@@ -201,7 +201,7 @@ class Minus(PrefixOperator):
     def apply_int(self, x, evaluation):
         "Minus[x_Integer]"
 
-        return Integer(-x.to_sympy())
+        return Integer(-x.value)
 
 
 class Plus(BinaryOperator, SympyFunction):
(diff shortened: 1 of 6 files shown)


rocky:
@TiagoCavalcante I would greatly appreciate it if you would break down this PR into the part that crucial for the 10% speedup and put the other stuff in another PR.  

We would fix the performance problems sooner and easier  if we were to focus attention and have those changes giving big wins and isolated those a single commit. 

Sure, I understand there is all this other stuff that might be changed and that's okay.  But please, keep the important stuff from all the other things that one might want to do. 

Thanks.

TiagoCavalcante:
Ok, of course, I'll do it too for my other PR tomorrow.

TiagoCavalcante:
@rocky benchmarking again in a better way (without using the machine at the same time), and split into groups I see all my changes except `Minus` and `MantissaExponent` have improvements:
image
image

How should I split this PR?

TiagoCavalcante:
Please ignore `SplitBy`.

rocky:
> @rocky benchmarking again in a better way (without using the machine at the same time), and split into groups I see all my changes except `Minus` and `MantissaExponent` have improvements:
> image
> image
> 
> How should I split this PR?

I'm trying to figure out which is the part that is doing the most speedups. So I'd like to try to replicate the benchmark. Then I'll pare the PR down into the essential part so that it is in one PR. Stay tuned...

rocky:
@TiagoCavalcante merging into a branch so I can duplicate the results and see if I can pinpoint what's what.

PR #1579, opened 2021-09-06
Improve MakeBoxes performance by rewriting some rules

rocky:
Continuation of  #1574


rocky:
@TiagoCavalcante I am seeing pretty small changes and in contrast to what you have maybe greater amount of loss elsewhere. 


report


Will commit, but again I think we should look for better and indisputable wins.

